PR #14, opened 2012-12-17
Groovy devel

dangr:
Additional functionality: More precise likelihood field model and no_motion update.
Fixes: Clear frame_to_laser map on new map.


hershwg:
Hi Dangr, can you give me some background on these changes?  Tell me more about the new field model, how it is different from what was there before.   What problem does the no-motion update solve?  Can it be solved some other way?  (like setting the "update_min_d" parameter to -1).  Why is there a constant "17" in the new code?  (Only every 17th scan without motion seems to be used.)  What is the significance of clearing out the laser/frame map on new map?

I appreciate the contribution, but I need more information before I merge this.

dangr:
Absolutely, here goes:

New field model----------
David's actually the best person to speak to this, he's the one who 
worked the new model out.

No motion update--------
We'd often find that the reported pose was incorrect, and that just 
wiggling the robot in place was sufficient to improve the estimate. A 
no-motion update is just a way to do that without actually having to 
move the robot. I had not considered trying to set the update_min_d 
parameter like that. If it worked, I guess the service could be 
performed by an external node that made dynamic reconfigure calls to 
amcl. But, the dynamic reconfigure callback re-initializes the model, so 
I don't know if that would be useful.

Note, updating without motion is not something we want to do all the 
time - just sometimes, when higher-level planning decides it would be a 
good idea.

The %17 is there for computational reasons - I see I left a note in 
there that this might be pulled from a parameter rather than hardcoded.

Clearing out-------------
I found that when I had use_map_topic set to true, I wasn't getting 
updated particle clouds. I traced the issue to the fact that while the 
lasers were being cleared out when the map was received, the mapping 
from frame_name to laser_index was not being reset. So, in 
laserReceived, the laser would appear to be set up, but the retrieved 
laser_index would be wrong. The lasers_update.size() would also be 0.

On 12/18/2012 01:44 PM, Dave Hershberger wrote:

> Hi Dangr, can you give me some background on these changes?  Tell me
> more about the new field model, how it is different from what was
> there before.   What problem does the no-motion update solve?  Can it
> be solved some other way?  (like setting the "update_min_d" parameter
> to -1).  Why is there a constant "17" in the new code?  (Only every
> 17th scan without motion seems to be used.)  What is the significance
> of clearing out the laser/frame map on new map?
> 
> I appreciate the contribution, but I need more information before I
> merge this.
> 
> --- Reply to this email directly or view it on GitHub:
> https://github.com/ros-planning/navigation/pull/14#issuecomment-11499521
> 
> ## 
> 
> Daniel Grollman, Robot Doctor

Vecna Technologies, Inc.
36 Cambridge Park Drive
Cambridge, MA 02140
(617) 864-0636 x3047
http://www.vecna.com/robotics

Better Technology, Better World (TM)

hershwg:
For the clearing, that sounds like a definite bug, so I will pull that change in as soon as I can review and test it.

For the no-motion update, I would like something that felt like a cleaner solution.  Shouldn't amcl know on its own that an update would improve localization accuracy?  What is the higher-level logic like that does the decision-making?  Is it purely localization logic, or does it tie into other systems?

If it is true that the no-motion logic should really lie outside of amcl, I think I would prefer a topic subscription.  There is no return value from the service call and the timing of the return is also not meaningful.  The only thing the caller gets from the return of their service call is acknowledgement of receipt of the request.

Maybe a more flexible way to do it would be to subscribe to a topic of new type "UpdateLocalization" which has a boolean in it.  Any time amcl receives a "true" from that message, it triggers a no-motion update.  "false" values could be ignored.

For the field model, I'll have to read the changes and/or hear from David.

In general, it will be easier (for me) if things like these 3 changes are submitted as separate pull requests.

We are now quite close to the Groovy ROS release (mostly finishing up documentation at this point) so I would prefer to put changes like these into Hydro.  Maybe the bug fix in Groovy, I'll have to see.

hershwg:
Clearing change pulled into groovy-devel, thanks.

hershwg:
I have cherry-picked d670c7b (nomotion request service) into groovy-devel instead of merging this pull request, because after all the reversions and merges here, that is the only resulting change as far as I can tell.

dangr mentioned in a separate email that he would be dropping the likelihood field model optimizations for now, and I've already pulled in the bug-fix in the clearing code, so that addresses all 3 concerns of this pull request.  Closing.

PR #19, opened 2013-01-17
Adding support for absolute position inputs (GPS/LPS systems)

rgariepy:
We've tested this on our ROS Fuerte systems with a few GPS systems and an Optitrack system, seems good. Despite my name (rgariepy) being on this, the primary author/tester is Prasenjit Mukherjee (pmukherj@clearpathrobotics.com)

How do we backport this into the Fuerte/Electric repos?

### DIFF
--- a/robot_pose_ekf/robot_pose_ekf.launch
+++ b/robot_pose_ekf/robot_pose_ekf.launch
@@ -1,15 +1,17 @@
 <launch>
 
-<node pkg="robot_pose_ekf" type="robot_pose_ekf" name="robot_pose_ekf">
-  <param name="output_frame" value="odom_combined"/>
+<node pkg="robot_pose_ekf" type="robot_pose_ekf" name="robot_pose_ekf" output="screen">
+  <param name="output_frame" value="odom"/>
   <param name="freq" value="30.0"/>
+  <param name="debug" value="true"/>
   <param name="sensor_timeout" value="1.0"/>  
   <param name="odom_used" value="true"/>
   <param name="imu_used" value="true"/>
-  <param name="vo_used" value="true"/>
+  <param name="vo_used" value="false"/>
+  <param name="gps_used" value="true"/>
 
-  <remap from="odom" to="pr2_base_odometry/odom" />
 </node>
+
 
 </launch>
 
--- a/robot_pose_ekf/include/robot_pose_ekf/odom_estimation.h
+++ b/robot_pose_ekf/include/robot_pose_ekf/odom_estimation.h
@@ -69,12 +69,13 @@ public:
   /** update the extended Kalman filter
    * \param odom_active specifies if the odometry sensor is active or not
    * \param imu_active specifies if the imu sensor is active or not
+   * \param gps_active specifies if the gps sensor is active or not
    * \param vo_active specifies if the vo sensor is active or not
    * \param filter_time update the ekf up to this time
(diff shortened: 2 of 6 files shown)


robot_pose_ekf/robot_pose_ekf.launch
```diff
@@ -1,15 +1,17 @@
...
+<node pkg="robot_pose_ekf" type="robot_pose_ekf" name="robot_pose_ekf" output="screen">
+  <param name="output_frame" value="odom"/>
   <param name="freq" value="30.0"/>
+  <param name="debug" value="true"/>
   <param name="sensor_timeout" value="1.0"/>  
   <param name="odom_used" value="true"/>
   <param name="imu_used" value="true"/>
-  <param name="vo_used" value="true"/>
+  <param name="vo_used" value="false"/>
+  <param name="gps_used" value="true"/>
```
hershwg, line 11: This file is, I believe, an example for how to use robot_pose_ekf.  In case anyone references it from their own launch files, it would be better to copy the file and make your changes there.  Perhaps call the new file "gps_example.launch".

hershwg:
Thanks for contributing!

Overall I like this.  Before I merge it, I would like to see some regression tests added in the test directory which exercise the new functionality.

I committed a small change to the test launch files (in groovy-devel branch) which you will need to merge in before 'make test' will succeed.  Apparently the rosbag usage has changed since those tests were last run.

You say you've tested this on ROS Fuerte, but it seems to me your change was made on the groovy-devel branch.   Is that right?  Should I be applying the change to the groovy-devel branch?

Backporting to Fuerte or Electric is hopefully pretty easy.   Fuerte is in the 'navigation-1.8' branch and Electric is in the 'navigation-1.6' branch.  You just check out the branch in question, then do `git cherry-pick <commit-id>`, test it, and push when it works.

robot_pose_ekf/robot_pose_ekf.launch
pmukherj, line 11: Agreed. Its probably more useful to have a separate launch to make the differences clear. I'll make that happen.

pmukherj:
Thanks for the comments! I'll look into the testing as soon as I can.
I believe it makes more sense for us to get the changes into Fuerte first. We can backport to Electric if required then. Apologies for it being in groovy-devel, we should switch that to Electric.

pmukherj:
Hello,

Apologies for the radio silence, a few other projects diverted our attention. I was just setting up some regression tests, when I noticed the current test launch files in the test directory do not have bag files they call (ekf_test2.bag, zero_covariance.bag).  Was this by design? Maybe I'm missing something basic.

hershwg:
You are just missing the calls to rosbuild_download_test_data() in CMakeLists.txt.

You can either commit you bag files or host them on your own web server and I'll upload them to our "official" place.

hershwg:
I have also pulled this separately into hydro-devel.

hershwg:
I would still love to have the tests added that exercise this, but I wanted to make sure this got into the hydro release sooner rather than later.

pmukherj:
Absolutely! The test framework for GPS are complete (its actually checked in). I am running a few more tests (with different GPS receivers and such) as we speak. I am hoping to make a pull request into navigation-1.8 in the next day or two. It should be easily portable to groovy/hydra.

hershwg:
excellent, glad to hear it

PR #21, opened 2013-02-21
include default github .gitignore content

ablasdel:
This should make peoples lives a bit easier when working with the nav stack from git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,78 @@
 *~
-build
-bin
-lib
-*.pyc
 *.cfgc
 msg_gen
 srv_gen
 test/line_iterator
 test/utest
+
+#github standard C 
+# Object files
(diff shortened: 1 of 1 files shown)


hershwg:
Does this make a change in output that you see yourself?  From looking through it I don't see anything that would change.

I didn't spend more time on this because once the nav stack is catkinized we won't be building it in the source tree anyway, and this will mostly stop being an issue.

ablasdel:
BInney asked for this specifically so I just did it. To be honest I didn't check the output. I will do that now.

ablasdel:
I added nav specific lines to the .gitignore in this commit f590c5a43d1a41a99808ed0b4b892be9c13eb3e4 to clear up the build directories that need to be ignored for now.

PR #23, opened 2013-04-02
costmap_2d: fixed #22 by slightly decreasing inscribed radius.

### LINKED ISSUE #22: When very close to obstacle, costmap_2d incorrectly prevents robot from moving.
Background:

costmap_2d has a notion of "inflating" obstacles, first to the "inscribed radius" of the robot's footprint polygon with a value called "INSCRIBED_INFLATED_OBSTACLE" (253) and then down from there on a curve until it gets to the max inflation radius set in the parameters.

The idea with the inscribed radius is that once you inflate an obstacle, the _center_ of the robot can't be on a point marked INSCRIBED_INFLATED_OBSTACLE because you know that is closer than the minimum radius of the robot to an obstacle.  So then you can use that to make a really fast way of checking if the robot is allowed to be somewhere or not - first check the center of the robot for that value, if it is there, the answer is a quick "NO".  If it is something less than that, you have to take the perimeter of the robot into account and the orientation of it, etc.

Bug:

The current version of costmap_2d inflates obstacles just a little bit too much.  It uses the "ceil()" function to round UP to the nearest integer the robot's inscribed radius divided by the size of a map cell.  The result is, when the robot's bounding polygon (at its inscribed radius) is immediately next to an obstacle cell in the costmap, the obstacle cell will inflate to cover the current center of the robot.  Because it is over the current center, the local planner and global planner will never find any valid plan away from that point and the robot will be stuck forever.  Just zero-ing out the current robot position after inflation is not sufficient, because the robot may be rectangular and have room to drive straight ahead, but it would still think it can't move off of that spot.

hershwg:
This would fix #22 by changing the cell_inscribed_radius_ member value from an unsigned int equal to ceil( inscribed_radius / resolution ) to a _double_ equal to inscribed_radius / resolution.

It turns out that the only place in the code where cell_inscribed_radius_ is even used is in a "<=" comparison with a double value which is the distance between two grid cells.  Therefore this change will only change a few of the pixels around the perimeter of the obstacle-inflation circle.

If this will change peoples' robot behaviors in a bad way, I can certainly make a parameter which lets you choose the old way or the new way.  There are so many parameters already though that I hesitate to add another, plus I really think this is likely to only improve behavior and not cause problems.

### DIFF
--- a/costmap_2d/src/costmap_2d.cpp
+++ b/costmap_2d/src/costmap_2d.cpp
@@ -61,7 +61,7 @@ namespace costmap_2d{
     memset(markers_, 0, size_x_ * size_y_ * sizeof(unsigned char));
 
     //convert our inflations from world to cell distance
-    cell_inscribed_radius_ = cellDistance(inscribed_radius);
+    cell_inscribed_radius_ = max( 0.0, inscribed_radius / resolution_ );
     cell_circumscribed_radius_ = cellDistance(circumscribed_radius);
     cell_inflation_radius_ = cellDistance(inflation_radius);
 
--- a/costmap_2d/include/costmap_2d/costmap_2d.h
+++ b/costmap_2d/include/costmap_2d/costmap_2d.h
@@ -675,7 +675,8 @@ namespace costmap_2d {
       unsigned char** cached_costs_;
       double** cached_distances_;
       double inscribed_radius_, circumscribed_radius_, inflation_radius_;
-      unsigned int cell_inscribed_radius_, cell_circumscribed_radius_, cell_inflation_radius_;
+      double cell_inscribed_radius_;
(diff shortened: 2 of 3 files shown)


costmap_2d/src/costmap_2d.cpp
```diff
@@ -61,7 +61,7 @@
     memset(markers_, 0, size_x_ * size_y_ * sizeof(unsigned char));
 
     //convert our inflations from world to cell distance
-    cell_inscribed_radius_ = cellDistance(inscribed_radius);
+    cell_inscribed_radius_ = max( 0.0, inscribed_radius / resolution_ );
```
hershwg, line 64: This new line is the same as the contents of the cellDistance() function, but without the ceil() function or the cast to unsigned int.

PR #24, opened 2013-04-03
Catkinized hydro devel

jonbinney:
Catknizes the hydro-devel branch. Most of the work was already done by user @contradict ; I just pulled from his fork, fixed a few things to get it to compile, and rebased on the hydro-devel branch.

Although this builds for me, I suspect that we'll continue to find little bugs that keep other packages from building against it (might need to add more things to catkin_package commands, etc). I think that the best thing to do is to switch to catkin now, and fix these as we go.

I also have a catknized version of the groovy-devel branch in my fork, but catkinizing such a large project midway through groovy seems like a bad idea. One solution would be the following:
- groovy-devel branch continues to use rosbuild, only very basic bugfixes go here. releases for groovy continue to happen from this branch
- catkinized-groovy-devel branch gets created; people who are still on groovy but would like to use catkin can use this branch from source. backporting changes from the hydro-devel branch will be easier since they'll both be catkinized
- hydro-devel branch will be catknized.

If that sounds reasonable, @hershwg could you create a catkinized-groovy-devel branch, and I'll pull request the catknized groovy code into there?


tfoote:
That sounds like a good approach for managing the contributions for this stack.

Issue #3, opened 2012-11-12
Map server should crop unknown space


hershwg:
[Originally reported by moesenle on kforge trac]

It happens too often that people create a really large map (e.g. 200x200m, 2.5cm resolution ~ 64mb) that mostly contain unknown space. Such large maps actually caused rviz to eat up all my memory during the PR2 workshop last week. That makes be believe that it would be best if the map server actually crops everything that doesn't really contain any information, i.e. that's unknown space.

Find a patch for cropping the map attached.. I added a ros parameter crop that allows for enabling (currently default) or disabling that functionality.

[Download link for original attachment]
https://kforge.ros.org/navigation/trac/raw-attachment/ticket/5/map-server-crop-map.patch

DLu:
I don't think this should be implemented here (in map_server). Since the size of the global costmap is based on the size of the static map, you might want to leave unexplored regions of the map available. 

I think this would be better put into the map_saver with the (automatic) option to crop. 

Anyone feel strongly either way?

corot:
+1 to map_saver

Another usecase is maps without defined limits (typically on simulations,
like gazebo turtlebot demo)

2013/11/8 David Lu!! notifications@github.com

dornhege:
+1 map_server just serving the map without modification sounds the most intuitive to me.

mikeferguson:
seems there is a consensus that we don't want to implement this feature -- and the patch has also been lost when kforge was shutdown

PR #29, opened 2013-04-16
map_server: add crop_map.py to crop a map to the known area

dornhege:
Related to issue #3, but does that in a separate script that outputs the cropped yaml file + pgm.

The script is independent from anything and could also be merged into other devel branches or moved to another package.


hershwg:
Looks nice, thanks!

PR #35, opened 2013-05-03
Base local planner

DLu:


hershwg:
This is a pretty huge pull request, it is taking me a while to deal with it.

It is taking me a while to figure out which of these changes make a change to the old class API and which are adding or removing things from/to the old API.  It would be great also if your commit messages were a little more descriptive too, for instance one is called "World Model Hack" and contains only the _addition_ of (but not a call to) a function called footprintCostA().  What is it for?  Who needs it?  If you feel the need to call something a hack or name it obscurely, please add a comment in the code that explains why you are doing it.

hershwg:
Have you run this?  I'm getting a seg fault in footprint_costmap_plugin.cpp on line 65.  It's weird because line 65 is the closing paren of the namespace.  I thought it might be a library/source mismatch, but now I've just done a clean build of the whole nav stack and my dependent code and it still does it.

hershwg:
Oh right, I looked back at your chat and found that you had not tested it, sorry.

hershwg:
Closing this, since I manually merged from further down this branch.

PR #40, opened 2013-05-07
Get map action

jonbinney:
Adds an actionlib action server for getting the map from the map_server. Works the same way as the static_map service call, but with an action.

This pull requests requires a new action definition in common_msgs, and so can't be merged until this one is merged:
https://github.com/ros/common_msgs/pull/8


jonbinney:
ummmmmm don't merge this. i meant to pull request it to the catkinized layered costmap branch....closing...

PR #41, opened 2013-05-08
robot_pose_ekf: use tf_prefix to lookup and send transforms

mintar:


hershwg:
This looks good as far as I can tell.  Sadly "make test" is not reliable, sometimes passes, sometimes fails.  I would prefer you submit this with tests confirming that it does the right thing, but since the existing tests don't reliably work, I'm not going to be so demanding. :)

Are you using this and it works for you, I presume?

mintar:
I can understand you're picky about changing something as essential as robot_pose_ekf. But yes, we've been using this extensively, and it works great. This feature is needed if you're trying to run robot_pose_ekf on multiple robots on the same ROS core.

Off topic: Thanks for responding so quickly, I really feel that since switching to Github it makes sense to submit patches and report bugs again. We've reported this bug a long time ago [here](https://code.ros.org/trac/ros-pkg/ticket/5457), but never got any answer.

hershwg:
Github is indeed much better than code.ros.org, I totally agree.  I appreciate your persistence.

PR #42, opened 2013-05-08
Get map action

jonbinney:
Adds an actionlib action server for getting the map from the map_server. Works the same way as the static_map service call, but with an action.

This pull requests requires a new action definition in common_msgs, and so can't be merged until this one is merged:
ros/common_msgs#8

### DIFF
--- a/map_server/CMakeLists.txt
+++ b/map_server/CMakeLists.txt
@@ -9,6 +9,8 @@ find_package(catkin REQUIRED
             tf
             nav_msgs
             rostest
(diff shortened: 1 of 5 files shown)


hershwg:
ros/common_msgs#8 has now been merged.

Do you want to bring the source into our .rosinstall files, or wait until it gets released?  I recommend using the source - you never know how long it might take for a release to come out.

hershwg:
waiting on a fresh PR to hydro-devel...

PR #43, opened 2013-05-09
costmap_2d: fix radius warning when footprint is set

mintar:
If the footprint parameter is set, inscribed and circumscribed radii are
computed from that. However, the check was done _before_ that computation
and therefore used the default params. This commit
- moves the check after the computation
- changes the wording from "inscribed and circumscribed radii" to
  "inscribed or circumscribed radius".

Also see [this](http://answers.ros.org/question/62277/inscribed-and-circumscribed-radii-of-the-robot-used-despite-footprint-being-set/) ros answers thread.


hershwg:
Thanks!

Issue #38, opened 2013-05-07
Is move_base's parameter clearing_rotation_allowed actually working?


corot:
Setting clearing_rotation_allowed parameter as false doesn't prevent the robot to make in place rotations as it's supposed to do (to my understanding of the wiki documentation). In fact, it looks like the parameter value has no effect at all.

There'is any problem or I misunderstood the use of this parameter?

hershwg:
From the wiki (and confirmed by reading the source code):

"clearing_rotation_allowed (bool, default: true): Determines whether or not the robot will attempt an in-place rotation when attempting to clear out space. _Note: This parameter is only used when the default recovery behaviors are in use, meaning the user has not set the recovery_behaviors parameter to anything custom._"

If you have a "recovery_behaviors" section in your move_base parameters, that makes clearing_rotation_allowed be ignored.  If you have such a section and it contains a recovery behavior called "rotate_recovery" then just take it out.

However, I will say that sadly there are tons of situations that can make a robot just rotate in place which are not related to rotate_recovery.  Eliminating all of them is a subject too long for this bug discussion.

corot:
mmm....  but I'm using default recovery behaviors, and the log says "rotating to clear costmap" or something like this, so I think is a recovery behavior what makes the robot spins.

hershwg:
Well OK then! :)  Sorry, I didn't realize how clear your situation was.

What version of move_base are you running?

hershwg:
Looks like it is a spelling error in the code.  The code is waiting for "clearing_roatation_allowed".  Sigh.

corot:
Cool! it will be very useful to me. Do u want me to do a pull request? 

Btw, I use hydro-devel sources.

hershwg:
You can use it right away by misspelling the parameter in your rosparams the same way it is misspelled in the code.  I'll get a fix in later today.

PR #46, opened 2013-05-13
Fix issues #3 and #38 on hydro-devel

corot:
- For issue #3 I just applied the patch that Dave posted in the issue with a minimal change required to make it compile and added the script provided by dornhege.
- For issue #38 I just find replace clearing_roatation_allowed by clearing_rotation_allowed, what fixes the problem, as Dave said.

I hope I didn't something stupid... as is my second pull request. Please tell me if so.


corot:
yes...  I did something really stupid...  like breaking costmap2d.  Forget this pull request, and sorry if you lost a couple a while reading it.

PR #47, opened 2013-05-13
Pointvectors

DLu:
Based on discussion here: 
https://groups.google.com/group/ros-sig-navigation/browse_thread/thread/c6ec688240f59848


hershwg:
Thanks for this David!

PR #50, opened 2013-05-14
added missing message dependency

jonbinney:


hershwg:
Jon, as of yesterday hydro-devel is the correct branch to use for navigation and platformbot_navigation.

PR #51, opened 2013-05-14
Pull in commit that went to wrong branch

hershwg:


hershwg:
I added the dependency independently, closing.

PR #53, opened 2013-05-18
Working Unit Tests

DLu:
A bunch of work I did to get the existing costmap_2d tests to work with the new framework. 

Includes:
- Some minor work to get tests outside of costmap to compile/run (amcl/voxel_grid)
- Splitting module_tests.cpp into three files for the static, inflation and obstacle layers
- Note: Static tests were less relevant so I did not port them (yet?)
- Note2: module_tests.cpp is still around, but not compiled
- Additional functionality for easy testing (mostly adding a way to add obstacles without subscriptions) 
- A static map in two files (test/TenByTen.*) that mirror the static map used in module_tests

Errors fixed while testing: 
- Integer bounds for inflation are non-inclusive, so fixed one boundary error [1]
- Include the origin of observations for bounds checking [2]

Tests that remain broken:
- In inflation tests, one part of testInflation3 fails because it counts FREE_SPACE cells. However, the number the old test was expecting incudes all cells that have updated values. Since some of the updated values are 0, they are counted as free space, and thus the test reports more free space than it should. It's unclear whether this worked in the original tests. 

[1] https://github.com/DLu/navigation/blob/b9858f93c71bb3577b13d3d7f0524506f70c390e/costmap_2d/plugins/inflation_layer.cpp#L110
[2] https://github.com/DLu/navigation/blob/b9858f93c71bb3577b13d3d7f0524506f70c390e/costmap_2d/plugins/obstacle_layer.cpp#L400
[3] https://github.com/DLu/navigation/blob/b9858f93c71bb3577b13d3d7f0524506f70c390e/costmap_2d/test/inflation_tests.cpp#L306


hershwg:
Thanks, it looks like a bunch of work... but it doesn't compile for me.  I fixed a problem with the inlined computeCost() function in InflationLayer, but now there are other problems too.

hershwg:
OK, I fixed some more problems and got the inflation test to pass.  The problem was that the cost_scaling_factor parameter used to be set to 1.0 in the old test, but the API to do that was removed, so it was defaulting to 10.0.  I added a <param> line in the test's .launch file and now it works.

hershwg:
I merged this pull request on the command line, then added some commits of my own, and github did not notice, so I closed the pull request manually.

PR #55, opened 2013-05-20
Base Local Planner fixes

DLu:


base_local_planner/test/wavefront_map_accessor.h
```diff
@@ -17,7 +17,7 @@
 class WavefrontMapAccessor : public costmap_2d::Costmap2D {
   public:
     WavefrontMapAccessor(MapGrid* map, double outer_radius)
-      : costmap_2d::Costmap2D(map->size_x_, map->size_y_, 1, 0, 0, 5, 10, 15),
+      : costmap_2d::Costmap2D("some_frame", map->size_x_, map->size_y_, 1, 0, 0), // , 5, 10, 15
```
hershwg, line 20: Think about future readers of the code when you write comments.  The comment "//, 5, 10, 15" only makes sense to the person looking at the diff here, but to everyone else in the future it will be a mystery.  If you really need to keep that info in there, find some way to explain it.  If it's not useful, just delete it.

PR #59, opened 2013-05-28
the static_map flag needs to be set to true ...

piyushk:
...and should not simply set to load the static layer if the flag exists.


hershwg:
Thanks!

PR #61, opened 2013-05-29
Issue #9: parameter tf_broadcast added to allow disabling transform broadcasting

corot:


hershwg:
Thanks for this.

I also added dynamic-reconfigure support for the new variable and then cherry-picked the changes into groovy-devel and hydro-devel.

PR #65, opened 2013-05-29
making the image_loader API inside map_server reusable by other projects

piyushk:
I was using the image_loader API in a couple of projects here. This patch makes the catkin version available outside the package.


hershwg:
Good idea, thanks.

Issue #74, opened 2013-06-19
clear_costmap_recovery not working with multi-layer costmaps


corot:

corot:
Should not we close this? What to add something more?

hershwg:
Yup, closing.

PR #75, opened 2013-06-19
Issue #74: Looks for obstacles layer, but the default plugin name is obstacle_layer

corot:


corot:
aggg...  I didn't know github accumulates subsequent commits in the same pull request. Sorry for mangling things

hershwg:
Thanks for this!  I will expand on it a bit and make it more configurable, but you found the problem.

PR #82, opened 2013-07-02
Fix building osx

wjwwood:
These only include fixes I had to make in order to get things building on OS X. All of the fixes should be applied for Linux as well, but Linux tends to let things slide, so they don't tend to so up as problems.

Also I built with `catkin_make_isolated` to find the missing install rules for headers.


wjwwood:
I'll open another pull request with some not critical bug fixes, then I plan to do yet another pull request with pedantic changes, which you may or may not want to merge.

ZdenekM:
I can't even make it on Linux (12.04, ROS Groovy). Still getting:

``` bash
Linking CXX executable /home/imaterna/ros/catkin_ws/devel/lib/robot_pose_ekf/test_robot_pose_ekf
Linking CXX executable /home/imaterna/ros/catkin_ws/devel/lib/robot_pose_ekf/test_robot_pose_ekf_zero_covariance
/usr/bin/ld: cannot find -lgtest
collect2: ld returned 1 exit status
make[2]: *** [/home/imaterna/ros/catkin_ws/devel/lib/robot_pose_ekf/test_robot_pose_ekf] Error 1
make[1]: *** [navigation/robot_pose_ekf/CMakeFiles/test_robot_pose_ekf.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

wjwwood:
You don't have gtest installed?

ZdenekM:
I just find out that on 12.04 one must compile libgtest manually... See http://askubuntu.com/questions/145887/why-no-library-files-installed-for-google-test-on-12-04. Bit annoying but works :)

hershwg:
Thanks @wjwwood !

PR #84, opened 2013-07-04
Edits necessary for a source compile on OSX with catkin

hawesie:
Edits necessary for a source compile on OSX with catkin. All are to include necessary linker commands.


hershwg:
Thanks!

hawesie:
No probs. Thanks for maintaining this!

Issue #71, opened 2013-06-11
'Service already advertised' on startup


dgossow:
I get these on startup:

```
[ INFO] [1370991386.152623463]: Received a 1203 X 993 map at 0.050000 m/pix
process[slot_io-17]: started with pid [27776]
[ INFO] [1370991386.262703727]: Using plugin "obstacles"
[ INFO] [1370991386.268073389]:     Subscribed to Topics: head_cloud_marking head_cloud_clearing base_scan
[ERROR] [1370991386.392674340]: Tried to advertise a service that is already advertised in this node [/move_base/global_costmap/obstacles/set_parameters]
[ INFO] [1370991386.442401938]: Using plugin "inflater"
[ INFO] [1370991387.002144270]: Using plugin "obstacles"
[ WARN] [1370991387.113406247]: Map update loop missed its desired rate of 3.0000Hz... the loop actually took 0.6102 seconds
[ WARN] [1370991387.137904135]: The root link base_link has an inertia specified in the URDF, but KDL does not support a root link with an inertia.  As a workaround, you can add
 an extra dummy link to your URDF.
[ INFO] [1370991387.138267428]:     Subscribed to Topics: head_cloud_marking head_cloud_clearing base_scan
[ERROR] [1370991387.241637697]: Tried to advertise a service that is already advertised in this node [/move_base/local_costmap/obstacles/set_parameters]
...

Might not be a functional issue, but something seems to be wrong there.
```

hershwg:
Those are coming from dynamic reconfigure.  We must have something wrong with them because they are all supposed to have different namespaces.  There are lots of plugins each doing their own dynamic reconfigure server, each of which advertises a service for setting params.  If any two dynamic reconfigure servers are given the same name you will get this error.

DLu:
The problem occurs using the VoxelLayer, since both initialize the dsrv_ (dynamic reconfigure server) node with the same namespace. Working on a fix.

PR #98, opened 2013-07-30
Fix for #71

DLu:
Fix for Issue #71.

### DIFF
--- a/costmap_2d/plugins/voxel_layer.cpp
+++ b/costmap_2d/plugins/voxel_layer.cpp
@@ -34,9 +34,9 @@ void VoxelLayer::initMaps()
   ROS_ASSERT(voxel_grid_.sizeX() == size_x_ && voxel_grid_.sizeY() == size_y_);
 }
 
-void VoxelLayer::setupDynamicReconfigure()
+void VoxelLayer::setupDynamicReconfigure(ros::NodeHandle& nh)
 {
-  dsrv_ = new dynamic_reconfigure::Server<costmap_2d::VoxelPluginConfig>(private_nh);
+  dsrv_ = new dynamic_reconfigure::Server<costmap_2d::VoxelPluginConfig>(nh);
   dynamic_reconfigure::Server<costmap_2d::VoxelPluginConfig>::CallbackType cb = boost::bind(
       &VoxelLayer::reconfigureCB, this, _1, _2);
   dsrv_->setCallback(cb);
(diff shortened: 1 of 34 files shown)


hershwg:
Hey David, thanks for working on this.

Sorry to be picky about it, but could you change this to solve it with a virtual function instead of adding the flag that tells it whether the instance is a subclass or superclass?

Solution would look like:

```
class ObstacleLayer {
  virtual void setUpDynamicReconfigure() {
    dsrv_ = ...; // stuff specific to ObstacleLayer.
  }
  virtual void onInitialize() {
    setupDynamicReconfigure();
    ...; // other ObstacleLayer intiializations
  }
};

class VoxelLayer: public ObstacleLayer {
  virtual void setupDynamicReconfigure() {
    dsrv_ = ...; // stuff specific to VoxelLayer
  }
  virtual void onInitialize() {
    ObstacleLayer::onInitialize();
    ...; // voxel-specific init, excluding dynamic reconfigure.
  }
};
```

That way the call to setupDynamicReconfigure() in ObstacleLayer::onInitialize() will call the correct dynamic-reconfigure setup based on the object type, using the built-in mechanisms of C++.  It's more clear (in my opinion) and certainly more extendable than the flag version.

DLu:
Note, I don't have a handy hydro setup, so you should compile this just to be sure (if you weren't already). 

P.S. You're allowed to be picky. I had the feeling there was a better way.

costmap_2d/plugins/voxel_layer.cpp
```diff
@@ -20,10 +20,7 @@ void VoxelLayer::onInitialize()
   ObstacleLayer::onInitialize();
   ros::NodeHandle private_nh("~/" + name_);
 
-  dsrv_ = new dynamic_reconfigure::Server<costmap_2d::VoxelPluginConfig>(private_nh);
-  dynamic_reconfigure::Server<costmap_2d::VoxelPluginConfig>::CallbackType cb = boost::bind(
-      &VoxelLayer::reconfigureCB, this, _1, _2);
-  dsrv_->setCallback(cb);
+  setupDynamicReconfigure();
```
hershwg: This line is not needed, because the call to ObstacleLayer::onInitialize() has already called setupDynamicReconfigure().

hershwg:
also, to test your changes...

```
sudo apt-get install ros-hydro-navigation
cd ~
mkdir -p test-my-nav/src
cd test-my-nav/src
git clone -b hydro git@github.com:DLu/navigation
cd ~/test-my-nav
source /opt/ros/hydro/setup.bash
catkin_make -DCMAKE_BUILD_TYPE=RelWithDebInfo
catkin_make run_tests
source devel/setup.bash
catkin_test_results
```

hershwg:
lgtm.

PR #107, opened 2013-08-13
Dependencies in runtime due to default parameter settings.

stonier:
These packages were brought in as build_depends because of default settings in the code but they're currently only build_depends, so debs are not installed alongside move base, meaning any default launch of move_base fails.

Was this actually an oversight (this pull request fixes), or is there a good logical reason we're all missing here that these have been avoided as runtime dependencies?


stonier:
Also added rotate recovery as it too is one of the defaults.

hershwg:
Oversight!

hershwg:
thanks

PR #108, opened 2013-08-13
Trivial build updates

stonier:
These aren't high priority.
- Updates the deprecated catkin download function.
- Uniquely names the map_save target (doesn't change resulting filename) so it doesn't clash with map_store's target when building both from source in the same catkin_workspace.


hershwg:
trivial... yet awesome.

thanks!

PR #109, opened 2013-08-14
Fixed some more errors of the direct use of pointcloud object

mikegao88:
In addition, for the personal convenience, I added an initialization interface for the odometry_helper_ros module in the base_local_planner package, since the old one could not accept the odometry topic name other than "odom". Now in order to use the new version of the module, one must call the initialization function before using the module. Many other changes result actually from the indentation by the editor I use (qtcreator).


hershwg:
I don't have time to find which changes are real changes and which changes are just indentation changes.  I can't evaluate a pull request like this.

I want to accept your changes, they sound great, but I need to be able to look at the changes and see just the ones that fix the bug.

Thanks

mikegao88:
Sorry, I have opened up another pull request which is more clean, so this one can be closed.

hershwg:
I see it.  Thanks!

PR #110, opened 2013-08-15
Some more fixes for the pointcloud object bug and an improvement for the odometry_helper module

mikegao88:
I have tested this correction, and it works on my computer with ubuntu 12.04 32bit and ros-hydro. I reset everything, and now there are no indentation changes, and all the changes are in a different branch in order to keep the forked repository clean.


hershwg:
Excellent, thanks!

I'm going to change OdometryHelperRos::initialize() into OdometryHelperRos::setOdomTopic() so it is clearer and more flexible, but I'll keep the ability for the odom topic to be read from a parameter that you added to DWAPlannerROS.

Issue #112, opened 2013-08-16
amcl should publish pose and particle cloud as latched topics


piyushk:

hershwg:
I think it wouldn't hurt anything, since the pose and particle cloud both have proper timestamps in their headers.  If you want to make the change, test it, and submit a pull request, I'll merge it.

piyushk:
pull request opened. closing here.

PR #113, opened 2013-08-16
amcl_pose and particle cloud are now published latched 

piyushk:
modified for both amcl and fake_localization. closes https://github.com/ros-planning/navigation/issues/112


piyushk:
I've tested this in simulation. Seems to be working fine.

Issue #111, opened 2013-08-16
New indication method of the speed limits in dwa_local_planner package 


mikegao88:
In the dwa_local_planner package, the maximum translation speed is indicated by the user, but given the maximum velocities in x and y directions, is it more suitable to calculate the maximum translation speed of the robot with the right angle equation? I mean, like this:

limits.max_trans_vel = sqrt(limits.max_vel_x \* limits.max_vel_x + limits.max_vel_y \* limits.max_vel_y); (dwa_local_planner_ros.cpp)

Likewise, given the acceleration limits of the robot in x and y directions, the maximum acceleration speed is calculated in the same way rather than being indicated by the user:

limits.acc_limit_trans = sqrt(limits.acc_lim_x \* limits.acc_lim_x + limits.acc_lim_y \* limits.acc_lim_y); (dwa_local_planner_ros.cpp)

I hope some could discuss this little possible improvement with me. Thanks!

hershwg:
Well, I agree that max_trans_vel does seem redundant with some combination of max_vel_x and max_vel_y, but I disagree that it should be equal to hypot(max_x, max_y).  The space of velocities in x and y determined by max_vel_x and max_vel_y will be a rectangle.  If we set max_trans_vel to be hypot(x,y), we will have added a circular limit where the circle exactly circumscribes the previous rectangle.  So it won't have any affect on what is allowed.

Looking at the PR2 nav config, (https://github.com/PR2/pr2_navigation/blob/master/pr2_navigation_config/move_base/dwa_local_planner.yaml), I see that max_trans_vel is set to the max of max_vel_x and max_vel_y.

So I think it would be reasonable to make a change which:
- If max_trans_vel is specified in ros-params, use it
- If max_trans_vel is not specified, use max(max_vel_x, max_vel_y).

mikegao88:
Thanks for the interpretation! So should I take this correction and apply for a pull request?

hershwg:
yes. :)

PR #114, opened 2013-08-16
new max vel and acc limits indication methods

mikegao88:
The default values of the max_trans_vel and acc_limit_trans have been given in the cfg file, thus I have to revise the reconfigure file to realize the change as discussed.

Fix for #111

### DIFF
--- a/dwa_local_planner/src/dwa_planner_ros.cpp
+++ b/dwa_local_planner/src/dwa_planner_ros.cpp
@@ -63,18 +63,20 @@ namespace dwa_local_planner {
 
       // update generic local planner params
       base_local_planner::LocalPlannerLimits limits;
-      limits.max_trans_vel = config.max_trans_vel;
+      //limits.max_trans_vel = config.max_trans_vel;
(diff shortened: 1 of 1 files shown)


hershwg:
I need either a way to preserve backwards-compatibility for dwa_local_planner, like detecting if max_trans_vel has been set or not, OR an analysis of the code demonstrating that your change will make the robot act the same way that it would have before.

max_trans_vel is used in different places from max_vel_x and max_vel_y, so I'm not sure what would happen if one set max_vel_x = 1.0 and max_trans_vel = 2.0, for instance.  Would it ever try to go 2.0m/s?  Or would it change the slope of the footprint-scaling?

Overall I want to allow this change, but I want to be careful not to break other people's code.

Also, a note on style: you can just delete lines that you want to delete, you shouldn't leave them there but comment them out.  Pull requests show me the diff, so I know what is being removed.  The only place I leave code commented out is when I make code for doing debug-writes or similar, which I might want to re-activate at some point later.

mikeferguson:
More than year since last comment, closing.

PR #115, opened 2013-08-16
Remove unimplemented method from Costmap

DLu:
This screwed me up when I was writing some other code. But it was my own fault.


hershwg:
Ahh, the joys of c++.

PR #116, opened 2013-08-17
Fix bounds on clear costmap recovery

DLu:
Fixing two bugs in one:
Note that the actual order of parameters in setResetBounds is minx, maxx, miny, maxy, not x,y,x,y. 

Also, the new values reflect the actual scope of the map when the origin is not 0,0.


DLu:
There are also formatting issues with the method, which I will fix in a separate pull request.

hershwg:
Is the origin of the costmap really always in the center?  Is there any situation where it is at a corner?

DLu:
I'm not sure what I was thinking before. The updated calculation is what you would get calling the equivalent mapToWorld functions.

PR #120, opened 2013-08-27
Improvements to the static map layer to support updates

meyerj:
The current version of the static map layer does not work with dynamic updates as the costmap is not updated any more after the first map has been received. Although costmap_2d is primarily meant for static maps, there are also many use cases where the map is updated throughout the mission, e.g. for exploring unknown environments.

This pull requests contains the following updates:
- Fixed typo in `map_recieved_` variable
- Set `map_initialized_` to false in the map callback to enforce a costmap update
- Only call `layered_costmap_->resizeMap()` if the map parameters (size, resolution, origin) have changed. Otherwise the costmap would be fully reset for each incoming map.
- Changed log type for the "Received a ... map" message from INFO to DEBUG and added another one if the map parameters have changed (to not spam the console for each update).

### DIFF
--- a/costmap_2d/include/costmap_2d/static_layer.h
+++ b/costmap_2d/include/costmap_2d/static_layer.h
@@ -73,7 +73,7 @@ private:
   void reconfigureCB(costmap_2d::GenericPluginConfig &config, uint32_t level);
 
   std::string global_frame_; ///< @brief The global frame for the costmap
-  bool map_recieved_, map_initialized_;
+  bool map_received_, map_initialized_;
   bool track_unknown_space_;
   ros::Subscriber map_sub_;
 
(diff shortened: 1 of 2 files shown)


DLu:
Looks great. Can you add one ROS_INFO "Received Map" line into the constructor, just so that it will announce when it has stopped waiting for a map? 

P.S. Calling it a typo is generous. I just can't spell.

meyerj:
We have been using costmap_2d with dynamically updated maps from [hector_slam](http://www.ros.org/wiki/hector_slam) since electric with approx. one new map per second. It's probably not the most efficient thing to do, but it works.

For the long run, wouldn't it be possible to rename the static_layer to just map_layer and add support for partial map updates using the new `map_msgs/OccupancyGridUpdate` topic?

PR #134, opened 2013-09-29
Checkins to reflect the end of Willow Garage

DLu:
- Changes maintainer to non-Willow email address
- Changes download location for test data

### DIFF
--- a/amcl/CMakeLists.txt
+++ b/amcl/CMakeLists.txt
@@ -76,41 +76,41 @@ if(CATKIN_ENABLE_TESTING)
 
   # Bags
   catkin_download_test_data(${PROJECT_NAME}_basic_localization_stage_indexed.bag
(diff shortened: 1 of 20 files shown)


dirk-thomas:
Since it seems that you have taken over maintainance of navigation it would be great if you could set that status officially (for Groovy and Hydro) as described here: http://wiki.ros.org/MaintenanceGuide#Claiming_Maintainership

DLu:
Why not harvest this information from package.xml?

dirk-thomas:
Simply because the maintenance status is not embedded in the package.xml file.

It only contains the name and email of a the maintainer. A package can come form any location and nobody might even have write access to it when the maintainer stops to respond. Therefore this information needs to be located in a central location which is accessible for everyone to contribute.

PR #136, opened 2013-10-14
check for CATKIN_ENABLE_TESTING

bulwahn:
Since version 0.5.68, catkin provides to optionally configure tests. This commit adjusts the CMakeLists.txt to opt-in the tests.

### DIFF
--- a/base_local_planner/CMakeLists.txt
+++ b/base_local_planner/CMakeLists.txt
@@ -125,17 +125,19 @@ install(DIRECTORY include/${PROJECT_NAME}/
   PATTERN ".svn" EXCLUDE
 )
 
-catkin_add_gtest(base_local_planner_utest
-	test/gtest_main.cpp
-	test/utest.cpp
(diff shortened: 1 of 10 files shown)


DLu:
What version of catkin did hydro ship with?

mikeferguson:
Hydro is already up to 0.5.74

DLu:
What did it start with?

mikeferguson:
Not sure, but 0.5.68 is >3mo old, which certainly means it was available during the beta period. The CATKIN_ENABLE_TESTING thing is specifically called out as hydro-specific fix to implement in the maintainers guide: http://wiki.ros.org/MaintenanceGuide

bulwahn:
The "first" official release of hydro had version 0.5.73. The CATKIN_ENABLE_TESTING flag was added with 0.5.68 during the beta period (or even shortly before that). So, every version of a hydro distribution on any machine will support the CATKIN_ENABLE_TESTING flag.

PR #139, opened 2013-10-17
robot_pose_ekf absolute sensor incorporation update to hydro-devel

pmukherj:
-Fixing some covariance bugs
-Adding the ability to incorporate absolute IMU/compass values into the filter
-Bringing rostest of gps code forward to hydro_devel from navigation-1.8. 

Two other quick notes:
-I made the test bags a dependency, so they are always downloaded when the test is run. Not sure if there are any objections to that
-The gps test bag will have to be moved from pr.willowgarage.com to download.ros.org.


pmukherj:
Closing deprecated pull request. To be re-based and resubmitted.

PR #145, opened 2013-11-03
Bugfixes

contradict:
Two small things that look like bugs to me, and one more ignore.


acornacorn:
last 2 look fine.  Assign to Hersh to check the "Keep consistent allignment_cost scale" change.

hersh:
Hi @contradict.

I see that the change in 6a72c61 looks like it would be consistent with the code earlier in the file, but before I merge this, I need to know that a real robot has run with the old code and with the new code and that it works better or at least not worse than before.

The robot navigation behavior in my experience is very very sensitive to changes in this code, so even if it looks more correct one way, the only real way to evaluate is whether robots move better.  I'm sorry I didn't have time at Willow to get a comprehensive regression test suite together, so I can't just run an automated test.

What robot do you have?  Diff-drive or holonomic?

DLu:
I have seen this issue as well. See #147 . Can you split this into two pull requests, one for the two easy fixes, and one for #147 ?

contradict:
I used it for this robot:

http://www.flickr.com/photos/nasahqphoto/8985027541/in/set-72157634001929773

It only have qualitative data, but these fixes produce better behaviour on our robot. I agree that it is very sensitive and would be happy to split the patch set up.

contradict:
OK, two new pull requests #154 and #153  filed, feel free to close this one. #154 is definitely better than this one since it merges cleanly with the current state.

HWiese1980:
I don't know what it is but it's still happening to my (simulated) robot. I'm not able to test it on a real robot because we don't have one yet. But that shouldn't make a difference, should it?

I pulled the latest hydro-devel branch and recompiled the whole navigation stack. And it still happens. Do I have to configure something to see your bug fix working? Or should it work out of the box?

PR #150, opened 2013-11-13
Add a first_map_only parameter so we keep reusing the first received static map.

corot:
We need this for a rather exotic case: when running our robots inside a concert (https://github.com/robotics-in-concert), the first time we subscribe we get the map, but not in subsequent subscriptions (due to clearing costmaps) because we didn't unsubscribe.
An easier solution is just to call shutdown in the map subscriber, but we still have a big delay (maps is shared in a server) until receiving the map, so this flag is much better option.
I didn't verify that it doesn't interfere with map updates (mostly because is the first time I see this feature!), but apart from this, we have this change running happily in our robots.


trainman419:
It looks like you've included a bunch of irrelevant changes to the inflation layer and fake_localization in this PR. It will be easier to review and accept if you split those changes out into separate pull requests.

It sounds like what you really want is for the reset on the static layer to reset it to the most recently received map, rather than the first map that was received; this should handle dynamic maps better. If that's a no-op, then you may want to look into what calls the reset function, and make it not reset the static costmap at all.

corot:
Yeah, stupid me I kept adding changes to the same branch. I will isolate the useful patch as soon as I find time.

PR #154, opened 2013-11-21
Keep consistent allignment_cost scale

contradict:
Use the configured alignment cost instead of resetting to 1.0. This
condition (farther from goal than forward_point_distance) is probably
met at the start of navigation, it seems this cost should be obeyed
until the robot gets close, then ignored completely.


hersh:
I'm going ahead and merging this.  @contradict said this works better on his robot, which appears to be (essentially) holonomic based on the image he linked in #145.  It looks like a similar wheel arrangement to the PR2.

Since it has been split out into a single change like this it should be easy to revert if it breaks the behavior for a bunch of other people.

PR #155, opened 2013-12-02
Install crop_map python script.

jihoonl:


DLu:
RE: Issue #3

jihoonl:
@DLu  is crop_map.py legacy code to be removed?

DLu:
I haven't really looked through this yet, but wanted to link the relevant issue.

DLu:
Thanks for this!

PR #157, opened 2013-12-11
navfn: fix parallel build error from missing dep

wjwwood:
When building I would get errors because the messages would not generate their headers before the navfn library build was started.

### DIFF
--- a/base_local_planner/CMakeLists.txt
+++ b/base_local_planner/CMakeLists.txt
@@ -71,7 +71,7 @@ catkin_package(
 #set(ROS_COMPILE_FLAGS "-g" ${ROS_COMPILE_FLAGS})
 #set(ROS_LINK_FLAGS "-g" ${ROS_LINK_FLAGS})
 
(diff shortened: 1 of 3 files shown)


wjwwood:
@dirk-thomas can you review this for a merge? It is blocking building this from source on the first try.

wjwwood:
Also @DLu (`@` completion messed up...).

dirk-thomas:
+1

DLu:
@DLu is not qualified enough to have an opinion.

mikeferguson:
+1 from me as well

wjwwood:
@DLu I was asking you as the maintainer :smile::

https://github.com/ros-planning/navigation/blob/hydro-devel/navfn/package.xml#L12

It seems that the patch is good, do you want to merge it and release navigation (at some point)?

I wanted to give you the opportunity to do so, but if you don't want to or don't have time I can do it for you.

PR #161, opened 2014-01-22
global_planner: Fix bgp_plugin.xml file and install it

RainCT:


martimorta:
Great!
This fixes the problem of the planner not being registered.
Thanks @RainCT

PR #162, opened 2014-01-23
Reversed the resulting plan vector in global_planner

martimorta:
I bumped in a problem that the robot couldn't move, it is detailed here:
http://answers.ros.org/question/120786/global_planner-finds-a-path-but-is-not-able-to-find-a-valid-trajectory

I found out that the plan sent the path points from goal to origin, so I reversed the plan vector to fix it.
I hope there's a better solution but this one works.


DLu:
Thanks for the fix!

PR #171, opened 2014-02-12
inflation_layer: place .top() & .pop() calls together

RainCT:
Hi,

Just some little cleanup. Feel free to ignore it (but let me know, so I can delete the branch).

The code was calling .pop() after inserting additional elements to the queue. This isn't a problem here, since the new elements should have a higher cost, but I personally find it cleaner if top&pop are always together.

### DIFF
--- a/costmap_2d/plugins/inflation_layer.cpp
+++ b/costmap_2d/plugins/inflation_layer.cpp
@@ -152,6 +152,7 @@ void InflationLayer::updateCosts(costmap_2d::Costmap2D& master_grid, int min_i, 
   {
     //get the highest priority cell and pop it off the priority queue
     const CellData& current_cell = inflation_queue_.top();
+    inflation_queue_.pop();
 
     unsigned int index = current_cell.index_;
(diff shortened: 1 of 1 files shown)


trainman419:
:+1:

mikeferguson:
So, it turns out, this breaks several tests, since we pop something before we pull the data from a reference to it (the following lines). These were now failing on the buildfarm:
- http://jenkins.ros.org/job/devel-hydro-navigation/ARCH_PARAM=amd64,UBUNTU_PARAM=precise,label=devel/3/testReport/junit/(root)/costmap/testInflation3/ 
- http://jenkins.ros.org/job/devel-hydro-navigation/ARCH_PARAM=amd64,UBUNTU_PARAM=precise,label=devel/3/testReport/junit/(root)/costmap/testInflation/

I've pushed a small patch: https://github.com/ros-planning/navigation/commit/8d031e5407797c7243ef47ea2b351ff0371dd95e that fixes these tests on my local machine.

PR #174, opened 2014-02-22
Fixes unclean costmap2d shutdown

adamantivm:
This change in the order of the declaration of subscriber and message filter resolves an issue we were seeing that caused costmap2d to crash or hang when shutting down.

The cause for the issue was that the message filter was being destructed first and then when the subscriber destructor kicked-in, it would try to unregister from the message filter which triggered a lock on a deleted mutex.

With this minor change now shutdown occurs in the proper order, subscriber is first unregistered when destructed and then the message filter can die and rest in peace.

### DIFF
--- a/costmap_2d/include/costmap_2d/obstacle_layer.h
+++ b/costmap_2d/include/costmap_2d/obstacle_layer.h
@@ -158,8 +158,8 @@ protected:
 
   laser_geometry::LaserProjection projector_; ///< @brief Used to project laser scans into point clouds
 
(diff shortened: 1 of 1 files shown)


jonbinney:
+1

DLu:
Thanks for figuring that out!

adamantivm:
It wasn't without sweat and frustration, but I'm happy it helps!

On Mon, Feb 24, 2014 at 4:05 PM, David Lu!! notifications@github.comwrote:

> Thanks for figuring that out!
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/ros-planning/navigation/pull/174#issuecomment-35959244
> .

## 

-- Julian Cerruti

PR #176, opened 2014-02-28
add parameter to configure whether full costmap is published each time

paulbovbel:
useful for testing when you want to pull in the entire costmap each update


mikeferguson:
Seems useful, should probably be added to the wiki with a note that it is new in 1.11.7.

PR #177, opened 2014-03-02
Only publish costmaps if they have subscribers.

trainman419:
This shaves off a few cycles and improves performance.


DLu:
Thanks Austin!

PR #179, opened 2014-03-31
fix problem that isGoalReached checked goal without updating current_pose_

YoheiKakiuchi:
This patch is a part of 
https://github.com/ros-planning/navigation/commit/97c2c7844cc00b489d3daa21b505fe7f080e6d5b

isGoalReached returned true, before updating current_pose.

### DIFF
--- a/dwa_local_planner/src/dwa_planner_ros.cpp
+++ b/dwa_local_planner/src/dwa_planner_ros.cpp
@@ -141,6 +141,10 @@ namespace dwa_local_planner {
       ROS_ERROR("This planner has not been initialized, please call initialize() before using this planner");
       return false;
     }
+    if ( ! costmap_ros_->getRobotPose(current_pose_)) {
+      ROS_ERROR("Could not get robot pose");
+      return false;
+    }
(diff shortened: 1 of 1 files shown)


mikeferguson:
LatchedStopRotateController was previously broken (fixed in #193) -- could you take a look at that and let me know if this fix is still actually needed?

efernandez:
@YoheiKakiuchi @mikeferguson in #193 only the param is fixed. IMHO here the patch is completely un related, i.e. #193 activates https://github.com/efernandez/navigation/blob/5bdb3de8c62e39c43522da120183a18961d4ea9d/base_local_planner/src/latched_stop_rotate_controller.cpp#L87, while this PR affects https://github.com/efernandez/navigation/blob/5bdb3de8c62e39c43522da120183a18961d4ea9d/base_local_planner/src/latched_stop_rotate_controller.cpp#L88, but I don't know whether is really important to retrieve `current_pose` right before calling the method, but at least it seems to be in new versions, as shown in 97c2c78, mentioned by @YoheiKakiuchi

DLu:
I have this same change in a different branch. Thanks for the patch.

PR #182, opened 2014-04-04
Created a new parameter to move_base: 'num_spinner_threads'.

tpanzarella:
Depending upon the value passed to this new parameter, move_base will start up
and run with a single-threaded spinner (current behavior) or if passed a value
of 2 (threads) or greater, move_base will start a multi-threaded spinner (using
ros::AsyncSpinner) with the desired number of threads.

This behavior was created to address a latency issue that is described here:
http://answers.ros.org/question/148171/latency-when-sending-goal-to-move_base-and-it-starting/

I'm not sure of your testing procedure for integrating changes into the project nor move_base's overall thread safety. None-the-less, I thought I'd pass this along to you for review. I'm more than happy to follow proper protocol to get this patch (or similar) integrated to the master in order to have this functionality available in the core ROS move_base.  Thanks, Tom.

### DIFF
--- a/move_base/src/move_base_node.cpp
+++ b/move_base/src/move_base_node.cpp
@@ -29,14 +29,27 @@
 
 #include <move_base/move_base.h>
 
-int main(int argc, char** argv){
+int main(int argc, char** argv) {
   ros::init(argc, argv, "move_base_node");
   tf::TransformListener tf(ros::Duration(10));
 
-  move_base::MoveBase move_base( tf );
+  move_base::MoveBase move_base(tf);
 
-  //ros::MultiThreadedSpinner s;
-  ros::spin();
+  int num_spinner_threads = 1;
+  ros::NodeHandle private_nh("~");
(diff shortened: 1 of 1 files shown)


mikeferguson:
It is really too late to put such a change into hydro -- especially given the number of mutex errors we've recently found in costmaps. If someone wants to open a PR to incorporate this into jade, please do so.

For reference, there are currently 3 threads in move_base -- the main thread which will execute all the sensor callbacks, a thread which is started by the action server (which will also then be executing the executeCb) and the planner thread. The planner thread handles no callback queues, the action server thread processes only its own queue, and the main thread processes the global queue. For a time there was a major bug with the simple action server not processing the right queue, but that was fixed with  https://github.com/ros/actionlib/commit/6c7387669a1544d89144814b144a63113906038b -- I don't think this bug was present in hydro, but maybe, as I am quite surprised that the executeCb would be held up unless extra threads were handed to it.

PR #183, opened 2014-04-10
Fix bug in distance calculation for trajectory rollout

RainCT:
The parentheses for the square root in the distance calculation were wrong.

I suppose this is the reason why the "2 *" seemed necessary in the line below. However, I didn't test it.


mikeferguson:
+1

mikeferguson:
#184 includes this. We'll merge that one.

PR #190, opened 2014-05-02
uses odom child_frame_id to set robot_vel frame_id

efernandez:
If, instead, we want to take the `robot_frame_id`, we must pass the `costmap_ros_` to take it with

<pre>
costmap_ros_->getBaseFrameID();
</pre>

In that case, should I pass the `costmap_ros_` to the constructor?
Is there any plan or idea to do it?

### DIFF
--- a/base_local_planner/src/odometry_helper_ros.cpp
+++ b/base_local_planner/src/odometry_helper_ros.cpp
@@ -69,9 +69,10 @@ void OdometryHelperRos::getRobotVel(tf::Stamped<tf::Pose>& robot_vel) {
     global_vel.linear.x = base_odom_.twist.twist.linear.x;
     global_vel.linear.y = base_odom_.twist.twist.linear.y;
     global_vel.angular.z = base_odom_.twist.twist.angular.z;
+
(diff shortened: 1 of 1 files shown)


mikeferguson:
So it appears that frame_id_ has never been set in the past? Does this cause a change in behavior?

efernandez:
@mikeferguson You're right, `frame_id_` wasn't set at all.
I don't think this commit make any change: the `child_frame_id` is set, as it should, but only the velocity fields are really used by the planner.

All in all, IMHO this PR simply sets the `child_frame_id` to a valid value, while before it wasn't set to anything. Still, if you want, we might pass the `costmap_ros_` to take the `robot_frame_id` with

<pre>
costmap_ros_->getBaseFrameID();
</pre>

but I'm not sure it must be that way.

mikeferguson:
@DLu -- might you have any thoughts on this?

DLu:
I don't see anything wrong, but am not an expert on the inner workings of this class. +1

mikeferguson:
Ok, this passes all tests, and does seem to be at least _more_ correct.

PR #191, opened 2014-05-02
removes useless this->z_max = z_max assignment

efernandez:


mikeferguson:
Ah.. to see the change in context, I clicked through to view the whole file, and then saw this only existing in the SetModelBeam() function. Yes, this is correct, totally unneeded.

PR #193, opened 2014-05-02
fixes latch_xy_goal_tolerance param not taken

efernandez:
The `LatchedStopRotateController` does NOT get the `latch_xy_goal_tolerance` param.
Since the default value is `false`, it does nothing!


efernandez:
I've just updated the description of this PR.
Please, could somebody review this?
IMHO, this must be merged, because the `latch_xy_goal_tolerance` functionality is broken.

mikeferguson:
@efernandez I haven't gotten a chance to look at this in detail yet, but we will eventually merge it. The one thing I want to look into before merging is whether we should also be searching up the namespace when looking for the parameter.

mikeferguson:
Ok, so this does in fact replicate exactly what is present in the older trajectory_planner_ros, no need to search params then. LGTM.

PR #198, opened 2014-05-31
Fixed costmap_2d clearing from service /move_base/clear_costmaps

kmhallen:
Calling the service /move_base/clear_costmaps had little effect in both Hydro and Indigo.

### DIFF
--- a/costmap_2d/src/costmap_2d_ros.cpp
+++ b/costmap_2d/src/costmap_2d_ros.cpp
@@ -639,6 +639,8 @@ void Costmap2DROS::resume()
 
 void Costmap2DROS::resetLayers()
 {
+  Costmap2D* top = layered_costmap_->getCostmap();
+  top->resetMap(0, 0, top->getSizeInCellsX(), top->getSizeInCellsY());
   std::vector < boost::shared_ptr<Layer> > *plugins = layered_costmap_->getPlugins();
   for (vector<boost::shared_ptr<Layer> >::iterator plugin = plugins->begin(); plugin != plugins->end();
       ++plugin)


mikeferguson:
LGTM -- one question that I have is: should we actually clear the plugins (layers) first? And then clear the costmap? There really shouldn't be multiple threads servicing the update/service right now, but if you have a non-clearing layer that actually updates the costmap in between the clearing of the costmap and the clearing of the plugin/layer, you would still probably have a problem.

efernandez:
BTW With `hydro` (1.11.8) version `move_base` crashes after calling `rosservice call /move_base/clear_costmaps`; usually it happens after 2-3 consecutive calls.
It looks like a threading issue.
@kmhallen Does this PR solve that issue?

kmhallen:
@efernandez I have not noticed crashes with that service, and I usually call the service 2-3 times for good measure. (1.11.8 hydro and indigo)

efernandez:
@kmhallen I run on 32bit, what about you?
I'll try to debug it when I have some time... I can put a backtrace here if you don't mind.

kmhallen:
@efernandez 
Below are the systems I have tested on:
Ubuntu 14.04 64bit Indigo
Ubuntu 13.04 64bit Hydro

I won't be able to help too much with a backtrace. I only started looking into the source code when I found the clearing problem.

mikeferguson:
Could also be what this user is reporting: http://answers.ros.org/question/172948/move_base-crashed-when-call-clear_costmaps-service/

efernandez:
@mikeferguson I don't call it at high frequency, just manually with quite big costmaps.
Indeed, I've just tried to reproduce it, and it's enough with one call.
The backtrace is:

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0xaf2bfb40 (LWP 2737)]
__memset_sse2_rep () at ../sysdeps/i386/i686/multiarch/memset-sse2-rep.S:299
299 ../sysdeps/i386/i686/multiarch/memset-sse2-rep.S: No such file or directory.
(gdb) bt
#0  __memset_sse2_rep () at ../sysdeps/i386/i686/multiarch/memset-sse2-rep.S:299
#1  0xb0f6b505 in costmap_2d::InflationLayer::updateCosts(costmap_2d::Costmap2D&, int, int, int, int) ()
   from /opt/ros/hydro/lib/liblayers.so
#2  0xb7a3f332 in costmap_2d::LayeredCostmap::updateMap(double, double, double) () from /opt/ros/hydro/lib/libcostmap_2d.so
#3  0xb7a425dd in costmap_2d::Costmap2DROS::mapUpdateLoop(double) () from /opt/ros/hydro/lib/libcostmap_2d.so
#4  0xb7a49dd5 in boost::detail::thread_data<boost::_bi::bind_t<void, boost::_mfi::mf1<void, costmap_2d::Costmap2DROS, double>, boost::_bi::list2<boost::_bi::value<costmap_2d::Costmap2DROS*>, boost::_bi::value<double> > > >::run() () from /opt/ros/hydro/lib/libcostmap_2d.so
#5  0xb792f48c in thread_proxy () from /usr/lib/libboost_thread.so.1.46.1
#6  0xb7910d4c in start_thread (arg=0xaf2bfb40) at pthread_create.c:308
#7  0xb7b89bae in clone () at ../sysdeps/unix/sysv/linux/i386/clone.S:130
(gdb)
```

efernandez:
It seems that the offending line is at `costmap_2d` in `plugins/inflation_layer.cpp`:

``` cpp
memset(seen_, false, size_x * size_y * sizeof(bool));
```

efernandez:
This doesn't happen with latest `hydro-devel`; I'm using debians from version `1.11.4` (`ros-hydro-*`, like `base-local-planner`, `move_base`, etc.) and `ros-hydro-navigation 1.11.8`.
I think I should use the same for all of them, no? Indeed, this mean I was using `1.11.4`, not `1.11.8` (`ros-hydro-navigation` is just a metapackage).
And probably, something has fix my issue since `1.11.4` (there has been a lot of commits).
Sorry for the "noise".

efernandez:
BTW With `1.11.8` the `rosservice call /move_base/clear_costmaps` works perfect for me.
@kmhallen Which or how many layers are you using?

kmhallen:
Using unmodified code, it looks like the clear_costmaps service works as expected if there is a static map. It leaves the static map and clears everything else. Without a static map, nothing happens. This means the local costmap is never cleared. Tested with 1.11.8 and 1.11.9 debians for Hydro on Ubuntu 13.04.

The following is my configuration. I also tried combining all of the ObstacleLayers into one.

```
global_costmap:
  rolling_window: false
  update_frequency: 5.0
  publish_frequency: 0.5
  width: 102.0
  height: 102.0
  resolution: 0.1
  origin_x: -51.0
  origin_y: -51.0
  plugins: []
  plugins: 
      - {name: static_map,  type: "costmap_2d::StaticLayer"}
      - {name: laser,       type: "costmap_2d::ObstacleLayer"}
      - {name: lines,       type: "costmap_2d::ObstacleLayer"}
      - {name: flags,       type: "costmap_2d::ObstacleLayer"}
      - {name: inflation,   type: "costmap_2d::InflationLayer"}
local_costmap:
  rolling_window: true
  update_frequency: 25.0
  publish_frequency: 5.0
  width: 15.0
  height: 15.0
  resolution: 0.05
  origin_x: 0.0
  origin_y: 0.0
  plugins: []
  plugins: 
  #    - {name: static_map,  type: "costmap_2d::StaticLayer"}
      - {name: laser,       type: "costmap_2d::ObstacleLayer"}
      - {name: lines,       type: "costmap_2d::ObstacleLayer"}
      - {name: flags,       type: "costmap_2d::ObstacleLayer"}
      - {name: inflation,   type: "costmap_2d::InflationLayer"}
```

DLu:
One quirk that clear_costmaps has with layered costmaps is that it must know which layers to clear. This is badly documented, I concede. The current behavior is to clear all layers with a given substring, set by [this parameter](https://github.com/ros-planning/navigation/blob/hydro-devel/clear_costmap_recovery/src/clear_costmap_recovery.cpp#L62). However, as discussed in #211, this is a limited solution. I have begun work on a more general solution where you can specify a list of layers to clear in [this branch](https://github.com/DLu/navigation/tree/clear_recovery_array), which I will push once I test it a bit further.

efernandez:
@kmhallen I've put `obstacle_` to my obstacle layers in both local and global costmaps, as suggested by @DLu . Tested in `1.11.8` and both are cleared by the recovery behaviors; before that nothing was cleared by them, but manually I was able to do it (calling the service from the terminal).

efernandez:
@kmhallen After testing a little more, I've seen that the local costmap is NOT cleared, as you mentioned. It has an obstacle_ layer, but not static_layer (as it's normal for the local costmap).

mikeferguson:
Ok, the comments by @efernandez have got me thinking now. And I actually think this may be a bigger bug than just the clear service.

Basically, each layer maintains it's own internal costmap. Then each time the layered_costmap is updated (https://github.com/ros-planning/navigation/blob/hydro-devel/costmap_2d/src/layered_costmap.cpp#L78), it should clear the costmap stored by the layered_costmap (line 115) and then update it using each of the plugin layers (line 122). Thus, clearing the layers should be enough to make costmap clearing work, but it appears not to be so.

With the static_layer, we use UpdateWithTrueOverride() to update the costmap from the layer. This has the effect of copying the entire layer onto the costmap, even if a cell has NO_INFORMATION (thus, if you clear the static_layer, the next time the layer updates the costmap, it will definitely clears). With the local costmap, there isn't any such layer, the obstacle/voxel layers are either UpdateWithOverride() (which won't update a cell if it has NO_INFORMATION in the layer) or UpdateWithMax().

This then raises two questions in my mind:
- Why doesn't line 115 in layered_costmap.cpp do the same thing as this PR? @DLu what is up with with the logic of updateMap() -- why are the min/max set to +/-1e30 when calling plugin->updateBounds?
- Should an obstacle/voxel layer be able to do UpdateWithTrueOverride, particularly when used as the base layer of a local costmap?

DLu:
@mikeferguson 
Line 115 sets the area that is about to be updated to the default value, which is set up on line 52. 

The max/min set to +/- 1e30 is a bad hack. Should technically be DoubleMin/DoubleMax. It represents an invalid bounding box, so that the first layer to update the bounding box will change the minimum values to something guaranteed to be smaller than +1e30. 

The only difference between UpdateWithOverride and updateWithTrueOverride is how it handles unknown values in an individual layer. If the master costmap has NO_INFORMATION, then updateWithOverride will have no effect. If the master has FREE_SPACE, then track_unknown must have been false, and as a result, the unknown value will not be written into the costmap.

mikeferguson:
RE: +/-1e30: @DLu that makes a bit more sense -- updateBounds (in layer.h) should really have a doxygen comment about what it should do, because that isn't documented anywhere else it seems.

mikeferguson:
@DLu is this addressed by #230?

DLu:
I think the origin of this issue was specifying which layers to clear, so not an issue in the actual code, just @kmhallen's configuration and my documentation. 

@kmhallen  Have you gotten your costmaps to clear yet? 

We may want to open a separate issue for repeatedly calling the service.

kmhallen:
With a StaticLayer, clearing works by initializing the map to the static map.

mikeferguson:
Ok, so this is still needed for costmaps lacking static layer. I guess the other approach in the future would be to make the first layer in the stack use UpdateWithTrueOverride(), but that could make things difficult to debug.

PR #199, opened 2014-05-31
No need to check `range > 0` for infinity or NaN

efernandez:


mikeferguson:
See REP117 (http://www.ros.org/reps/rep-0117.html). -inf has a different meaning than +inf, and so this code is correct.

efernandez:
Ok, thanks for clarifying it.

PR #201, opened 2014-06-03
Use hypot(x, y) instead of sqrt(x*x, y*y)

efernandez:
This follows #184

`amcl` not covered because it fails to find `#include <cmath>`.

### DIFF
--- a/base_local_planner/src/goal_functions.cpp
+++ b/base_local_planner/src/goal_functions.cpp
@@ -39,10 +39,7 @@
 namespace base_local_planner {
 
   double getGoalPositionDistance(const tf::Stamped<tf::Pose>& global_pose, double goal_x, double goal_y) {
-    double dist =
-        (goal_x - global_pose.getOrigin().x()) * (goal_x - global_pose.getOrigin().x()) +
-        (goal_y - global_pose.getOrigin().y()) * (goal_y - global_pose.getOrigin().y());
-    return sqrt(dist);
+    return ::hypot(goal_x - global_pose.getOrigin().x(), goal_y - global_pose.getOrigin().y());
   }
 
   double getGoalOrientationAngleDifference(const tf::Stamped<tf::Pose>& global_pose, double goal_th) {
--- a/base_local_planner/src/obstacle_cost_function.cpp
+++ b/base_local_planner/src/obstacle_cost_function.cpp
@@ -100,7 +100,7 @@ double ObstacleCostFunction::scoreTrajectory(Trajectory &traj) {
 }
 
 double ObstacleCostFunction::getScalingFactor(Trajectory &traj, double scaling_speed, double max_trans_vel, double max_scaling_factor) {
-  double vmag = sqrt(traj.xv_ * traj.xv_ + traj.yv_ * traj.yv_);
+  double vmag = ::hypot(traj.xv_, traj.yv_);
 
   //if we're over a certain speed threshold, we'll scale the robot's
   //footprint to make it either slow down or stay further from walls
--- a/costmap_2d/include/costmap_2d/costmap_2d.h
+++ b/costmap_2d/include/costmap_2d/costmap_2d.h
@@ -370,7 +370,7 @@ protected:
       unsigned int offset = y0 * size_x_ + x0;
 
       //we need to chose how much to scale our dominant dimension, based on the maximum length of the line
-      double dist = sqrt((x0 - x1) * (x0 - x1) + (y0 - y1) * (y0 - y1));
+      double dist = ::hypot(x0 - x1, y0 - y1);
       double scale = std::min(1.0, max_length / dist);
 
       //if x is dominant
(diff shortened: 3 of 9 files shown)


mikeferguson:
@efernandez This is breaking tests: http://jenkins.ros.org/job/devel-indigo-navigation/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/4/testReport/

mikeferguson:
@efernandez Hydro seems to have even more failed tests: http://jenkins.ros.org/job/devel-hydro-navigation/16/testReport/

efernandez:
@mikeferguson I've been investigating this with @bmagyar, looking into the `cmath` library. The tests also fail in my machine (sorry for not testing). Apparently, `hypot(x, y)` and `sqrt(x*x, y*y)` are not completely equivalent in practice, because of the numeric precision. In theory it should be the same (http://en.cppreference.com/w/cpp/numeric/math/hypot), but I'm not a 100% sure because of two factors:
- At the end, the `hypot` code is different to `sqrt(x*x, y*y)`, although I haven't seen the code; this is all that I've seen from Boost (http://www.boost.org/doc/libs/1_34_1/doc/html/boost_math/math_special_functions.html#boost_math.math_special_functions.hypot), which I think it was moved to the standard. There's also this reference pseudo-code: http://en.wikipedia.org/wiki/Hypot#Implementation, and some articles covering how to implement `sqrt(x*x + y*y)` safely: http://www.johndcook.com/blog/2010/06/02/whats-so-hard-about-finding-a-hypotenuse/
- It might be a bug (I don't think so) or a setting (it could be) that makes `hypot` use `long double` even for `double` input arguments.

I see two options here:
- Revert this PR
- Update the tests (if they're right) with the new results using `hypot`.

efernandez:
BTW, there were 4 cases were I used `hypot` instead of `hypotf`, which should be the replacement for `sqrtf`. However, I've tried that and the tests still fail; I tested with `rostest costmap_2d obstacle_tests.launch` only.

mikeferguson:
I actually think the tests are suspect. In particular https://github.com/ros-planning/navigation/blob/hydro-devel/costmap_2d/test/obstacle_tests.cpp#L90 should set the diagonal to be LETHAL (254), right?

This is the final map (i added printmap at L95) under the old code:

```
   0   0   0   0   0   0   0 254 254 254

   0   0   0   0   0   0   0 254 254 254

   0   0   0 254 254 254   0   0   0   0

   0   0   0   0   0   0   0   0   0   0

   0   0   0   0   0   0   0   0   0   0

   0   0   0   0 254   0   0 254 254 254

   0   0   0   0 254   0   0 254 254 254

   0   0   0   0   0   0   0 254 254 254

   0   0   0   0   0   0   0   0   0   0

   0   0   0   0   0   0   0   0   0 254
```

0,0 is the upper left, 9,9 is lower right. And the new code looks like:

```
   0   0   0   0   0   0   0 254 254 254

   0 254   0   0   0   0   0 254 254 254

   0   0 254 254 254 254   0   0   0   0

   0   0   0 254   0   0   0   0   0   0

   0   0   0   0 254   0   0   0   0   0

   0   0   0   0 254 254   0 254 254 254

   0   0   0   0 254   0 254 254 254 254

   0   0   0   0   0   0   0 254 254 254

   0   0   0   0   0   0   0   0 254   0

   0   0   0   0   0   0   0   0   0 254
```

I have no idea why the old code doesn't fill in the diagonal, nor why the new code does, nor has this test been changed in quite some time.

@DLu any ideas? You're the author on these tests it looks like.

efernandez:
@mikeferguson @DLu For me this test also looks strange; I mean, I don't see why `with_static` should be equal to `obs_after`, if the diagonal is filled in with new `LETHAL` values.
Furthermore, shouldn't the cell `<0, 0>` be marked as `LETHAL (254)`?
I don't really understand what the `layers.updateMap(0,0,0);` instruction does?

mikeferguson:
Anybody have any further thoughts on this? We should probably put a release out to include #202, but I'm not sure if we should revert this, or actually need to fix the tests (and can do that later).

mikeferguson:
@efernandez Tracked this down. PR to fix is here: https://github.com/ros-planning/navigation/pull/204 I'm going to wait until the current devel test finishes before merging so that I can get separate tests on the farm. Basically, the issue here was that subtraction of unsigned numbers was causing the hypot to be wrong. I added static_casts to force everything to doubles.

I've also added numerous comments in the tests, which turned out be correct, but the logic of why was fairly involved.

efernandez:
Nice catch @mikeferguson !

DLu:
What is the logic behind the ::hypot notation? To me the two colons seem clunky.

efernandez:
@DLu I guess ``std::hypot` failed with my compiler... but I've just tried in another computer and it works.
Therefore, I agree on using `std::hypot` instead: here and also for #184

PR #202, opened 2014-06-09
Fixes uninitialised variable 'combination_method'

### LINKED ISSUE #187: static layer sometimes doesn't use map to create the global costmap
Sometimes, the static layer doesn't create the global costmap properly, creating only an empty costmap. 

To exemplify, this is the map being published:

screenshot from 2014-04-28 17 44 24

This is the global costmap when the static layer doesn't load the map properly

screenshot from 2014-04-28 17 44 51

If you try to relaunch, it loads the map properly, and you have the following global costmap:

screenshot from 2014-04-28 17 45 09

This problem seems to only appear the first time you try to launch navigation, after that it looks like it always creates the global costmap properly

stonier:
This will close #187, refer to the discussion there for details.

### DIFF
--- a/costmap_2d/cfg/VoxelPlugin.cfg
+++ b/costmap_2d/cfg/VoxelPlugin.cfg
@@ -12,4 +12,10 @@ gen.add("z_voxels", int_t, 0, "The number of voxels to in each vertical column."
 gen.add("unknown_threshold", int_t, 0, 'The number of unknown cells allowed in a column considered to be known', 15, 0, 16)
 gen.add("mark_threshold", int_t, 0, 'The maximum number of marked cells allowed in a column considered to be free', 0, 0, 16)
 
+combo_enum = gen.enum([ gen.const("Overwrite", int_t, 0, "b"),
+                        gen.const("Maximum",   int_t, 1, "a") ],
+                      "Method for combining layers enum")
+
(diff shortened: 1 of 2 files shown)


mikeferguson:
@stonier Does combination_method == 0 actually work? It isn't clear from the other ticket if == 0 works at all. If it doesn't work, we probably should not even expose it in dynamic reconfigure for now.

stonier:
`combination_method == 0` is a complete failure. The obstacle (voxel) layer coming in on the first iteration will completely write over the whole map with zeroes everywhere except for locally sensed obstacles, resulting in what @BFALacerda posted:

global_costmap

stonier:
> we probably should not even expose it in dynamic reconfigure for now.

Yeah, hard coding it was one of the options I posed in #187. Are there use cases where you use a VoxelLayer without a StaticLayer though? This problem here is specific to the case where you have a StaticLayer underneath the VoxelLayer (i.e. typical global_costmap configuration).

mikeferguson:
I just noticed this is actually a change in voxel layer, not static layer. In which case, there may actually be a reason to overwrite in some other use cases. This PR will fix it for most users and still allow the other option for those that desire it. LGTM.

mikeferguson:
I'll get a new release out into hydro/indigo as soon as we sort out the test issues of #201

PR #204, opened 2014-06-09
fix hypot issues, add comments to tests from tracking this down

mikeferguson:


efernandez:
:+1: 

I've added a few comments regarding the `hypot` issues, where it could be enough using `int` to store the result of the substraction. However, now I cannot run the tests, so I've simply play a little with a few values running the code below in http://www.compileonline.com/compile_cpp11_online.php.

``` cpp
#include <iostream>
#include <cmath>

using namespace std;

int main()
{
   unsigned int x0 = 20;
   unsigned int x1 = 10;

   unsigned int y0 = 10;
   unsigned int y1 = 20;

   double res_sqrt       = sqrt((x1-x0)*(x1-x0) + (y1-y0)*(y1-y0));
   double res_hypot      = ::hypot(x1-x0, y1-y0);
   double res_hypot_cast = ::hypot(static_cast<int>(x1-x0), static_cast<int>(y1-y0));

   cout << "sqrt       = " << res_sqrt       << endl;
   cout << "hypot      = " << res_hypot      << endl;
   cout << "hypot cast = " << res_hypot_cast << endl;

   return 0;
}
```

mikeferguson:
Both good points, I've fixed up the PR

efernandez:
Good!, but I also want to mention that using `unsigned int` for arithmetic operations isn't safe; they should be used only for bitmasks. Indeed, consider the following example:

``` cpp
#include <iostream>

using namespace std;

int main()
{
    unsigned int i = 30, j =2;
    unsigned int a = j - i;
    int b = a;
    cout << "a = " << a << endl;
    cout << "b = " << b << endl;

   return 0;
}
```

This simulates what the compiler does, i.e. a temporary `unsigned int` is created, which is later interpreted as an `int`.
I'm not proposing any change, I'm just pointing this out, to make it clear that the temporary rvalues of the `unsigned int` expressions are wrong, and in these particular cases, the assignment or cast simply interprets them as `int`, and at the end this behavior seems to be OK for this piece of code.
So, we can leave it like that, but if possible, we could simple use `int` instead of `unsigned int` for the arguments of these methods.

Take the following notes (from one of my colleagues) into consideration:

_Unexpected behavior mix signed with unsigned_

This code produces unexpected behavior All calculations of 2 - 3 or 2 + (-3) produce a big positive number. This is because c converts the negative value to unsigned int. Unfortunately, cppcheck, g++, clang++ nor Intel compiler never provide any warning for lines 10 & 15, even with -Wall or -Wextra. 
Only Intel and Clang complain because we have the comparison between the unsigned expression and 0.
- Only Intel Compiler provides a warning pointless comparison of unsigned integer with zero for lines 8 & 12 (and only when compiling with ccache! why?).
- clang++ provides a "rr.cpp:8:12: warning: comparison of unsigned expression >= 0"

``` cpp
#include <iostream>

int main()
{
unsigned int a = 2;
int sn = -3;
int sp = 3;
if (a + sn >= 0)
  std::cout << "bad addition" << std::endl;
std::cout << (a + sn) << ":" << (sn + a) << std::endl;

if (a - sp > 0)
  std::cout << "bad substraction" << std::endl;

std::cout << (a - sp) << ":" << (-sp + a) << std::endl;
}
```

If you cannot change "a" into an int, the workaround is simply prepending (int) for each usage of
 (int)a.

PR #209, opened 2014-06-11
Remove unused robot_radius parameter

stonier:
@mikeferguson can you confirm this? Looking at the code and experimenting without it, it seems this parameter is unused. Perhaps this was made redundant by the `robot_radius` parameter now in [Costmap2D.cfg](https://github.com/ros-planning/navigation/blob/88e802d89a28d5dbee0ce10ab3c0ab1931830e30/costmap_2d/cfg/Costmap2D.cfg#L20).

For further discussion, see https://github.com/turtlebot/turtlebot_apps/issues/94.


mikeferguson:
Yeah, after looking at the underlying code, it would appear this should not be here.

PR #210, opened 2014-06-12
adds goal to end of path in global_planner, makes move_base's planService search outwards

### LINKED ISSUE #207: global_planner/GlobalPlanner doesn't set the orientation of the last point in the plan
TrajectoryPlannerROS (as it currently is) uses the orientation of the last point in the global plan to turn to once it reaches the goal position.  That works fine with NavFN, which copies the original goal (with orientation) onto the end of the list.  In global_planner's planner_core.cpp, however, pushing that point onto the list is commented out (line 299).  Any particular reason for that?  If you couple global_planner with TrajectoryPlannerROS the robot always turns to a yaw of 0.

KaijenHsiao:
fixes #207

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -404,7 +404,6 @@ namespace move_base {
       ROS_ERROR("move_base cannot make a plan for you because it could not get the start pose of the robot");
       return false;
     }
-    
     geometry_msgs::PoseStamped start;
     //if the user does not specify a start pose, identified by an empty frame id, then use the robot's pose
     if(req.start.header.frame_id == "")
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -394,13 +394,11 @@ namespace move_base {
       ROS_ERROR("move_base must be in an inactive state to make a plan for an external user");
       return false;
     }
-    
     //make sure we have a costmap for our planner
     if(planner_costmap_ros_ == NULL){
       ROS_ERROR("move_base cannot make a plan for you because it doesn't have a costmap");
       return false;
     }
-
     tf::Stamped<tf::Pose> global_pose;
     if(!planner_costmap_ros_->getRobotPose(global_pose)){
       ROS_ERROR("move_base cannot make a plan for you because it could not get the start pose of the robot");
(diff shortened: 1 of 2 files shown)


move_base/src/move_base.cpp
```diff
@@ -414,31 +414,61 @@ namespace move_base {
...
-          else
-            ROS_DEBUG_NAMED("move_base","Failed to find a  plan to point (%.2f, %.2f)", p.pose.position.x, p.pose.position.y);
-        }
-        p.pose.position.x += resolution*3.0;
+    if(!planner_->makePlan(start, req.goal, global_plan) || global_plan.empty()){
+      ROS_DEBUG_NAMED("move_base","Failed to find a plan to exact goal of (%.2f, %.2f), searching for a feasible goal within tolerance", 
+          req.goal.pose.position.x, req.goal.pose.position.y);
+      
+      //update the copy of the costmap the planner uses
+      clearCostmapWindows(2 * clearing_radius_, 2 * clearing_radius_);
```
mikeferguson: Previously, this was run before any planning, should it still be called before the first makePlan? (I initially commented on the commit, not the PR, recreating comment here)

move_base/src/move_base.cpp
```diff
@@ -414,31 +414,61 @@ namespace move_base {
...
+		if(x_offset < 1e-9 && x_mult < -1.0 + 1e-9) continue;
+
+		p.pose.position.y = req.goal.pose.position.y + y_offset * y_mult;
+		p.pose.position.x = req.goal.pose.position.x + x_offset * x_mult;
+
+		if(planner_->makePlan(start, p, global_plan)){
+		  if(!global_plan.empty()){
+
+		    //adding the (unreachable) original goal to the end of the global plan, in case the local planner can get you there
+                    //(the reachable goal should have been added by the planner, assuming your local planner uses it)
```
mikeferguson: "assuming your _local_ planner uses it" ? should that read "global" planner?

move_base/src/move_base.cpp
```diff
@@ -394,7 +394,7 @@ namespace move_base {
       ROS_ERROR("move_base must be in an inactive state to make a plan for an external user");
       return false;
     }
-
+
```
mikeferguson: If there up being changes for issues below, try to also remove these trailing spaces

mikeferguson:
Other than the questions/comments above, LGTM.

move_base/src/move_base.cpp
KaijenHsiao: Good point, I'll shift it.

move_base/src/move_base.cpp
KaijenHsiao: I've changed it to just "the final reachable goal should have been added by the global planner".
(I did mean local--we said we wanted to have the adding of the final goal be in the global planner and not in move_base, for generality.  TrajectoryPlannerROS expects it to be there, but if you write your own global and local planners, you could conceivably not want that.)

KaijenHsiao:
@mikeferguson updated the pull request based on your comments--please merge if it still looks good.

mikeferguson:
@KaijenHsiao looks good - can you squash these commit together as one?

KaijenHsiao:
hate hate git rebase

PR #212, opened 2014-06-12
renames acc_lim_th to acc_lim_theta

efernandez:
`acc_lim_th` param is taken in `trajectory_planner_ros.cpp`, but in `trajectory_planner.cpp` and the dynamic reconfigure `cfg` file uses `acc_lim_theta`, so I think this should be renamed.

### DIFF
--- a/base_local_planner/src/trajectory_planner_ros.cpp
+++ b/base_local_planner/src/trajectory_planner_ros.cpp
@@ -118,7 +118,7 @@ namespace base_local_planner {
       private_nh.param("xy_goal_tolerance", xy_goal_tolerance_, 0.10);
       private_nh.param("acc_lim_x", acc_lim_x_, 2.5);
       private_nh.param("acc_lim_y", acc_lim_y_, 2.5);
-      private_nh.param("acc_lim_th", acc_lim_theta_, 3.2);
+      private_nh.param("acc_lim_theta", acc_lim_theta_, 3.2);
 
       private_nh.param("stop_time_buffer", stop_time_buffer, 0.2);
 


mikeferguson:
Huh, so it is called acc_lim_theta on the wiki, and I just checked each of my config files and they are also acc_lim_theta, I'm guessing nobody has ever been able to properly set the limit?

mikeferguson:
Actually, looks like at least turtlebot has used "acc_lim_th" before -- I just pushed a change (6721c998b3d1a17c224801086cccdaa2cfea0e73) that will warn on using that param, and we will remove it for J-turtle (covered by ticket #213)

efernandez:
@mikeferguson Indeed, there are several parameters which are somehow repeated.
Take for example the trajectory planner. There are 3 important files regarding the parameters (in the `base_local_planner` pkg):
- `src/trajectory_planner_ros.cpp` : Takes params from the param server
- `src/trajectory_planner.cpp` : Doesn't take any param from the param server, so it falls back to the defauls in the dynamic reconfigure cfg defaults
- `cfg/BaseLocalPlanner.cfg` : Defines dynamic reconfigure params; if they aren't in the param server, the defaults are the ones in this cfg file (see this example I've made to check it, for instance: https://github.com/efernandez/ros_test/blob/master/launch/dynamic_reconfigure_test.launch)

Although I haven't verified it thoroughly, I think that now we have to set the following parameters "twice"; I mean, with the name they have in the param server (read by `src/trajectory_planner_ros.cpp`, and passed to the `TrajectoryPlanner` constructor defined in `src/trajectory_planner.cpp`), and also with the name they have in `cfg/BaseLocalPlanner.cfg`, because the first call to the `reconfigureCB` (which is defined in `src/trajectory_planner_ros.cpp`, but at the end calls the `reconfigure` method defined in `src/trajectory_planner.cpp`) ---there's always one initial call--- overwrites them with the defaults in the cfg (unless the same params, with the same name, are in the param server).

Below you have the list of affected parameters (right: params taken by `src/trajectory_planner_ros.cpp`; left: dynamic reconfigure params used by `src/trajectory_planner.cpp`, which the equation used for the constructor call):

``` cpp
acc_lim_theta = acc_lim_th // mentioned in this PR
max_vel_theta = max_rotational_vel
min_vel_theta = -max_rotational_vel
min_in_place_vel_theta = min_in_place_rotational_vel
pdist_scale = path_distance_bias
gdist_scale = goal_distance_bias
simple_attractor = false // hard-coded to false
```

Note that some params are simply passed to the `TrajectoryPlanner`, but others are also used in the ROS wrapper. I also think that it doesn't make sense to take this params with `private_nh.param`, since the dynamic reconfigure server automatically takes them if they are defined in the server, as i'ts shown with the test I've put in my github (link above).

efernandez:
@mikeferguson @DLu Any update on this? Should I send a PR to remove the redudant params, or are they there for some reason?

DLu:
The redundancy is the result of sloppy dynamic_reconfig-ification. I have remove those redundancies where possible in other packages, but apparently can lead to bugs like #187.

PR #214, opened 2014-06-13
add goal to end of path in global_planner, changed move_base's planService to search outwards

### LINKED ISSUE #207: global_planner/GlobalPlanner doesn't set the orientation of the last point in the plan

KaijenHsiao:
fixes #207


mikeferguson:
Thanks!

PR #216, opened 2014-06-17
Remove unnecessary colons

DLu:
Follow up to #201 and #204. Works on my machine at least...

### DIFF
--- a/base_local_planner/src/goal_functions.cpp
+++ b/base_local_planner/src/goal_functions.cpp
@@ -39,7 +39,7 @@
 namespace base_local_planner {
 
   double getGoalPositionDistance(const tf::Stamped<tf::Pose>& global_pose, double goal_x, double goal_y) {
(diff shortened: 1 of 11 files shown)


mikeferguson:
It seems that it would be fine, but I'm not sure why the :: was there in the first place -- @efernandez was there a particular reason to global namespace the hypot calls?

efernandez:
For some reason it didn't work the first time I try with std. I think it was some stupid mistake, because now compiles with std. If it compiles for you as well, go for it.

UPDATE: github's mobile app is quite frustrating... I didn't realized this was already a PR.

efernandez:
:+1: 

FYI I tried with `std::hypot` and that simply fails to compile. With `hypot` is enough (no need for `::hypot`).

mikeferguson:
Ok, that explains how those got there then.

PR #218, opened 2014-06-26
not use dynparam to decrease the speed to move, use service call directly

garaemon:
Currently move_slow_and_clear plugin uses `rosrun dynamic_reconfigure dynparam` to change the speed of base.
However, if filesystem I/O is super heavy, such as on NFS, this `rosrun ...` takes so much time.

This PR uses roscpp service call directly rather than using external process to avoid the problem described above.

### DIFF
--- a/move_slow_and_clear/include/move_slow_and_clear/move_slow_and_clear.h
+++ b/move_slow_and_clear/include/move_slow_and_clear/move_slow_and_clear.h
@@ -41,6 +41,7 @@
 #include <nav_core/recovery_behavior.h>
 #include <costmap_2d/costmap_2d_ros.h>
 #include <boost/thread.hpp>
+#include <dynamic_reconfigure/Reconfigure.h>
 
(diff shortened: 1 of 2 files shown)


garaemon:
updated comment and error message

DLu:
LGTM. How extensively have you tested this?

garaemon:
We are testing this patch on our PR2 during 1 week and no problem is found yet.

We are running move_base on c2. The flesystem of c2 is the same filesystem to c1 via NFS.
Sometimes file accessing on c2 takes a lot of time (about 20 seconds) and it means that `rosrun dynamic_reconfigure dynparam` takes 20 seconds to execute.
After applying this patch, the latency seems better to us.

mikeferguson:
@DLu do you intend to release another set of groovy debs at any point? If so I would suggest doing it soon as the groovy buildfarm will likely be torn down once indigo is released.

PR #219, opened 2014-06-26
not use dynparam to decrease the speed to move, use service call directly

garaemon:
hydro version of https://github.com/ros-planning/navigation/pull/218

### DIFF
--- a/move_slow_and_clear/include/move_slow_and_clear/move_slow_and_clear.h
+++ b/move_slow_and_clear/include/move_slow_and_clear/move_slow_and_clear.h
@@ -41,6 +41,7 @@
 #include <nav_core/recovery_behavior.h>
 #include <costmap_2d/costmap_2d_ros.h>
 #include <boost/thread.hpp>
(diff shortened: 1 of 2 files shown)


efernandez:
Nice!

garaemon:
updated, please check the latest code.

mikeferguson:
This might have broken the build? http://jenkins.ros.org/job/devel-indigo-navigation/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/15/console

Although it appears maybe the farm might be messed up? It can't seem to install libboost-all-dev?

mikeferguson:
Ok, not a broken build for tests (hydro succeeded http://jenkins.ros.org/job/devel-hydro-navigation/27/)

However, I now notice that move_slow_and_clear does not depend on dynamic_reconfigure -- that should probably be added to both the package.xml and CMakeLists else building debians will fail. @DLu, do you want me to do that?

DLu:
Yes, go ahead.

PR #224, opened 2014-07-15
Trajectory marker

efernandez:
This adds the possibility to publish a visualization marker array with the trajectories generated.
They are published if `publish_trajectories` param is set to `True` (available on the dynamic reconfigure `cfg` file). The trajectory markers are separated in different namespaces, depending on the logic that generates them.

This implementation shows only the trajectory endpoint; I think that drawing all the trajectory points is not really useful and shows so many points. The trajectory cost is not drawn because there is no clear upper bound for the conversion from cost to RGB.
I'm open to any advice to improve it.

The PR comes with several commits that removes empty lines, trailing white spaces and unused code or trivial changes.

@RainCT This is inspired in some implementation you did to handle markers and draw the trajectories. I've tried to keep the changes minimal, but maybe I should use your implementation at the end, if you agree.

### DIFF
--- a/base_local_planner/src/trajectory_planner.cpp
+++ b/base_local_planner/src/trajectory_planner.cpp
@@ -42,7 +42,7 @@
 #include <math.h>
 #include <angles/angles.h>
 
-
+#include <tf/transform_datatypes.h>
 
 #include <boost/algorithm/string.hpp>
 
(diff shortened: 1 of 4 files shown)


DLu:
Would it be possible to split this into a functional pull request and a separate white space pull request?

efernandez:
@DLu I've moved the white space and other fixes to  https://github.com/ros-planning/navigation/pull/226

base_local_planner/src/trajectory_planner.cpp
```diff
@@ -56,86 +56,165 @@ using namespace costmap_2d;
...
+      double cost)
+  {
+    visualization_msgs::Marker marker;
+
+    marker.header.frame_id = "odom";
+    marker.header.stamp = ros::Time::now();
+
+    marker.ns = ns;
+    marker.id = id;
+    marker.action = visualization_msgs::Marker::ADD;
```
DLu, line 72: Perhaps explicitly set the marker type to ARROW, or at least put a comment in saying that that's what it defaults to.

DLu:
What does the output of this actually look like?

Given they both share the Trajectory object, I feel like there could be some unification with the DWA trajectory publishing (which isn't the best solution, so could benefit from working with this)

efernandez:
@DLu Now `ARROW` is explicitly set.
Regarding the output, now I'm quite busy, but I'll try to upload some screenshots.
Anyway, I'd like to set the color depending on the cost, but since there's no clear upper bound for that cost I don't really now how to do it.

DLu:
I suppose that's part of the rationale for publishing the trajectories as a point cloud as DWA does, allowing rviz to automatically color them.

mikeferguson:
Hydro is EOL, so we can no longer merge this -- please consider a PR against a newer branch.

PR #225, opened 2014-07-15
Voxel grid model line iterator

efernandez:
This PR makes the `voxel_grid_model` use the `LineIterator` as in the `costmap_model`.
It's just a cosmetic change.

### DIFF
--- a/base_local_planner/src/voxel_grid_model.cpp
+++ b/base_local_planner/src/voxel_grid_model.cpp
@@ -34,6 +34,7 @@
 *
 * Author: Eitan Marder-Eppstein
 *********************************************************************/
+#include <base_local_planner/line_iterator.h>
 #include <base_local_planner/voxel_grid_model.h>
 
 using namespace std;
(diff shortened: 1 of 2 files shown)


DLu:
LGTM. Tested I assume?

efernandez:
@DLu Sorry, but I haven't tested.
I'd like that you tell me of a good example to exercise this code. If you can give me an example config file, I'd try it with the REEM in simulation (http://wiki.ros.org/Robots/REEM); I'd take care of adapting it.
I'm using `costmap` type for several observation sources (in different layers/plugins): laser, sonars, IRs, PCs.
Would it be enough if I change to `voxel` the `map_type`?

base_local_planner/src/voxel_grid_model.cpp
```diff
@@ -91,62 +92,13 @@ namespace base_local_planner {
...
-      den = deltay;
-      num = deltay / 2;
-      numadd = deltax;
-      numpixels = deltay;         // There are more y-values than x-values
-    }
-
-    for (int curpixel = 0; curpixel <= numpixels; curpixel++)
+    for( LineIterator line( x0, y0, x1, y1 ); line.isValid(); line.advance() )
     {
       point_cost = pointCost(x, y); //Score the current point
```
DLu: should be line.getX(), line.getY()

DLu:
Careful visual inspection looks good, but did your version compile?

efernandez:
Sure, it compiles, but I haven't tried it with a configuration that uses `voxel`.
If you want, give me some time and I'll try something in a few weeks.

mikeferguson:
Do any of the unit tests cover this function? If so, that should be a good start to run, if not, maybe it would be a good time to add them....

DLu:
The code does not compile as is. It will work if you change the line I suggested (x and y aren't declared in that scope anymore). None of the unit tests specifically test the cost models, and there is nothing that tests the voxel model.

efernandez:
@DLu I've fixed the compilation problem.
Still I need some time to run/test it.
Writing a test will take more time... is there an example I can start with?

mikeferguson:
Hydro is EOL, so we can no longer merge this -- please consider a PR against a newer branch.

PR #227, opened 2014-07-17
Pose pulled from parameter server when new map received

Spkordell:
Previously, AMCL would load pose information from the parameter server when the node was started. When a map is received, this pose information is used to set the initial pose of the robot. If AMCL received a new map at a later point in time, after the robot's pose has changed, it would reinitialize the particle filter, using the pose information which was read when the node first started, essentially resetting the robot to it's old position every time a new map is received. This fix simply reads in the last pose saved on the parameter server each time a new map is received.


mikeferguson:
LGTM -- @DLu any objections?

DLu:
I haven't worked with AMCL as much, so I don't have much to add.

PR #229, opened 2014-07-22
Add destructors for all of the layers

DLu:
Add destructors for all the layers so that the dynamic parameter clients are destroyed as well.

### DIFF
--- a/costmap_2d/include/costmap_2d/footprint_layer.h
+++ b/costmap_2d/include/costmap_2d/footprint_layer.h
@@ -53,7 +53,7 @@ class FootprintLayer : public Layer
 {
 public:
   virtual void onInitialize();
-  ~FootprintLayer();
+  virtual ~FootprintLayer();
 
   virtual void updateBounds(double robot_x, double robot_y, double robot_yaw, double* min_x, double* min_y, double* max_x,
                              double* max_y);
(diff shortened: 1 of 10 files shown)


mikeferguson:
@DLu is your intent to merge this before or after a new release?

costmap_2d/include/costmap_2d/footprint_layer.h
```diff
@@ -53,6 +53,7 @@ class FootprintLayer : public Layer
 {
 public:
   virtual void onInitialize();
+  ~FootprintLayer();
```
mikeferguson: Should be virtual, no?

DLu: Probably. Should they all be?

DLu:
@mikeferguson I was intending for after the release, since this is mostly a precursor to my next pull request that runs the clear_costmap tests, but we can push it through before if you want.

costmap_2d/include/costmap_2d/footprint_layer.h
DLu: Fixed all.

mikeferguson:
Ok, I'm going to do a release right now -- and then we can merge this later today.

Issue #211, opened 2014-06-12
ClearCostmapRecovery only clears children of ObstacleLayers


severin-lemaignan:
This 'selection' is performed [here](https://github.com/ros-planning/navigation/blob/hydro-devel/clear_costmap_recovery/src/clear_costmap_recovery.cpp#L102). I'm not sure about the reason...

This makes it anyway difficult to develop custom layers that can be cleared as well (like [the sonar layer](https://github.com/DLu/navigation_layers/tree/master/sonar_layer)).

DLu:
Would an array of layers to clear be sufficient?

severin-lemaignan:
@DLu yes, it sounds like a good, flexible idea.

mikeferguson:
Implemented in #230

PR #230, opened 2014-07-23
Clear_costmap_recovery - Different parameter syntax

DLu:
The old behavior was to specify a string parameter, and any obstacle layer with that string within its name would be cleared. 

The new behavior is to specify a list of string names, and any costmap layer with a name that matches one on the list is cleared. 

Fix for #211. Possible fix for #198. 

Includes tests!


clear_costmap_recovery/src/clear_costmap_recovery.cpp
```diff
@@ -59,7 +59,16 @@ void ClearCostmapRecovery::initialize(std::string name, tf::TransformListener* t
...
 
     private_nh.param("reset_distance", reset_distance_, 3.0);
-    private_nh.param("layer_search_string", layer_search_string_, std::string("obstacle"));
+    
+    std::vector<std::string> clearable_layers_default, clearable_layers;
+    clearable_layers_default.push_back( std::string("obstacles") );
+    private_nh.param("layer_names", clearable_layers, clearable_layers_default);
+
+    for(unsigned i=0; i < clearable_layers.size(); i++) {
+        ROS_INFO("Using layer %s", clearable_layers[i].c_str());
```
mikeferguson: Perhaps change this to "Recovery behavior will clear layer %s" or something like that, so users have some idea what "using" means?

clear_costmap_recovery/src/clear_costmap_recovery.cpp
```diff
@@ -78,7 +87,7 @@ void ClearCostmapRecovery::runBehavior(){
     ROS_ERROR("The costmaps passed to the ClearCostmapRecovery object cannot be NULL. Doing nothing.");
     return;
   }
-  ROS_WARN("Clearing costmap to unstuck robot.");
+  ROS_WARN("Clearing costmap outside of radius %f.", reset_distance_);
```
mikeferguson: Might want to be keep the "to unstuck robot" part of the message, since a lot of ros-users questions reference this specific phrase.

costmap_2d/include/costmap_2d/costmap_layer.h
```diff
@@ -46,12 +46,17 @@ namespace costmap_2d
...
+    extra_min_x_( 1e6), extra_min_y_( 1e6),
+    extra_max_x_(-1e6), extra_max_y_(-1e6) {}
+
   bool isDiscretized()
   {
     return true;
   }
 
   virtual void matchSize();
+  void addExtraBounds(double mx0, double my0, double mx1, double my1);
```
mikeferguson, line 69: Can you add a doxygen comment on what exactly this function does?

mikeferguson:
I've added a couple of comments above, other than those, LGTM. I didn't review the tests.

mikeferguson:
Pinging @efernandez as well on this one, since he has been involved in the other discussions.

clear_costmap_recovery/src/clear_costmap_recovery.cpp
DLu: I just looked it up and apparently that's grammatically correct. I assumed it was a typo of unstick.

DLu:
Updated the two debug messages, and added a bunch of comments to costmap_layer.

mikeferguson:
LGTM -- I'm going to wait just a little while before merging as the test for the previous merge hasn't made it through the farm yet (a new "ros_comm" was released to indigo this morning, so the farm is swamped with jobs)

mikeferguson:
Other tests have cleared, this looks good, merging

PR #233, opened 2014-07-31
Added option to fix bug that resets particle weights on motion model updates

sachih:
There's a bug in amcl's motion model--the particle weights are reset in the motion model after the particles are propagated.  This isn't a problem if resampling happens on every step, but if you are using a resample_interval >1, amcl will essentially ignore all but the last observation before the resample.  

Just in case anyone is using resample_interval>1 and has a system working _with_ the bug, we made it an optional param that defaults to the old behavior.  (Feel free to debate which way it should go, or whether it should just be a fix and not a param at all.)

### DIFF
--- a/amcl/cfg/AMCL.cfg
+++ b/amcl/cfg/AMCL.cfg
@@ -55,7 +55,7 @@ odt = gen.enum([gen.const("diff_const", str_t, "diff", "Use diff odom model"),
                 gen.const("omni_corrected_const", str_t, "omni-corrected", "Use corrected omni odom model")],
                "Odom Models")
 gen.add("odom_model_type", str_t, 0, "Which model to use, diff, omni, diff-corrected, or omni-corrected", "diff", edit_method=odt)
-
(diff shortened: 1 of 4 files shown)


mikeferguson:
I'm still looking at this, but I want to add that the default resample_interval (if none is set) is 2. However, every config I can find online is set to 1.

mikeferguson:
@sachih after reviewing this in more detail, and looking at all of the publicly available configurations I can find, I would suggest we not make this a parameter (it adds a significant amount of code for a use case that doesn't appear to be in use anywhere). Could you clean  up this PR to simply fix the bug but not add the parameter? (It appears that the change is only 3 lines then, is that right?)

sachih:
Sure. We added the param since we weren't sure if you wanted to merge it as
a bug fix. Will fix and make a new request.

KaijenHsiao:
new pull request is #243

mikeferguson:
Closed in favor of new request.

PR #234, opened 2014-07-31
Added planner_overclock_ratio param to move_base

KaijenHsiao:
This param times the planner_frequency is the new frequency for the planning loop.  At this frequency, it checks if a new plan is needed based on a new goal or a stuck controller (or if it's time to plan again based on the old planner_frequency).

This eliminates the problem where you ask the robot to go to a new location, or the controller is unable to execute the plan, and instead of replanning, the robot stops and waits for the next planner loop (at a potentially very slow planner_frequency).

(Thanks to Sachi Hemachandra.)

### DIFF
--- a/move_base/cfg/MoveBase.cfg
+++ b/move_base/cfg/MoveBase.cfg
@@ -2,7 +2,7 @@
 
 PACKAGE = 'move_base'
 
-from dynamic_reconfigure.parameter_generator_catkin import ParameterGenerator, str_t, double_t, bool_t
+from dynamic_reconfigure.parameter_generator_catkin import ParameterGenerator, str_t, double_t, bool_t, int_t
 
 gen = ParameterGenerator()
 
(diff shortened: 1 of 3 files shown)


mikeferguson:
@DLu do you have any thoughts on this PR?

@KaijenHsiao is there possibly a better name than "planner_overclock_ratio"?

KaijenHsiao:
I'm sure there is. :)  Do you have any suggestions?  
I thought it should be something like check_replan_needed_planner_frequency_multiplier, but, um, shorter.  But I couldn't think of anything good.

DLu:
So the MoveBase planning logic is a mess, and I'm hoping to finish up some changes and refactoring soon. 

I think the behavior you want can be accomplished by calling planner_cond_.notify_one(), which the code already does here. It's entirely possible that this doesn't happen in practice, but I think that it should happen. Does this reflect what you've seen?

KaijenHsiao:
If I understand correctly, calling planner_cond_.notify_one() only does something useful if your planner_frequency is set to 0.  If you want the planner to run by default at a given slow frequency (such as 0.2 Hz), then calling planner_cond_.notify_one() does not help, since it is still sleeping for 1/planner_frequency seconds before trying to plan again.

KaijenHsiao:
Replaced by #282.

PR #235, opened 2014-08-01
Added visited set to global_planner A* 

sosentos:
This change was made to prevent loops, to improve speed, and for correctness.

(Thanks to Sachi Hemachandra.)

### DIFF
--- a/global_planner/src/grid_path.cpp
+++ b/global_planner/src/grid_path.cpp
@@ -38,38 +38,62 @@
 #include <global_planner/grid_path.h>
 #include <algorithm>
 #include <stdio.h>
+#include <set>
+
 namespace global_planner {
 
 bool GridPath::getPath(float* potential, double start_x, double start_y, double end_x, double end_y, std::vector<std::pair<float, float> >& path) {
     std::pair<float, float> current;
+    std::set<int> path_set; //this is used to ensure we don't get stuck in a loop (as it seems possible to do) 
     current.first = end_x;
     current.second = end_y;
 
     int start_index = getIndex(start_x, start_y);
-
+    path_set.insert(getIndex(end_x, end_y));
     path.push_back(current);
     while (getIndex(current.first, current.second) != start_index) {
         float min_val = 1e10;
         int min_x = 0, min_y = 0;
+        int min_ind = -1; 
+        //check the neighbors to find the node with the lowest cost 
         for (int xd = -1; xd <= 1; xd++) {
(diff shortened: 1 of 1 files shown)


mikeferguson:
@DLu - please review.

DLu:
I apologize for the delay in reviewing this. 

Can you give a set of circumstances where you get the looping behavior?

global_planner/src/grid_path.cpp
```diff
@@ -38,38 +38,62 @@
...
             for (int yd = -1; yd <= 1; yd++) {
                 if (xd == 0 && yd == 0)
                     continue;
+	
                 int x = current.first + xd, y = current.second + yd;
+
+                //make sure we don't connect because of wrap around 
+                if(x < 0 || x > xs_ -1 || y > ys_ -1 || y < 0){
+                  continue; 
+                }
```
DLu, line 69: This is not needed because the outlineMap function is called. https://github.com/ros-planning/navigation/blob/hydro-devel/global_planner/src/planner_core.cpp#L55

sachih, line 69: I have seen this behavior when using the grid path extractor with the A\* expansion. (A number of times it didn't return a path but got stuck in a loop eating up all the memory on the machine).

DLu:
Do you have a replicable example?

KaijenHsiao:
I suspect #248 is a fix to the issue behind this pull request as well.  (The visited set was put in just to prevent infinite loops.)

KaijenHsiao:
I will add, though, that your A\* implementation is, um, highly unorthodox.  And should probably have some comments explaining, for instance, why 1) it has no visited set / check for cheaper paths in the expansion (which I can see is not needed in the specific case of 4-connected grids with constant step cost, but that is not said anywhere), or 2) that it's actually doing an 8-connected search for a path through a gradient generated using 4-connected A\* expansion (and thus does not actually generate an optimal 8-connected path).

DLu:
Fair point. One of  the motivations for the refactoring of navfn was to create an A\* implementation, but since the gradient path is much poorer on A\* graphs, I haven't focused on it much. Definitely welcome pull requests, even if just for documentation.

KaijenHsiao:
I added a note to the wiki instead. :)

DLu:
@sosentos @KaijenHsiao are you okay with closing this?

KaijenHsiao:
@DLu Yes, please do.

PR #236, opened 2014-08-03
Allow map-switching while moving.

zdydek:
This change creates memory objects for a new map once it is received and swaps the new objects for the old.
The new particle filter is initialized with pose mean and covariance from the old one.
If the map's metadata is properly populated, this allows for on-the-fly map switches (even while the robot is moving).
Rotation is not supported at this time.


mikeferguson:
No longer mergeable, especially since hydro is EOL -- if still relevant, please open a new PR against indigo or higher

PR #239, opened 2014-08-06
The TF from IMU to base frame was inverted

ghost:
The IMU measurements are in the IMU frame of reference, and we would like to get them in the base footprint frame of reference. This change makes it so the transform applied to the measurements is from IMU Frame to base footprint frame of reference, rather than the opposite way.

This issue was observed while working with a Clearpath Husky with a UM6 IMU which is mounted rotated. We had been using a wrong imu_link -> base_link tf to compensate for this problem until identifying it. Presumably this effects any robot which has a non-flat mounted IMU.

### DIFF
--- a/robot_pose_ekf/src/odom_estimation_node.cpp
+++ b/robot_pose_ekf/src/odom_estimation_node.cpp
@@ -244,7 +244,7 @@ namespace estimation
       return;
     }
     StampedTransform base_imu_offset;
-    robot_state_.lookupTransform(base_footprint_frame_, imu->header.frame_id, imu_stamp_, base_imu_offset);
+    robot_state_.lookupTransform(imu->header.frame_id, base_footprint_frame_, imu_stamp_, base_imu_offset);
     imu_meas_ = imu_meas_ * base_imu_offset;
 
     imu_time_  = Time::now();


mikeferguson:
This does not seem right, the first parameter to lookup transform should be the target_frame, which should be (and currently is) the base_footprint.

ghost:
I see what you mean. Poking around different packages that I happen to be familiar with (gmapping, hector_slam, the rviz imu plugin ...) they all seem to use this function call the way that I had been, with the source frame first. Maybe this is an ambiguity in the documentation, and source_frame really means something different from what you and I are understanding here. I'm not sure...

mikeferguson:
@akalmbach can you point me at the code you are looking at in gmapping/hector_slam -- I find no use of lookupTransform in gmapping, and all of the uses in hector are destination then source (https://github.com/tu-darmstadt-ros-pkg/hector_slam/search?q=lookuptransform&ref=cmdform).

TF is pretty specific on this:  "The direction of the transform returned will be from the target_frame to the source_frame. Which if applied to data, will transform data in the source_frame into the target_frame". (source: http://wiki.ros.org/tf/Overview/Using%20Published%20Transforms#lookupTransform)

mikeferguson:
Closing this, please reopen with comment if you find documentation to show that this is actually a correct fix.

PR #241, opened 2014-08-27
Fixed GridPath::getPath

mistoll:
The algorithm must not walk upwards, thus the minimum value must be saved between loops.

(The old method breaks, when start position equals end position)

### DIFF
--- a/global_planner/src/grid_path.cpp
+++ b/global_planner/src/grid_path.cpp
@@ -48,8 +48,8 @@ bool GridPath::getPath(float* potential, double start_x, double start_y, double 
     int start_index = getIndex(start_x, start_y);
 
     path.push_back(current);
+    float min_val = 1e10;
     while (getIndex(current.first, current.second) != start_index) {
-        float min_val = 1e10;
         int min_x = 0, min_y = 0;
         for (int xd = -1; xd <= 1; xd++) {
             for (int yd = -1; yd <= 1; yd++) {


mikeferguson:
heads up to @DLu

DLu:
I'm not sure I follow the motivation of the proposed changes. 

1) I fail to see how the algorithm will behave any differently if the start position and the end position are the same, with the addition of the PR. 

2) It is true that the minimum value should continually decrease as the path planning goes on, but I don't think it is necessary to enforce that.

mistoll:
Sorry, for the late reply, I'm currently lacking internet.

I found a  constellation with use_grid_path: true, use_dijkstra: false and start approx end where the loop won't exit The result is a bad alloc exception, when the vector get's too large. That's bad so you need to change something. Unfortunately I'm not familar with the move_base setup. Maybe there are some more parameter which influence this behavior.

The image shows the potential when this happens. Here start is cell (390, 75) and end is (391,73). The path toggles around (390,74) and (389,75) infinitely.
potential
So there might be no minimum at the start position. Thus this is a problem of the potential, not traceback. But it would be nice not to get a bad alloc, when this happens.

To answer your questions
1)
Any of the adjacent cells have a value less than 1e10. Thus the original algorithm always finds a (min_x, min_y) and the condition (min_x == 0 && min_y == 0) will never be hit.
There is at least one case where this results into a bad alloc exception.
With the change, the condition will be hit eventually.
This might happen, when start == end, but also when the potential does have local minimas. In the other cases the algorithm should behave identically.

2)
This alorithm can not walk upwards. If there's a local minimum it would just walk back and forth until the memory is full. Of course it's up to your, how to handle that case. In terms of computing power the fix is really cheap.

DLu:
I think #248 fixes the problem with the potential you found, @mistoll. Please let me know if that branch works for you. 

Also, testing this use case made me shut down my computer twice, so I hope you're happy! :-) That's clearly suboptimal behavior.

mistoll:
I'm glad you found the aktual bug behind this behavior. #248 woks for me.
Those infinite loops are anoying ;-)

It's not important, but IMO the loop will never hit the condition

>  if (min_x == 0 && min_y == 0)
>    return false;
> As you added a loop couter, that might not be worth the effort looking into it.

DLu:
If there's a valid path, yes, it will never hit that condition. I'm not too worried about it though. 

Closing this in favor of #248.

PR #243, opened 2014-08-29
bug fix to remove particle weights being reset when motion model is updated

KaijenHsiao:
same as pull request #233 without the param


mikeferguson:
Thanks!

Issue #242, opened 2014-08-27
Proposal of fixing missing Findgtest.cmake warning in clear_costmap_recovery and costmap_2d


arne48:
For solving this warning I would propose:
To change the "find_package(gtest)" instruction in the CMakeLists.txt of  
clear_costmap_recovery in line 43 
and costmap_2d in line 143 
to  "find_package(GTest)". 
This will lead to bigger problems when testing is intensified in the future.
So hopefully this little typo can be fixed quite soon.

DLu:
Thanks for the suggested fix. It would be good to fix this. However, on my machine, changing it to `find_package(GTest REQUIRED)` results in the following. 

```
-- costmap_2d: 1 messages, 0 services
CMake Error at /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:91 (MESSAGE):
  Could NOT find GTest (missing: GTEST_LIBRARY GTEST_MAIN_LIBRARY)
Call Stack (most recent call first):
  /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:252 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-2.8/Modules/FindGTest.cmake:150 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  navigation/costmap_2d/CMakeLists.txt:143 (find_package)


-- Configuring incomplete, errors occurred!
make: *** [cmake_check_build_system] Error 1
Invoking "make cmake_check_build_system" failed
```

(Changing it without the REQUIRED results in different errors.)

I am not a GTest/CMake expert. Do you know what the problem might be?

(my repo: https://github.com/DLu/navigation/commit/feae132229a1ef6674f95474f49d75f670d2cf93)

arne48:
To me this looks like the GTest library isn't installed correctly on your system. 
While preparing GTest I hat the problem that on Ubuntu there is no precompiled library available.
So at first I installed the "libgtest-dev" package from the repository  and then did this to build and put the library in place:
cd /usr/src/gtest
sudo cmake .
sudo make
sudo cp libg\* /usr/lib/
Maybe you can try this out and report if this solved your problem. I was trying some other things on the tests and they are running through now.

DLu:
I figured it was something like that. That seemed to solve the problem, although I can't imagine why there's no precompiled library.

mikeferguson:
[...]

PR #244, opened 2014-08-29
Fix Gtest

DLu:
Fix for #242

### DIFF
--- a/clear_costmap_recovery/CMakeLists.txt
+++ b/clear_costmap_recovery/CMakeLists.txt
@@ -40,7 +40,7 @@ target_link_libraries(clear_costmap_recovery layers ${catkin_LIBRARIES})
 if(CATKIN_ENABLE_TESTING)
   # Find package test dependencies
   find_package(rostest REQUIRED)
-  find_package(gtest)
+  find_package(GTest REQUIRED)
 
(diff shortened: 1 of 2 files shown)


mikeferguson:
@DLu this doesn't seem to work on my system:

```
CMake Error at /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:108 (message):
  Could NOT find GTest (missing: GTEST_LIBRARY GTEST_MAIN_LIBRARY)
Call Stack (most recent call first):
  /usr/share/cmake-2.8/Modules/FindPackageHandleStandardArgs.cmake:315 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-2.8/Modules/FindGTest.cmake:150 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  navigation/costmap_2d/CMakeLists.txt:143 (find_package)

-- Configuring incomplete, errors occurred!
```

Are you installing some other package to get gtest (if so, we would need to update the package.xml)?

DLu:
Did you try the fix mentioned in #242? Perhaps that's not the right way, but that's how I got it to work?

mikeferguson:
This won't work on the farm, see #252

PR #245, opened 2014-08-29
Hydro devel amcl small fix

sachih:
Updated the cfg text to reflect the new observation model


mikeferguson:
Thanks! Have you added this also to the wiki (http://wiki.ros.org/amcl)?

PR #248, opened 2014-09-03
Global planner fix

DLu:
Fix for the bug reported in #241. 

The actual problem was the adjustment I make to the end of the found path (expander.clearEndpoint) which overwrote previous potentials calculated by the algorithm. When the end of the path is sufficiently close to the start (as @mistoll reported) , this behavior would overwrite the starting minimum, causing the algorithm to loop forever. This PR fixes the overwriting problem, and in my experiments fixes the infinite looping. 

Just to be on the safe side, I also added a loop counter to GridPath to avoid the infinite loop in the chance that it occurs in some other case.


mistoll:
Great work, the patch works for me.

PR #250, opened 2014-09-15
costmap_2d: export library layers

kmhallen:
Many layer plugins also instantiate a FootprintLayer. Layers created outside the costmap_2d package were not able to link against liblayers. The outside layers would still compile, but loading an outside layer before a layer in costmap_2d caused a plugin load error.


mikeferguson:
LGTM

Issue #158, opened 2013-12-12
robot_pose_ekf param output_frame_ is set as topic-name


tobiasneumann:
The param output_frame_ is used as the toppic name:
odom_estimation_node.cpp line 95

and seems to have no effect on the frame_id of the Pose-msg

mikeferguson:
#168 fixes the tf frame aspects -- topic name remains set to output_frame_

PR #168, opened 2014-02-11
Fix for issue #167: hardcoded tf frames

muhrix:
This is my suggested fix for issue #167. At the moment, robot_pose_ekf makes use of hardcoded string literals "odom_combined" and "base_footprint" as names for output and base frames. Since support for tf_prefix was dropped, having multiple robots with separate tf trees means having unique frame names (e.g., `robot1_base_footprint`).

The changes are backwards compatible with tf_prefix and hence it would not break any launch file. I have made some tests myself, but I hope someone else can assert these proposed changes.

It is rather straightforward: I added 2 string variables (output_frame_ and base_footprint_frame_) to OdomEstimation class to represent output and base frames, along with 2 mutator methods to set their values after fetching them from the parameter server. Then, the hardcoded references (string literals) to the output ("odom_combined") and base ("base_footprint") frames were replaced with output_frame_ and base_footprint_frame_, respectively.

With these changes it is possible to set the output and base frames using parameters "output_frame" and "base_footprint_frame".

I modified the example launch file to illustrate how to set the base frame as well as the output frame.

PR #251, opened 2014-09-18
Hardcode EKF topic to ~/odom_combined

jspricke:


mikeferguson:
This has the potential to break some configs -- however it does seem like the right fix, and has come up several times before. We fixed pieces of the tf vs. topic name issue in https://github.com/ros-planning/navigation/pull/168.

This would fix https://github.com/ros-planning/navigation/issues/158

I'm in favor of merging, any objections from other users/maintainers following this?

mintar:
LGTM

PR #252, opened 2014-09-21
catkin provides gtest, no need for find_package

mikeferguson:
Proper fix for #242 (see comments from dirk), replaces #244


mikeferguson:
@DLu @arne48 please review that this meets your needs.

DLu:
If it works on the build farm, LGTM.

arne48:
This is a much easier option so solve that issue, as the one I proposed.
Reduce the amount of not needed invocations is always the preferable option. 
So this is a good fix.

PR #254, opened 2014-09-24
base_local_planner: adds waitForTransform

### LINKED ISSUE #196: waitForTransform missing in base_local_planner/goal_functions
It appears there is a waitForTransform call missing before [transforming the plan here in L110 of goal_functions.cpp](https://github.com/ros-planning/navigation/blob/groovy/base_local_planner/src/goal_functions.cpp#L110). This pops up every once in a while on ROS Answers and users have traced the extrapolation error back to the previously mentioned transfom call. See examples: [1](http://answers.ros.org/question/63360/extrapolation-error-using-hector_mapping-move_base/), [2](http://answers.ros.org/question/106996/hydro-move_basetf2-extrapolation-error-into-the-future-errror/), [3](http://answers.ros.org/question/63360/extrapolation-error-using-hector_mapping-move_base/), [4](http://answers.ros.org/question/40541/extrapolation-error-of-local-cost-map-in-navigation/), [5](http://answers.ros.org/question/164911/move_base-and-extrapolation-errors-into-the-future/).

bit-pirate:
The first waitForTransform (Line 107 - 112) resolves #196. I also added the second one to avoid similiar issues in other situations.

An alternative to using `waitForTransform` could be requesting the latest available transform (`ros::Time(0)`). If applicable, I generally prefer this alternative, but since I'm not too familiar with the various parts of the navi stack, I decided to stick as close to the current code and logic as possible.

### DIFF
--- a/base_local_planner/src/goal_functions.cpp
+++ b/base_local_planner/src/goal_functions.cpp
@@ -104,9 +104,12 @@ namespace base_local_planner {
 
       // get plan_to_global_transform from plan frame to global_frame
       tf::StampedTransform plan_to_global_transform;
-      tf.lookupTransform(global_frame, ros::Time(), 
-          plan_pose.header.frame_id, plan_pose.header.stamp, 
-          plan_pose.header.frame_id, plan_to_global_transform);
+      tf.waitForTransform(global_frame, ros::Time::now(),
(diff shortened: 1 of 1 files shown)


mikeferguson:
LGTM, #196 already has a ton of up-votes.

PR #258, opened 2014-10-02
costmap_2d: tf_prefix support in obstacle_layer

mani-monaj:
`costmap_2d` supports `tf_prefix` for `global_frame` and `robot_base_frame` [(1)](https://github.com/ros-planning/navigation/blob/e9c3ac822ac4ddb9af2605493fd7c37044cdf2d8/costmap_2d/src/costmap_2d_ros.cpp#L81), however the`obstacle_layer` does not. This PR adds `tf_prefix` support to `obstacle_layer`.


mikeferguson:
LGTM

PR #260, opened 2014-10-08
add global planner run depend in meta package.

jihoonl:


mikeferguson:
I'm just going to point out that by adding this to the metapackage, global_planner will now show up in the list of navigation packages in the header on the wiki. @DLu -- do you consider the global_planner to be stable enough for that?

DLu:
Yes.

PR #263, opened 2014-10-22
prevent div by zero in raytraceLine

artivis:
While mapping a polygon from world to map with costmap_2d woldToMap it might happen that two consecutive points end-up with the same coordinates, transforming a segment into a single point.
Then calling polygonOutlineCells calls raytraceLine which raises a signal Floating point exception -> floating point divide by zero.


mikeferguson:
It seems there is a type -- "double scale" became "doublecale"

artivis:
Oups my bad. Sorry for that.
It's fixed.

mikeferguson:
No problem -- the divzero is a good catch.

PR #265, opened 2014-10-29
Add DepthImage observation source

efernandez:
This allows to generate the clearing `xyz` points using the `PinholeCameraModel` of the depth images, for those pixels/points where there's no "echo"/"return" (invalid pixels).

See https://github.com/ros-perception/image_pipeline/pull/104

This PR also needs https://github.com/ros-perception/image_pipeline/pull/108 to be merged before, since it tries to avoid copying code from `depth_image_proc`. This other PR simply exports some headers and extends the `convert` function to allow for setting `range_max` for invalid pixels.

### DIFF
--- a/costmap_2d/plugins/obstacle_layer.cpp
+++ b/costmap_2d/plugins/obstacle_layer.cpp
@@ -1,6 +1,14 @@
 #include<costmap_2d/obstacle_layer.h>
 #include<costmap_2d/costmap_math.h>
 
+#include <sensor_msgs/image_encodings.h>
+#include <image_transport/camera_common.h>
+#include <pcl_conversions/pcl_conversions.h>
+
+#include <pcl/point_types.h>
+
+#include <limits>
+
 #include <pluginlib/class_list_macros.h>
 PLUGINLIB_EXPORT_CLASS(costmap_2d::ObstacleLayer, costmap_2d::Layer)
 
@@ -10,6 +18,78 @@ using costmap_2d::FREE_SPACE;
 
 using costmap_2d::ObservationBuffer;
 using costmap_2d::Observation;
+
+// Encapsulate differences between processing float and uint16_t depths
+template<typename T> struct DepthTraits {};
+
+template<>
+struct DepthTraits<uint16_t>
+{
+  static inline bool valid(uint16_t depth) { return depth != 0; }
+  static inline float toMeters(uint16_t depth) { return depth * 0.001f; } // originally mm
+  static inline uint16_t fromMeters(float depth) { return (depth * 1000.0f) + 0.5f; }
+  static inline void initializeBuffer(std::vector<uint8_t>& buffer) {} // Do nothing - already zero-filled
+};
+
+template<>
+struct DepthTraits<float>
+{
+  static inline bool valid(float depth) { return std::isfinite(depth); }
+  static inline float toMeters(float depth) { return depth; }
+  static inline float fromMeters(float depth) { return depth; }
+
+  static inline void initializeBuffer(std::vector<uint8_t>& buffer)
+  {
+    float* start = reinterpret_cast<float*>(&buffer[0]);
+    float* end = reinterpret_cast<float*>(&buffer[0] + buffer.size());
+    std::fill(start, end, std::numeric_limits<float>::quiet_NaN());
+  }
+};
+
+template<typename T>
+void convert(const sensor_msgs::ImageConstPtr& depth_msg, pcl::PointCloud<pcl::PointXYZ>::Ptr& cloud_msg, const image_geometry::PinholeCameraModel& model, double range_max = 0.0)
+{
+  // Use correct principal point from calibration
+  float center_x = model.cx();
+  float center_y = model.cy();
+
+  // Combine unit conversion (if necessary) with scaling by focal length for computing (X,Y)
+  double unit_scaling = DepthTraits<T>::toMeters( T(1) );
+  float constant_x = unit_scaling / model.fx();
+  float constant_y = unit_scaling / model.fy();
+  float bad_point = std::numeric_limits<float>::quiet_NaN();
+
+  pcl::PointCloud<pcl::PointXYZ>::iterator pt_iter = cloud_msg->begin();
+  const T* depth_row = reinterpret_cast<const T*>(&depth_msg->data[0]);
+  int row_step = depth_msg->step / sizeof(T);
+  for (int v = 0; v < (int)cloud_msg->height; ++v, depth_row += row_step)
+  {
+    for (int u = 0; u < (int)cloud_msg->width; ++u)
+    {
+      pcl::PointXYZ& pt = *pt_iter++;
+      T depth = depth_row[u];
+
+      // Missing points denoted by NaNs
+      if (!DepthTraits<T>::valid(depth))
+      {
+        if (range_max != 0.0)
+        {
+          depth = DepthTraits<T>::fromMeters(range_max);
+        }
+        else
+        {
+          pt.x = pt.y = pt.z = bad_point;
+          continue;
+        }
+      }
+
+      // Fill in XYZ
+      pt.x = (u - center_x) * depth * constant_x;
+      pt.y = (v - center_y) * depth * constant_y;
+      pt.z = DepthTraits<T>::toMeters(depth);
+    }
+  }
+}
 
 namespace costmap_2d
 {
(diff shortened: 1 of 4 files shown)


efernandez:
This is an example configuration extracted from the `global_costmap.yaml` that I'm using (only the part related with the `DepthImage`:

```
global_costmap:

  plugins:
    - name: obstacle_rgbd_layer
      type: 'costmap_2d::ObstacleLayer'

  obstacle_rgbd_layer:
    enabled: true
    observation_sources: depth_image
    combination_method: 0

    depth_image:
      sensor_frame: base_rgbd_camera_optical_frame
      data_type: DepthImage
      topic: base_rgbd_camera/depth/image_raw
      expected_update_rate: 1.0
      observation_persistence: 0.0
      marking: true
      clearing: true
      raytrace_range: 4.0
      obstacle_range: 3.8
      min_obstacle_height: 0.1
      max_obstacle_height: 1.8
```

mikeferguson:
https://github.com/ros-perception/image_pipeline/pull/108 needs to also be ported to indigo-devel, and image_pipeline really needs to be re-released into both hydro and indigo before we merge this. Ideally we would also put a minimum version on image_pipeline -- however, since we are using the same branch for hydro+indigo right now, there isn't a good way to do that.

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -61,27 +70,14 @@ void ObstacleLayer::onInitialize()
     source_node.param("inf_is_valid", inf_is_valid, false);
     source_node.param("clearing", clearing, false);
     source_node.param("marking", marking, true);
+    source_node.param("marking", marking, true);
```
DLu: Duplicated line.

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -61,27 +70,14 @@ void ObstacleLayer::onInitialize()
...
-    }
-
-    std::string raytrace_range_param_name, obstacle_range_param_name;
-
-    //get the obstacle range for the sensor
-    double obstacle_range = 2.5;
-    if (source_node.searchParam("obstacle_range", obstacle_range_param_name))
-    {
-      source_node.getParam(obstacle_range_param_name, obstacle_range);
-    }
```
DLu, line 87: Is there a reason you changed this functionality? I'm not sure anyone is using the obstacle_range parameter in a namespace other than the specific costmap namespace, but I don't see why you'd lump the change in with this PR.

efernandez, line 87: You're right, I shouldn't have add this change here.
I think it's strange that other params (also in the specific costmap namespace) are retrieved only with a simple `source_node.param`, isn't it?. I think these two other params can be retrieved the same way.

I'm going to remove the commit that did this... but if you want I can send it on a separate PR.

costmap_2d/plugins/obstacle_layer.cpp
efernandez: Solved and removed from this PR... now it's in https://github.com/ros-planning/navigation/pull/268

efernandez:
@DLu Note that last week I updated the PR following your comments. Let me know if there's something else needed in order to merge this.

efernandez:
I've just updated the PR to check that the `PinholeCameraModel` is initialized before using it.
I've also uploaded a video showing the `costmap_2d` update with this `DepthImage` layer for a robot using a simulated RGBD sensor.

I've uploaded a video showing how the costmap gets updated: http://youtu.be/HpwLxY4QN2E
The video ends showing that clearing also works for far away readings (by moving the wall in the simulation).

I'm using this modified version of the rgbd sensor (openni) gazebo plugin in order to clear obstacles when they are far away from the sensor max range:
https://github.com/pal-robotics-forks/gazebo_ros_pkgs/pull/2

@DLu Note that I've done this because of this issue:
http://answers.ros.org/question/191684/voxelgridplugin-clearing-voxels

Note that the openni driver for RGBD cameras puts a bad point (`NaN`), so the `DepthImage` observation source also manage it in the same way for far away readings.

mikeferguson:
@efernandez Is this going to compile against image_pipeline in Indigo? (it looks like not, since it has PCL stuff in it).

Also, note that github says this can't be merged automatically.

paulbovbel:
@efernandez, is there a reason you wouldn't use depthimage_to_laserscan for this?

efernandez:
@mikeferguson This PR is for `hydro-devel`; for `indigo` I can create another PR if you need it. Is there an easy way to know why this can't be automatically merged (the PR is against `hydro-devel`)?

@paulbovbel AFAIK, with [depthimage_to_laserscan](http://wiki.ros.org/depthimage_to_laserscan) I can only have a slice at a given distance. One option could be to have a single `LaserScan` with the minimum range for each `yaw` angle, as a projection onto the ground. Does `depthimage_to_laserscan` supports this?

efernandez:
@mikeferguson Solved, now it can be automatically merged; I've just rebased the new `hydro-devel` commits on the PR branch.

efernandez:
@paulbovbel OK, I see that with `scan_height` I can actually have a the same effect, right?

paulbovbel:
Yes it does. This actually optimizes the processing downstream as well, since there's only one observation in the buffer created per yaw angle, rather than multiple with a full depthimage->pointcloud conversion.

We could keep the functionality encapsulated within the depthimage_to_laserscan node, or maybe even reuse the artifacts from the DepthImageToLaserScanROS library, rather than recreating the functionality and maintaining it in two places.

efernandez:
@paulbovbel That's true. The only difference is indeed that the `DepthImage` `data_type` of this PR performs a depthimage->pointcloud conversion. With this in an `ObstacleLayer`, I agree that it's actually better to use the `depthimage_to_laserscan` in the first place.

However, apart from this, I still see one case not covered by the costmap_2d, which might be probably addressed in a different way: if the point cloud is filtered or change in some way, the `range_max` (`+Inf`) values must be added to allow for clearing the obstacles. At this point is where with only the point cloud in `costmap_2d` it's not possible to replace `+Inf` (or `NaN` as they're now) values with `range_max`, because we need the camera model (from the `camera_info`).
I think the best option here is to add that `range_max` in the node touching the point cloud (which probably means subscribing to the depth image and the camera info) and output the point cloud with the `range_max`, so we allow for clearing.
Originally, I also proposed a changed in `depth_image_proc` to accommodate an option for this, but that will get propagated to all the nodes subscribed to it.

paulbovbel:
This is currently done in point cloud to laserscan, although the camera info (min max angle) is entered in manually. You're right, it would be worthwhile to provide an option to pull this data from camera_info as well.

It would be very useful to somehow indicate how much negative space is captured in a point cloud, but this seems to be a limitation of the data structure. Not all point clouds necessarily have an associated depthimage and/or camera_info.

efernandez:
@paulbovbel It seems that both `depthimage_to_laserscan` and `pointcloud_to_laserscan` are good enough if at the end we use a 2D costmap (`ObstacleLayer`).

The problem when the sensor is seeing the ground (and hence it must be filtered) can be solved using `pointcloud_to_laserscan`, instead of `depthimage_to_laserscan`, which doesn't support it. :)
I wonder whether it's possible or not to filter the ground (points under a given height) in `pointcloud_to_laserscan`...?

paulbovbel:
@efernandez Please check the wiki (http://wiki.ros.org/pointcloud_to_laserscan), the min_height and target_frame parameters are there for this reason. You can accomplish the same with depthimage_to_laserscan as long as your sensor is parallel with the floor.

efernandez:
@paulbovbel Thanks!
Anyway, even when the sensor is parallel to the floor, the `scan_height` cannot be set to the number of rows on the images, as it can be seeing the ground. In that case I can either look for the appropriate number, or use `pointcloud_to_laserscan`, as you're suggesting.

@DLu With the options commented by @paulbovbel , I think the `DepthImage` added to `costmap_2d` is not really needed.

DLu:
@efernandez @paulbovbel 
If using a VoxelMap however, depthimage_to_laserscan would not be sufficient, correct?

paulbovbel:
@dlu no, I guess that would be a separate pull if necessary. Would you see a benefit from adding a depthimage source to voxelmap? Or would that be left to depth_image_proc for the pointcloud conversion?

DLu:
@paulbovbel If I'm understanding correctly, the tricky part is how to handle the edge cases of Inf and NaN in the DepthImage when putting them into a PointCloud. It seems like that might be useful based on the discussion here, although I'm not enough of an expert on DepthImages to really say.

efernandez:
@DLu @paulbovbel AFAIK the alternative is `pointcloud_to_laserscan` using the `LaserScan` observation source of `costmap_2d`. With `depthimage_to_laserscan` (with the current implementation) you cannot include the whole depth image (all the rows) without including the ground (basically because there's no transformation into the `base_link` or similar). If I have time during this week, I'll upload a pair of videos showing that.

Regarding `VoxelLayer`, the input would usually be a point cloud, so for a correct clearing you need the camera info in order to reproject the depth image into the max range at the appropriate ray (if the point cloud was generated from a depth image, of course).

vrabaud:
FYI, https://github.com/ros-perception/image_pipeline/pull/109 got merged

mikeferguson:
We will need to fork for indigo before we can merge any version of depth image (given the different APIs)

efernandez:
I ended up following more or less @paulbovbel advices. The pointcloud_to_laserscan is actually a nice approach to simply use LaserScan from costmap_2d.
In fact, in my particular case the camera wasn't parallel to the ground, so depthimage_to_laserscan wasn't an option as well, but it could be used in that case.

DLu:
Ok, how should we proceed from here?
A) Close this PR. 
B) Close this and open an issue for future integration
C) Keep working to reseat this.

mikeferguson:
At this point, I would suggest we close it -- if we want to add this feature, we should add to Jade.

PR #266, opened 2014-10-29
Call deleteMaps before initMaps

efernandez:
It seems it was a memory leak, because calling `initMaps` allocates new memory for `costmap_` without de-allocating the previous one!

### DIFF
--- a/costmap_2d/src/costmap_2d.cpp
+++ b/costmap_2d/src/costmap_2d.cpp
@@ -64,6 +64,7 @@ void Costmap2D::deleteMaps()
 void Costmap2D::initMaps(unsigned int size_x, unsigned int size_y)
 {
   boost::unique_lock < boost::shared_mutex > lock(*access_);
+  delete[] costmap_;
   costmap_ = new unsigned char[size_x * size_y];
 }
 


efernandez:
@DLu I haven't tested it, but I think there was a memory leak there.

mikeferguson:
Not keen on this form of the change -- APIs should be easy to use correctly and hard to use incorrectly,  what we should do is update initMaps() so that it automatically prevents a memory leak, by adding the following before allocating new memory:

```
 if (costmap_)
    delete[] costmap_;
```

efernandez:
@mikeferguson Done.

mikeferguson:
Ok, maybe I wasn't clear there -- I was suggesting we should put that code inside the initMaps function -- anytime you are reallocating the pointer, you should make sure to deallocate it first.

mikeferguson:
So that if someone forgets to call deleteMaps, they still don't leak memory...

efernandez:
Sure!

BTW, I think it's enough to put `delete[] costmap_;` without checking with the `if (costmap_)`, as in `deleteMaps()`.

PR #268, opened 2014-10-31
Simplify get params

efernandez:
For the `obstacle_range` and `raytrace_range` params, which could be read as the other params in the costmap namespace (e.g. `marking`, `clearing`, ...).

### DIFF
--- a/costmap_2d/plugins/obstacle_layer.cpp
+++ b/costmap_2d/plugins/obstacle_layer.cpp
@@ -47,7 +47,7 @@ void ObstacleLayer::onInitialize()
     ros::NodeHandle source_node(nh, source);
 
     //get the parameters for the specific topic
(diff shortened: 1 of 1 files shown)


mikeferguson:
I think this will break navigation on PR2 -- See https://github.com/PR2/pr2_navigation/blob/hydro-devel/pr2_navigation_config/move_base/costmap_common_params.yaml

Basically, they define obstacle _range and raytrace_range at the costmap level, and then search up to it, rather than defining those for each sensor.

mikeferguson:
Turtlebot_navigation does the same: https://github.com/turtlebot/turtlebot_apps/blob/hydro/turtlebot_navigation/param/costmap_common_params.yaml

DLu:
-1. Is there a reason to do this other than simplifying the code?

efernandez:
@DLu No reason... closing the issue.

PR #269, opened 2014-10-31
Global planner heuristic

DLu:
As pointed out by @KaijenHsiao and mateus03 (from answers.ros.org), global planner plans in Euclidean space, but used a Manhattan Distance heuristic, when using the A\* option. This pull request fixes that and changes one edge case so that it always finds a path when one is available.

### DIFF
--- a/global_planner/src/astar.cpp
+++ b/global_planner/src/astar.cpp
@@ -84,7 +84,7 @@ void AStarExpansion::add(unsigned char* costs, float* potential, float prev_pote
 
     potential[next_i] = p_calc_->calculatePotential(potential, costs[next_i] + neutral_cost_, next_i, prev_potential);
     int x = next_i % nx_, y = next_i / nx_;
-    float distance = abs(end_x - x) + abs(end_y - y);
+    float distance = hypot(end_x - x, end_y - y);
 
     queue_.push_back(Index(next_i, potential[next_i] + distance * neutral_cost_));
     std::push_heap(queue_.begin(), queue_.end(), greater1());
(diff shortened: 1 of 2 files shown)


KaijenHsiao:
Er.  Given that your potential expansion is only done over a 4-connected grid, it seems to me that Manhattan Distance is the more-correct heuristic to use (assuming you want to leave it with 4-connected potentials).

If you want it to always find a path (in the 8-connected sense) when one is available, you'd also need to implement the rest of the pieces of A\* that aren't needed for the 4-connected, fixed-step-cost case: change calculatePotentials in astar.cpp to add potentials to diagonals (right now it only adds potentials to up-down-left-right), add a visited nodes list to stop visiting nodes that have been fully expanded, and add to the node expansion checking of whether the path costs to a given neighbor node have decreased (rather than just skipping over nodes that already contain potentials).

mikeferguson:
Is the consensus that this should not be merged? If so @DLu please close.

KaijenHsiao:
Hold on, just because it doesn't make it into proper 8-connected A\* doesn't
mean it isn't worth merging.  Upon further reflection, Euclidean distance
is probably more useful, and certainly fixing of edge cases is useful!
I'll try it out this week.

KaijenHsiao:
OK, well, I tried it out and it doesn't seem to hurt anything.  

Although upon further further reflection, it is not clear to me that Euclidean distance is actually more useful in this weird situation of doing the A\* search with 4-connected potentials and then searching back through for 8-connected paths (since the actual path costs are still Manhattan distances, which means that Manhattan distance is a more accurate heuristic than Euclidean distance).  But at worst it might make it expand slightly more than otherwise, and it's at least a more intuitive heuristic anyway.  I didn't see any obvious change in behavior when just running random goals, but I also haven't stared carefully at specific examples or what gets expanded in the search.

Anyway, the edge-case-fixing seems useful.  I say, may as well merge.

PR #273, opened 2014-11-09
Style fixes, remove undefined function, future proof use of std::

mikeferguson:


mikeferguson:
@DLu please take a quick look at these and merge -- not really anything of substance here, I did add a note about your IROS paper, since thats where I actually figured out what the heck updatebounds was doing.

DLu:
LGTM. If it compiles, merge it.

mikeferguson:
It does compile, and pass tests.

PR #274, opened 2014-11-09
Fix voxel grid update when robot base moves

mikeferguson:
While the costmap is correct, the underlying voxel grid ends up being incorrect when the robot moves as the underlying voxel_grid is not cleared out prior to copying the data back into the new location.

I went back through the history, and found https://github.com/ros-planning/navigation/commit/b6d26e8e14051b1073619d2f1ef1b6b66025053a, which shows that resetMaps used to also clear the voxel_grid_ prior to plugin re-write.

I do need to test this fully to make sure that it doesn't cause issues (note the deactivate/activate wrapped around reset() function right above it), but wanted to open a PR so I could put in this extra information and links.

### DIFF
--- a/costmap_2d/include/costmap_2d/voxel_layer.h
+++ b/costmap_2d/include/costmap_2d/voxel_layer.h
@@ -86,6 +86,8 @@ public:
 protected:
   virtual void setupDynamicReconfigure(ros::NodeHandle& nh);
 
+  virtual void resetMaps();
+
 private:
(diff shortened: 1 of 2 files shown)


mikeferguson:
Tested quite a bit today, doesn't appear to cause any issues, and definitely makes the voxel grid visualization correct again (also presumably fixes problems with clearing things that never really existed).

DLu:
+1

Issue #131, opened 2013-09-25
ABORTED status "sticks" to subsequent goal


pirobot:
I posted this on [answers.ros.org](http://answers.ros.org/question/83116/actionlib-aborted-status-sticks-between-goals/) but I figured it belongs here.

I am using the move_base action server and the SimpleActionClient (Python) to send a series of navigation goals to my robot. At the moment I am doing this in simulation, not on a real robot. My setup is Ubuntu 12.04 and ROS Groovy Debian install.

The problem I am having is that when a navigation goal is aborted because the goal cannot be reached, the next goal location in the list is also aborted even though it is perfectly reachable. It's as if the first ABORTED status "sticks" long enough to get attached to the next goal. Is there something I have to do after an aborted goal to clear the status? I've already tried cancel_goal() and cancel_all_goals() after a goal is aborted but it does not change the behavior.

pirobot:
The script at the [following link](https://github.com/pirobot/rbx1/blob/master/rbx1_nav/nodes/nav_abort_test.py) reliably demonstrates the problem if one of the waypoints is located inside an obstacle that exists on a loaded map.  For example, if the third waypoint in the script is inside an obstacle, then move_base aborts that goal but then also aborts on the fourth waypoint even though the fourth waypoint is reachable as can be proved by removing the third waypoint from the list and running the script again.

mikeferguson:
One suggestion -- rostopic echo the /result topic and see if it's actually a bug in the navigation stack, or actionlib client.

pirobot:
Here is the output from rostopic echo /move_base/result when running the test script.  Note that both the third and fourth goals are aborted:

```
$ rostopic echo /move_base/result
header: 
  seq: 0
  stamp: 
    secs: 1380555457
    nsecs: 545758313
  frame_id: ''
status: 
  goal_id: 
    stamp: 
      secs: 1380555449
      nsecs: 211750984
    id: /nav_test_8450_1380555448686-1-1380555449.212
  status: 3
  text: Goal reached.
result: 

---
header: 
  seq: 1
  stamp: 
    secs: 1380555464
    nsecs: 550348662
  frame_id: ''
status: 
  goal_id: 
    stamp: 
      secs: 1380555458
      nsecs: 549788951
    id: /nav_test_8450_1380555448686-2-1380555458.550
  status: 3
  text: Goal reached.
result: 

---
header: 
  seq: 2
  stamp: 
    secs: 1380555470
    nsecs: 905009851
[...]

PR #280, opened 2014-11-21
adding lock to planner makePlan fail case

phil0stine:
A planner mutex is created in the [planner thread](https://github.com/phil0stine/navigation/blob/35eba4f3d02efd6a5c14c2ac4244bdb6dd3b9027/move_base/src/move_base.cpp#L598-619), then locked and unlocked.

`makePlan` is called _after_ the mutex is unlocked, and in the meantime, it is possible the controller has executed its recovery behaviors, set `runPlanner_` to [false](https://github.com/phil0stine/navigation/blob/35eba4f3d02efd6a5c14c2ac4244bdb6dd3b9027/move_base/src/move_base.cpp#L987-989), and called [resetState](https://github.com/phil0stine/navigation/blob/35eba4f3d02efd6a5c14c2ac4244bdb6dd3b9027/move_base/src/move_base.cpp#L1145-1149), which sets `state_` to PLANNING from CLEARING.

Now, when `makePlan` returns, if `gotPlan` is false, it will fall into the second [condition](https://github.com/phil0stine/navigation/blob/35eba4f3d02efd6a5c14c2ac4244bdb6dd3b9027/move_base/src/move_base.cpp#L647), since `state_` has just been set to PLANNING.

Then, it will set `state_` to CLEARING, which will be the current state the next time a goal is sent to move_base. Even if the new goal is valid, a recovery behavior will be performed.

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -649,12 +649,14 @@ namespace move_base {
         ros::Time attempt_end = last_valid_plan_ + ros::Duration(planner_patience_);
 
         //check if we've tried to make a plan for over our time limit
-        if(ros::Time::now() > attempt_end){
+	lock.lock();
+        if(ros::Time::now() > attempt_end && runPlanner_){
           //we'll move into our obstacle clearing mode
           state_ = CLEARING;
           publishZeroVelocity();
           recovery_trigger_ = PLANNING_R;
         }
+	lock.unlock();
       }
 
       if(!p_freq_change_ && planner_frequency_ > 0)


move_base/src/move_base.cpp
```diff
@@ -649,12 +649,14 @@ namespace move_base {
         ros::Time attempt_end = last_valid_plan_ + ros::Duration(planner_patience_);
 
         //check if we've tried to make a plan for over our time limit
-        if(ros::Time::now() > attempt_end){
+	lock.lock();
```
mikeferguson, line 652: please fix the indentation on this line

mikeferguson:
Thanks for the detailed PR. I took a quick look at this, but I will take a closer look over the weekend (the lock/unlock logic in here is quite involved)

phil0stine:
I will submit a new PR, and yes the logic is involved. I had not see [#131](https://github.com/ros-planning/navigation/issues/131) until you mentioned it, but it describes the exact behavior that I was seeing on the robot. This patch resolved the issue.

My only concern was whether _not_ calling `publishZeroVelocity` in the block would have unintended consequences, but it looks like all cases where `runPlanner_` is set to false would either already have zero velocity (eg, goal is reached), or explicitly calls `publishZero`.

PR #281, opened 2014-11-22
adding lock to planner makePlan fail case

phil0stine:
Same issue described (here)[https://github.com/ros-planning/navigation/pull/280]. As I say in the previous thread, I believe this is a fix for #131.

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -649,12 +649,14 @@ namespace move_base {
         ros::Time attempt_end = last_valid_plan_ + ros::Duration(planner_patience_);
 
         //check if we've tried to make a plan for over our time limit
-        if(ros::Time::now() > attempt_end){
+	 lock.lock();
+        if(ros::Time::now() > attempt_end && runPlanner_){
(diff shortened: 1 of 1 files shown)


mikeferguson:
This is clearly proper, as every other access to runPlanner_ is under lock. However, I'm not entirely sure your description in #280 matches what is actually happening and thus there may still be an issue. In executeCycle, when we run out recovery behaviors, runPlanner_ is set to false long before resetState() is called (unless we have some crazy optimizations going on in the compiler).

phil0stine:
I do think what I described is what is actually happening, though I don't recall noticing the issue until we ran a slower global planner (sbpl).

I agree that in executeCycle, runPlanner_ is correctly set to false, and then resetState is called, re-setting state_ to PLANNING (which would be the state whenever a new goal is received).

My contention is that since the planner thread gives up the lock before calling makePlan, if the above sequence of events occurs after the plan thread gives up the lock but before makePlan returns, then the planner thread would incorrectly set state_ to CLEARING for the next received goal.

PR #282, opened 2014-11-23
use timer rather than rate for immediate wakeup

mikeferguson:
This is an alternative to #234.

I don't usually use a planner_frequency > 0, so I haven't been able to test this, but in theory it allows you to _immediately_ wake the planner when a new plan comes in, without adding additional parameters or even waiting for any amount of time.

@KaijenHsiao would you have time to possibly test this?

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -594,25 +594,28 @@ namespace move_base {
     return global_pose_msg;
   }
 
+  void MoveBase::wakePlanner(const ros::TimerEvent& event)
+  {
+    // we have slept long enough for rate
+    planner_cond_.notify_one();
+  }
+
   void MoveBase::planThread(){
     ROS_DEBUG_NAMED("move_base_plan_thread","Starting planner thread...");
     ros::NodeHandle n;
-    ros::Rate r(planner_frequency_);
+    ros::Timer timer;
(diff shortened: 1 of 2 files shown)


KaijenHsiao:
I'll hopefully be able to try it out next week.

move_base/src/move_base.cpp
```diff
@@ -658,11 +661,17 @@ namespace move_base {
...
 
-      if(!p_freq_change_ && planner_frequency_ > 0)
-        r.sleep();
-
       //take the mutex for the next iteration
       lock.lock();
+
+      //setup sleep interface if needed
+      if(planner_frequency_ > 0){
+        ros::Duration sleep_time = ros::Time::now() - (start_time + ros::Duration(1.0/planner_frequency_));
```
KaijenHsiao: sleep_time is negative--the line should read 
ros::Duration sleep_time = (start_time + ros::Duration(1.0/planner_frequency_)) - ros::Time::now();

KaijenHsiao:
I noted the one change needed to make this work in the line comments.  
Otherwise, it works great!  Thanks for the improvement.  I'll close my pull request when you've made the fix and merged.

mikeferguson:
Updated to include fix.

Issue #213, opened 2014-06-12
Updates for J-Turtle


mikeferguson:
J-turtle is more than a year away, so we are likely to forget about things. This ticket will track the things that should be done for J-turtle:
- Remove param and warning for acc_lim_th (https://github.com/ros-planning/navigation/commit/6721c998b3d1a17c224801086cccdaa2cfea0e73)
- Fix name of image_loader library (https://github.com/ros-planning/navigation/issues/208)
- Remove PCL (https://github.com/ros-planning/navigation/issues/264)

DLu:
Should we start an actual branch to start accumulating these?

mikeferguson:
We will, I figure around March or April. If we create a branch now, we'll have a lot of work to do keeping it in sync with bugfixes still going into indigo right now.

mikeferguson:
As it stands, the only real improvement that has noticeable effects on folks is the base_global_planner change. We also don't have any work done towards the PCL stuff.

DLu:
I was also thinking about some additional "breaking" changes, namely:
- Getting rid of the "footprint_layer" (which is really part of the obstacle layer)
- Changing DWA to be plugin-based.

mikeferguson:
@DLu those both sound good. I think the best approach initially (as it sounds like those both entail some development) is to have feature branch for each of these new developments, which we can then merge in. If we have several good reasons to start the jade-devel branch earlier, we certainly can.

mikeferguson:
Also, if you are going to modify DWA to be plugin-based, we should make sure to have a pretty significant code review of that. I can help out, and I imagine the Yujin guys would be interested in helping as well.

mikeferguson:
Remove PCL has it's own ticket, I've handled the other items listed above in the new jade-devel branch

PR #283, opened 2014-11-23
adding makePlan function with returned cost to base_global_planner

phil0stine:
Request an additional function to base global planner interface that allows the user to call makePlan and get the cost associated with a successful plan. An example use case is one where the user has a handful of potential goals and wishes to choose between them. 

Though the costs themselves may not have intuitive value, they can be compared against one another when choosing between different goals.

NavfnROS has some similar functions, but these are unique to that particular class and are therefore not extensible when testing other planners using pluginlib. Whether or not this is the adopted interface, I think there should be some way to query global planner costs for given goals.

### DIFF
--- a/nav_core/include/nav_core/base_global_planner.h
+++ b/nav_core/include/nav_core/base_global_planner.h
@@ -58,6 +58,22 @@ namespace nav_core {
           const geometry_msgs::PoseStamped& goal, std::vector<geometry_msgs::PoseStamped>& plan) = 0;
 
       /**
+       * @brief Given a goal pose in the world, compute a plan
+       * @param start The start pose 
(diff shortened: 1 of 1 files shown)


mikeferguson:
So, while we don't do the greatest job of preserving ABI -- this is a case where we are changing the ABI of a class used as a plugin interface. As the navigation stack is > 1.0, this is a generally bad idea.

We can certainly target to include this change in ROS Jade, but I don't think we should merge this with hydro/indigo as they are intended to be stable. Please see #213 for a list of things we are targeting to ROS Jade.

mikeferguson:
I have cherry-picked this into jade-devel

phil0stine:
Awesome thanks!

PR #285, opened 2014-11-27
Remove cell_data.h

mikeferguson:
@DLu -- it appears this class and all support code is found in inflation_layer.h and that no file anywhere in the navigation stack #includes this file.


DLu:
+1

Issue #267, opened 2014-10-30
Voxel layer not cleared near borders


AlexReimann:
There seems to be a problem with the clearing algorithm of the voxel layer in case of a tilted sensor.

Setup:
Settings was set so that  the camera is tilted by 45° pointing to the floor.
The transformations did not change (robot was not moved).
I moved my hand into the area of the sensor and then removed it again.
My hand was marked in the voxel layer but some points were not removed after taking the hand out.

![not clearing](https://cloud.githubusercontent.com/assets/7090417/4856031/7723801a-60b1-11e4-96c7-a39bdd311d4d.png)
side view

![not clearing2](https://cloud.githubusercontent.com/assets/7090417/4855947/98d74c7a-60af-11e4-956f-66b102114414.png)
top view

![not clearing3](https://cloud.githubusercontent.com/assets/7090417/4855949/9de758a4-60af-11e4-8571-ebc16917de0d.png)
front view

Red boxes are the voxel layer visualized with a size of 0.05 m, which is the same as the voxel layer is set to (z_resolution is 0.05 and map resolution also 0.05 m). The greenish points are the clearing_endpoints.

So bascially it is possible that some voxels can be marked at a position but not be cleared by the ray tracing algorithm even though the the ray should go through them. This only seems to be happen near the border.

View from the sensor frame. Comparison with and without the voxels, which shows that there are marking enpoints behind the boxs / voxels:
![clearing endpoints](https://cloud.githubusercontent.com/assets/7090417/4856042/a6acdd2c-60b1-11e4-8efb-b79f919dd63b.png)

Costmap configuration: http://pastebin.com/RGqHUKxd

Is it just like this because of the inaccuracy of the Bresenham algorithm?

DLu:
Hi Alex. 
I'm trying to figure out exactly what I'm looking at. Does this [this diagram](https://docs.google.com/drawings/d/1E3QqgEAzZDbzY1OksuDYZ0oER9zSmaYbuHMYecB4uLA/edit?usp=sharing) look like what you have from the side view?

AlexReimann:
Updated comment with side top, front view and changed the text to make it a bit more clear how the setup is (also changed the setup a bit to something more easy to understand).

AlexReimann:
I first thought the following two a bit odd lines would be the culprit, but doesn't help with the problem:

https://github.com/ros-planning/navigation/blob/hydro-devel/costmap_2d/plugins/voxel_layer.cpp#L261
[...]

PR #293, opened 2014-12-17
Accurate Bresenham raytracing implementation and clearing for corner cases

AlexReimann:
Changes / fixes clearing with Bresenham raytracing using a more accurate version
Adds parameters for accuracy settings
Adds parameter for including corner cases into clearing

See also https://github.com/ros-planning/navigation/issues/267

Implementation based on: http://graphics.idav.ucdavis.edu/education/GraphicsNotes/CAGDNotes/Bresenhams-Algorithm.pdf

### DIFF
--- a/costmap_2d/plugins/voxel_layer.cpp
+++ b/costmap_2d/plugins/voxel_layer.cpp
@@ -62,6 +62,12 @@ void VoxelLayer::onInitialize()
     voxel_pub_ = private_nh.advertise < costmap_2d::VoxelGrid > ("voxel_grid", 1);
 
   clearing_endpoints_pub_ = private_nh.advertise<sensor_msgs::PointCloud>( "clearing_endpoints", 1 );
+
+  private_nh.param("clear_corner_cases", clear_corner_cases_, false);
+
+  int accuracy_multiplier_bits = 10;
+  private_nh.param("accuracy_multiplier_bits", accuracy_multiplier_bits, 10);
+  voxel_grid_.setAccuracyMultiplierBits(accuracy_multiplier_bits);
 }
 
 void VoxelLayer::setupDynamicReconfigure(ros::NodeHandle& nh)
@@ -88,6 +94,8 @@ void VoxelLayer::reconfigureCB(costmap_2d::VoxelPluginConfig &config, uint32_t l
   unknown_threshold_ = config.unknown_threshold + (VOXEL_BITS - size_z_);
   mark_threshold_ = config.mark_threshold;
   combination_method_ = config.combination_method;
+  clear_corner_cases_ = config.clear_corner_cases;
+  voxel_grid_.setAccuracyMultiplierBits(config.accuracy_multiplier_bits);
   matchSize();
 }
 
@@ -165,12 +173,7 @@ void VoxelLayer::updateBounds(double robot_x, double robot_y, double robot_yaw, 
 
       //now we need to compute the map coordinates for the observation
       unsigned int mx, my, mz;
-      if (cloud.points[i].z < origin_z_)
-      {
-        if (!worldToMap3D(cloud.points[i].x, cloud.points[i].y, origin_z_, mx, my, mz))
-          continue;
-      }
-      else if (!worldToMap3D(cloud.points[i].x, cloud.points[i].y, cloud.points[i].z, mx, my, mz))
+      if (!worldToMap3D(cloud.points[i].x, cloud.points[i].y, cloud.points[i].z, mx, my, mz))
       {
         continue;
       }
@@ -353,10 +356,9 @@ void VoxelLayer::raytraceFreespace(const Observation& clearing_observation, doub
     {
       unsigned int cell_raytrace_range = cellDistance(clearing_observation.raytrace_range_);
 
-      //voxel_grid_.markVoxelLine(sensor_x, sensor_y, sensor_z, point_x, point_y, point_z);
       voxel_grid_.clearVoxelLineInMap(sensor_x, sensor_y, sensor_z, point_x, point_y, point_z, costmap_,
                                       unknown_threshold_, mark_threshold_, FREE_SPACE, NO_INFORMATION,
-                                      cell_raytrace_range);
+                                      cell_raytrace_range, clear_corner_cases_);
 
       updateRaytraceBounds(ox, oy, wpx, wpy, clearing_observation.raytrace_range_, min_x, min_y, max_x, max_y);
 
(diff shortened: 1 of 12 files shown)


costmap_2d/plugins/voxel_layer.cpp
```diff
@@ -62,6 +62,14 @@ void VoxelLayer::onInitialize()
     voxel_pub_ = private_nh.advertise < costmap_2d::VoxelGrid > ("voxel_grid", 1);
 
   clearing_endpoints_pub_ = private_nh.advertise<sensor_msgs::PointCloud>( "clearing_endpoints", 1 );
+
+  combination_method_ = 1;
```
mikeferguson: should this actually be here?

voxel_grid/include/voxel_grid/voxel_grid.h
```diff
@@ -86,6 +86,27 @@ class VoxelGrid
   void reset();
   uint32_t* getData() { return data_; }
 
+  /**
+   * @brief Sets the multiplier which determines accuracy
+   * @param Number of bits which should be used for accuracy
+   *
+   * Number of bits should be less than:
+   * (bits representing the max of an integer) minus
+   * (bits needed to represent the * maximum raytrace distance (= maximum sensor range) in cells)
```
mikeferguson: Should that star (*) be in the middle of the line (before maximum raytrace)?

voxel_grid/include/voxel_grid/voxel_grid.h
```diff
@@ -226,15 +248,20 @@ class VoxelGrid
   template <class ActionType>
   inline void raytraceLine(
     ActionType at, double x0, double y0, double z0,
-    double x1, double y1, double z1, unsigned int max_length = UINT_MAX)
+    double x1, double y1, double z1, unsigned int max_length = UINT_MAX, bool include_corner_cases = false)
   {
-    int dx = int(x1) - int(x0);
-    int dy = int(y1) - int(y0);
-    int dz = int(z1) - int(z0);
+    double dx = x1 - x0;
```
mikeferguson, line 259: do you have any numbers on the performance penalty of switching all this from int to double?

voxel_grid/include/voxel_grid/voxel_grid.h
```diff
@@ -226,15 +248,20 @@ class VoxelGrid
...
+    double abs_dx = fabs(accuracy_multiplier_ * dx);
+    double abs_dy = fabs(accuracy_multiplier_ * dy);
+    double abs_dz = fabs(accuracy_multiplier_ * dz);
 
-    unsigned int abs_dx = abs(dx);
-    unsigned int abs_dy = abs(dy);
-    unsigned int abs_dz = abs(dz);
+    if(abs_dx > INT_MAX || abs_dy > INT_MAX || abs_dz > INT_MAX)
+    {
+      ROS_ERROR("Voxel grid: Accuracy multiplier is set too high. Possible integer overflow while clearing.");
```
mikeferguson: this should probably be a ROS_ERROR_THROTTLE, else too high of a setting is likely to tank your computer with messages.

voxel_grid/include/voxel_grid/voxel_grid.h
AlexReimann, line 259: Nope, but it still uses integers in the actual ray tracing which is the most work.

voxel_grid/include/voxel_grid/voxel_grid.h
```diff
@@ -243,38 +270,72 @@ class VoxelGrid
...
+    double scale = std::min(1.0, max_length / dist);
 
-    //is x dominant
     if (abs_dx >= max(abs_dy, abs_dz))
     {
-      int error_y = abs_dx / 2;
-      int error_z = abs_dx / 2;
+      int error_y = (int)(abs_dy * x_lost_rounding - abs_dx * y_lost_rounding);
+      int error_z = (int)(abs_dz * x_lost_rounding - abs_dx * z_lost_rounding);
+      int number_of_cells = abs(int(x0) - int(x1));
```
mikeferguson: this is maybe a little nitpicky -- but this number is not actually number_of_cells yet (until multiplied by scale). I would suggest fixing these lines (3 of them, plus 6 calls to bresenham functions) to do the multiplication by scale here (and also make them unsigned int to begin with). Long term, this should avoid people potentially changing one line below and not the others.

mikeferguson:
This looks pretty good -- I would suggest adding a link to that PDF somewhere in the code for any future developer/maintainer.

This is a pretty big change to a core piece of the navigation stack, so we will leave this PR open at least a week so that others have time to test it, and comment if they have any issues.

voxel_grid/include/voxel_grid/voxel_grid.h
AlexReimann: It is used with and without multiplying by scale, but after looking at it again it doesn't really make sense:
It is later used with min() and scale has to be between 0 and 1. So I'll just multiply it always with scale.
(The whole thing should be already done way before here imho, but maybe in a later patch)

AlexReimann:
Known bug: Clearing corner cases can lead to underflow / overflow of the clearing which then leads to actually marking
So on hold till I find a fix which does not take up too much computation

AlexReimann:
The issue is the accuracy and it only happens when the end point is directly on the z- axis border of voxel grid cells. Also occurence depends on the accuracy.
One fix would be to check if the end point is on the border and add / subtract a little offset, might be computational "heavy".

The other one would be to remove corner clearing for the last cleared cell. Basically it does this right now: https://docs.google.com/drawings/d/1YeeKCFF7LxVRLOkIBawkwEzePvb2JeTrjfjLt8b18Bk/edit?usp=sharing

So if we skip the clearing of additional corner cells after clearing the last cell it fixes it. But might under clear? I will do some more testing on that again.. if it works we might want to go with that.
Opinions on that?

Third option would be to check for underflow /overflow and don't clear in this case. But this is even more computational heavy with the current implementation than the first proposed fix.

I have another pull request incoming after this which changes the way updating is handled. It makes the whole thing around 30% faster (with a kinect) and also allows to visualize which cells are cleared (which is really nice for debugging). It would also allow an underflow detection without slowing it down much (maybe 1% ?).
Non-optimized version, which is actually slower but has the visualisation, is here https://github.com/AlexReimann/navigation/commit/e20243d643d0d497352a176797efb1b1945f3096

AlexReimann:
I introduced a cached updating scheme of the costmap which fetches overflows / underflows in case of corner case clearing.
It also makes the code faster than the original implementation in the case that number of voxels << number of cleared voxels (e. g. 3d sensor where most voxels get cleared a few hundred times depending on the resolution).
When using corner case clearing it now runs at around the same speed as the original implementation or faster (depends on the length of rays).
Also adds visualization of the cleared voxels.
This needs some serious testing.

AlexReimann:
Should I re-open the pull request in indigo-devel?

mikeferguson:
Yes please.

AlexReimann:
Reopend on indigo-devel: #316

PR #295, opened 2014-12-19
Break infinite loop when tolerance 0 is used

v4hn:
With an increment of 0, there wouldn't be an increment...

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -426,7 +426,7 @@ namespace move_base {
       bool found_legal = false;
       float resolution = planner_costmap_ros_->getCostmap()->getResolution();
       float search_increment = resolution*3.0;
-      if(req.tolerance < search_increment) search_increment = req.tolerance;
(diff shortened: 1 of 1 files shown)


mikeferguson:
Tests failing after merge -- not sure why exactly, but the greenroom loop test is timing out: http://jenkins.ros.org/job/devel-indigo-navigation/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/61/ all other tests have runtimes nearly the same as before.

v4hn:
Heya, thanks for the quick handling of the request.

I just had a look at the test and it does not have _anything_ to do with move_base
as far as I can see. I'm therefore pretty confident that this is independent of
my bugfix.

mikeferguson:
Yeah, it seemed pretty suspect as being not related to the PR. I've kicked the farm to rerun tests to see if it was actually a farm failure (because the test completes on my local desktop)

mikeferguson:
Yup, back to stable -- must have been an issue with the farm

PR #296, opened 2014-12-23
Add ARCHIVE_DESTINATION for static builds

garyservin:
This is required for building static libraries, e.g. for Android


mikeferguson:
Is there a reason this isn't applied to all of the packages? (for instance, no change to move_base, but there is a library in there?)

garyservin:
Sorry, forgot to add that one. Added.

PR #297, opened 2014-12-23
Mark move_base headers for installation

garyservin:
This is needed to use move_base on Android either with rosjava using JNI or as a native activity.

Do you think an `if(ANDROID)` is necessary around this part?


move_base/CMakeLists.txt
```diff
@@ -70,3 +70,10 @@ install(
     LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
     RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
+
+## Mark cpp header files for installation
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.h"
+  PATTERN ".svn" EXCLUDE
```
mikeferguson: since we use git, not svn, this line should not be needed.

mikeferguson:
Can you squash these two commits into one (cleaner history)

garyservin:
Done! thanks for the help!

mikeferguson:
Thanks for the contributions -- I'll merge this as soon as the currently running devel job completes.

PR #298, opened 2015-01-04
Switch to tf2

vrabaud:
Please do not merge yet as:
- I am not done with robot_pose_ekf yet
- I am discussing with @tfoote to see if some code I have put in there could not be merged in tf2
  But you can still give feedback :)


mikeferguson:
We should also probably target this at jade-devel (which I'll create shortly)

vrabaud:
well, I believe some headers are not meant to be public (e.g the ones in costmap_2d/include that match the plugins in costmap_2d/plugins/). Maybe a cleanup on that side would help know whether we break the API or not.

mikeferguson:
This could potentially go into indigo -- but i really don't want to put big changes into hydro at this point with only a few months before the buildfarm gets torn down. It takes a really long time to find runtime issues that creep up in navigation, and I really don't want to leave the last release with newly introduced bugs.

vrabaud:
The code ot updated and now compiles, it needs that patch though: https://github.com/ros/geometry_experimental/pull/76

mikeferguson:
@vrabaud -- Thanks for this -- I'll try to give it some runtime testing over the next week or two (since it is a big change)

vrabaud:
thx, I just wanted to get some feedback. I'll make another update soonish then (I need to get the tests to compile too).

vrabaud:
sorry to be slow on that one: I actually spent the day fixing a deadlock in tf2 that a navigation test revealed ....

mikeferguson:
@vrabaud, no problem -- I haven't had time to test it either. Also, just FYI, I'm planning to fork for indigo-devel and jade-revel tomorrow (we've been meaning to do it for a while). Once I do, we should target this to jade-devel (and then decide if we are going to back port to indigo as well).

vrabaud:
please let me know once you have jade-devel and I'll make my PR to it. I did not see any noticeable performance improvement or degradation with tf2 so we can probably keep that PR for Jade.

All tests are fine now btw so once my PRs make it to tf2, we are good to go.

PR #301, opened 2015-01-22
Swap map service

liz-murphy:
Adds a service which a client can use to send a new map (nav_msgs/OccupancyGrid) and new initial pose (geometry_msgs/Pose) for amcl to load and localize against.

### DIFF
--- a/amcl/package.xml
+++ b/amcl/package.xml
@@ -30,14 +30,12 @@
     <build_depend>std_srvs</build_depend>
     <build_depend>tf</build_depend>
     <build_depend>nav_msgs</build_depend>
-    <build_depend>geometry_msgs</build_depend>
     <build_depend>message_generation</build_depend>
 
     <run_depend>roscpp</run_depend>
     <run_depend>dynamic_reconfigure</run_depend>
     <run_depend>tf</run_depend>
     <run_depend>nav_msgs</run_depend>
-    <run_depend>run_msgs</run_depend>
     <run_depend>message_runtime</run_depend>
 
     <test_depend>map_server</test_depend>
--- a/amcl/package.xml
+++ b/amcl/package.xml
@@ -29,10 +29,16 @@
     <build_depend>rostest</build_depend>
     <build_depend>std_srvs</build_depend>
     <build_depend>tf</build_depend>
+    <build_depend>nav_msgs</build_depend>
+    <build_depend>geometry_msgs</build_depend>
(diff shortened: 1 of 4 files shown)


amcl/package.xml
```diff
@@ -29,10 +29,16 @@
...
     <build_depend>tf</build_depend>
+    <build_depend>nav_msgs</build_depend>
+    <build_depend>geometry_msgs</build_depend>
+    <build_depend>message_generation</build_depend>
 
     <run_depend>roscpp</run_depend>
     <run_depend>dynamic_reconfigure</run_depend>
     <run_depend>tf</run_depend>
+    <run_depend>nav_msgs</run_depend>
+    <run_depend>run_msgs</run_depend>
```
mikeferguson: I presume this is supposed to be geometry_msgs

amcl/srv/SwapMap.srv
```diff
@@ -0,0 +1,5 @@
+ nav_msgs/OccupancyGrid map_msg
```
mikeferguson: I suggest this be called "map" rather than "map_msg" -- we already know it is a message.

mikeferguson:
A couple of general comments:
- I would suggest not putting a new service message definition here in AMCL -- I actually think that should be upstreamed into nav_msgs as SetMap (with the same definition you have right now) as there is already a GetMap service (http://docs.ros.org/api/nav_msgs/html/srv/GetMap.html).
- If we decide to keep the message here, I see no dependency on the generated messages in your CMake -- this means it may or may not build properly on the farm.

liz-murphy:
Fixed issues and created a corresponding PR for the service message in nav_msgs: https://github.com/ros/common_msgs/pull/51

mikeferguson:
This is pending upstream merge of common_msgs change -- and will also need to be retargeted to the proper branch (let's wait and see if we can get this into indigo-devel -- I'm all for it on nav side, we just have to see about the message backport).

stwirth:
This has to be adapted to use a PoseWithCovarianceStamped.

stwirth:
I hacked on it, PR following now.

stwirth:
Superseded by #321.

mikeferguson:
It looks like the messages never got into hydro-devel? @stwirth are you planning on getting that merged (and released) before hydro gets EOL, or should we close this PR?

stwirth:
@mikeferguson close please!

PR #304, opened 2015-02-05
Global planner orientation

DLu:
Create a variety of options for outputting orientations from the global planner. 

By default, each point will have no orientation, which is the current behavior. 

The "Forward" mode orients the robot toward the next point on the path. 

The "Interpolation" mode interpolates between the first and last pose. 

The "Mixture" model points forward most of the time until the last straightaway, whereupon it starts interpolating to its final position.

### DIFF
--- a/global_planner/cfg/GlobalPlanner.cfg
+++ b/global_planner/cfg/GlobalPlanner.cfg
@@ -10,4 +10,13 @@ gen.add("neutral_cost", int_t,   0, "Neutral Cost",  50, 1, 255)
 gen.add("cost_factor", double_t, 0, "Factor to multiply each cost from costmap by", 3.0, 0.01, 5.0)
 gen.add("publish_potential", bool_t, 0, "Publish Potential Costmap", True)
 
+orientation_enum = gen.enum([ 
+    gen.const("None",        int_t, 0, "No orientations added except goal orientation"),
+    gen.const("Forward",     int_t, 1, "Orientations point to the next point on the path"),
+    gen.const("Interpolate", int_t, 2, "Orientations are a linear blend of start and goal pose"),
+                            ], "How to set the orientation of each point")
+
+gen.add("orientation_mode",  int_t, 0, "How to set the orientation of each point", 1, 0, 2,
+        edit_method=orientation_enum)
+
 exit(gen.generate(PACKAGE, "global_planner", "GlobalPlanner"))
--- a/global_planner/cfg/GlobalPlanner.cfg
+++ b/global_planner/cfg/GlobalPlanner.cfg
@@ -14,9 +14,10 @@ orientation_enum = gen.enum([
     gen.const("None",        int_t, 0, "No orientations added except goal orientation"),
     gen.const("Forward",     int_t, 1, "Orientations point to the next point on the path"),
     gen.const("Interpolate", int_t, 2, "Orientations are a linear blend of start and goal pose"),
+    gen.const("Mixture",     int_t, 3, "Forward orientation until last straightaway"),
                             ], "How to set the orientation of each point")
 
-gen.add("orientation_mode",  int_t, 0, "How to set the orientation of each point", 1, 0, 2,
+gen.add("orientation_mode",  int_t, 0, "How to set the orientation of each point", 1, 0, 3,
         edit_method=orientation_enum)
 
 exit(gen.generate(PACKAGE, "global_planner", "GlobalPlanner"))
--- a/global_planner/include/global_planner/orientation_filter.h
+++ b/global_planner/include/global_planner/orientation_filter.h
@@ -9,7 +9,8 @@ namespace global_planner {
 
 class OrientationFilter {
     public:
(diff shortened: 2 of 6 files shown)


global_planner/include/global_planner/orientation_filter.h
```diff
@@ -0,0 +1,31 @@
+/*********************************************************************
+ * Author: David V. Lu!!
```
mikeferguson, line 2: This file needs a BSD license header added

global_planner/src/orientation_filter.cpp
```diff
@@ -0,0 +1,81 @@
+
```
mikeferguson, line 1: Needs a license header

global_planner/cfg/GlobalPlanner.cfg
```diff
@@ -10,4 +10,14 @@ gen.add("neutral_cost", int_t,   0, "Neutral Cost",  50, 1, 255)
 gen.add("cost_factor", double_t, 0, "Factor to multiply each cost from costmap by", 3.0, 0.01, 5.0)
 gen.add("publish_potential", bool_t, 0, "Publish Potential Costmap", True)
 
+orientation_enum = gen.enum([ 
+    gen.const("None",        int_t, 0, "No orientations added except goal orientation"),
+    gen.const("Forward",     int_t, 1, "Orientations point to the next point on the path"),
+    gen.const("Interpolate", int_t, 2, "Orientations are a linear blend of start and goal pose"),
+    gen.const("Mixture",     int_t, 3, "Forward orientation until last straightaway"),
```
mikeferguson, line 17: Is there a better name than "Mixture"? (or is this based off some paper where that name was used -- if so, we should include the link to said paper)

global_planner/include/global_planner/orientation_filter.h
```diff
@@ -0,0 +1,31 @@
...
+ *********************************************************************/
+#ifndef _ORIENTATION_FILTER_H
+#define _ORIENTATION_FILTER_H
```
mikeferguson, line 5: These really should be GLOBAL_PLANNER_ORIENTATION_FILTER_H to comply with ROS style guide.

global_planner/src/orientation_filter.cpp
```diff
@@ -0,0 +1,81 @@
...
+            for(int i=0;i<n-1;i++){
+                pointToNext(path, i);
+            }
+            
+            int i=n-3;
+            double last = getYaw(path[i]);
+            while( i>0 ){
+                double new_angle = getYaw(path[i-1]);
+                double diff = fabs(angles::shortest_angular_distance(new_angle, last));
+                // ROS_INFO("%f %f (%d)", new_angle, diff, i);
```
mikeferguson, line 42: suggest either remove these, or use ROS_DEBUG_NAMED()

global_planner/src/orientation_filter.cpp
```diff
@@ -0,0 +1,81 @@
...
+                if( diff>0.35)
+                    break;
+                else
+                    i--;
+            }
+            // ROS_INFO("%f %f", path[i].pose.position.x, path[i].pose.position.y);
```
mikeferguson, line 48: remove or use ROS_DEBUG_NAMED()

global_planner/include/global_planner/planner_core.h
```diff
@@ -185,6 +186,7 @@ class GlobalPlanner : public nav_core::BaseGlobalPlanner {
         PotentialCalculator* p_calc_;
         Expander* planner_;
         Traceback* path_maker_;
+        OrientationFilter* o_filter_;
```
mikeferguson, line 189: This is only used in like 4 places -- so I would recommend making the name a bit more verbose (orientation_filter_) so people aren't having to look back and search for what the heck it is. (in the future, p_calc_ is another great variable name to update)

mikeferguson:
Sorry to take so long getting to this -- there are a couple issues I noted above with naming, logging and licensing (especially licensing -- those have to be fixed before merging). I would also suggest compressing this to a single commit since the intermediates don't appear to have much value in the history.

global_planner/cfg/GlobalPlanner.cfg
DLu, line 17: No, I made it up.

DLu:
Closed in favor of #309.

bxc237:
Hi, I tried the "Interpolation" mode on my omni-wheel robot with DWA local planner. It seems that the mode has no effect. The robot would still move to the goal first, then turn in place to obey the goal orientation (just like mode "None"). I am using kinetic version. Any advise or guidance would be greatly appreciated.

reinzor:
The `orientation filter` updates the orientation of the poses in the global plan. This does not necessarily mean that the local planner is using these orientation. If you are using the `dwa_local_planner` for example, this is not the case.

bxc237:
@reinzor Thanks for the explanation. Then which local planner should I choose to work with the orientation filter?

PR #307, opened 2015-02-11
Fix indexing bug in OccupancyGridUpdate callback function

Julse:
Wrong indexing for updated costmap area.


mikeferguson:
How does this even compile? I don't see where size_x comes from

Julse:
Hi,
you're right. It was supposed to be 'size_x_', which is the size of the costmap in x direction. 
I changed it.

mikeferguson:
@DLu wanta take a look at this?

DLu:
+1

PR #310, opened 2015-02-18
Fixing memory leak and clean up freeing operations

servos:
Main fix is to a memory leak that was present in the costmap_2d/inflation_layer. Also includes other clean up to explicitly free memory that was allocated and ensure memory is allocated when needed.

### DIFF
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -182,6 +182,7 @@ private:
   double resolution_;
 
   bool* seen_;
+  int seen_size_;
 
   unsigned char** cached_costs_;
   double** cached_distances_;
--- a/costmap_2d/plugins/inflation_layer.cpp
+++ b/costmap_2d/plugins/inflation_layer.cpp
@@ -55,6 +55,9 @@ InflationLayer::InflationLayer()
   , cell_inflation_radius_(0)
   , cached_cell_inflation_radius_(0)
   , dsrv_(NULL)
+  , seen_(NULL)
+  , cached_costs_(NULL)
+  , cached_distances_(NULL)
 {
   access_ = new boost::shared_mutex();
 }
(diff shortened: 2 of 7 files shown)


global_planner/src/dijkstra.cpp
```diff
@@ -49,6 +49,15 @@ DijkstraExpansion::DijkstraExpansion(PotentialCalculator* p_calc, int nx, int ny
     priorityIncrement_ = 2 * neutral_cost_;
 }
 
+DijkstraExpansion::~DijkstraExpansion() {
+  printf("Dijkstra deconstruct \n");
```
DLu: Do you have a strong reason for including this debug print?

costmap_2d/plugins/inflation_layer.cpp
```diff
@@ -292,19 +311,25 @@ void InflationLayer::deleteKernels()
...
   if (cached_costs_ != NULL)
   {
     for (unsigned int i = 0; i <= cached_cell_inflation_radius_ + 1; ++i)
     {
-      delete[] cached_costs_[i];
+      if (cached_costs_[i])
+        delete[] cached_costs_[i];
     }
-    delete[] cached_costs_;
+    if (cached_costs_)
```
DLu: Is this necessary? We just checked if cached_costs_ was null.

costmap_2d/include/costmap_2d/inflation_layer.h
```diff
@@ -194,4 +195,4 @@ class InflationLayer : public Layer
 
 }  // namespace costmap_2d
 
-#endif  // COSTMAP_2D_INFLATION_LAYER_H_
\ No newline at end of file
+#endif  // COSTMAP_2D_INFLATION_LAYER_H_
```
DLu, line 198: In the future, please try not to include these nonfunctional changes in functional PRs.

DLu:
Overall, looks good. Thanks for cleaning up the sloppiness.

mikeferguson:
Looks like this has been updated, for reference @servos please post a comment in the future when you push new commits -- otherwise we don't actually get any notice that commits have been pushed.

I'll merge this later today -- I want to wait for the current CI job to finish before adding another commit.

PR #312, opened 2015-02-21
Remove unused dependencies in move_base_msgs.

stonier:
The roscpp one in particular is heavy - this gets in the way of easily
generating messages from almost no dependencies.

### DIFF
--- a/move_base_msgs/CMakeLists.txt
+++ b/move_base_msgs/CMakeLists.txt
@@ -3,11 +3,9 @@ project(move_base_msgs)
 
 find_package(catkin REQUIRED
     COMPONENTS
-        roscpp
         message_generation
         actionlib_msgs
         geometry_msgs
-        std_msgs
         )
 
 
@@ -21,7 +19,6 @@ add_action_files(
 generate_messages(
     DEPENDENCIES
         actionlib_msgs
-        std_msgs
         geometry_msgs
 )
 
--- a/move_base_msgs/package.xml
+++ b/move_base_msgs/package.xml
@@ -15,12 +15,10 @@
 
     <buildtool_depend>catkin</buildtool_depend>
 
-    <build_depend>roscpp</build_depend>
     <build_depend>actionlib_msgs</build_depend>
     <build_depend>geometry_msgs</build_depend>
     <build_depend>message_generation</build_depend>
 
-    <run_depend>roscpp</run_depend>
     <run_depend>actionlib_msgs</run_depend>
     <run_depend>geometry_msgs</run_depend>
     <run_depend>message_runtime</run_depend>


stonier:
I'd actually like to pull these messages and others (e.g. robot_pose_ekf, map_store) into a separate message only repository. Something like [ros_planning_msgs](https://github.com/stonier/ros_planning_msgs).

For indigo, it's probably wise to only move packages that don't have an api change (e.g. move_base_msgs, but not robot_pose_ekf/map_store), but migrate them in for a future release. Could we make this happen? I'd be happy to do maintenance for it.

mikeferguson:
@stonier I suppose we could split out messages -- but just to be clear I think the only messages are move_base_msgs -- I don't see any messages in robot_pose_ekf or map_server as they use things from the common_msgs repo.

Could you give me a little more background on what you are trying to accomplish for a workflow?

Also, I think "ros_planning_msgs" is too broad for what is here. I actually think, for this case, breaking just move_base_msgs out into it's own repo would be the correct path -- would that work for you?

stonier:
- robot_pose_ekf [services](https://github.com/ros-planning/navigation/blob/indigo-devel/robot_pose_ekf/CMakeLists.txt#L25)
- map_store [messages](https://github.com/ros-planning/map_store/blob/hydro-devel/CMakeLists.txt#L13) and [services](https://github.com/ros-planning/map_store/blob/hydro-devel/CMakeLists.txt#L18)

Yeah, I agree, `ros_planning_msgs` feels too broad, considering navigation and moveit and others are relatively independant. Nonetheless, that's at least three packages worth of messages - be annoying to have them all as singleton repositories and `navigation_msgs` would be strange given that there is already a `common_msgs/nav_msgs` package.

With regards to explanation - quite often I'm just building message packages for other platforms or languages (e.g. windows/rosjava). These are often required for talking _with_ the navigation stack, but the navigation stack isn't required. Unfortunately with catkin there is no way to build them without having to build the entire stack which is oft impossible. Hence, msg only repositories are a good thing.

As for going forward, breaking out just `move_base_msgs` to a singleton package for indigo would be a good start for now. We'll be deprecating our own use of `map_store` shortly and I don't need `robot_pose_ekf` messages...and doing these would break existing indigo api anyway. If we need to, we can think about these other packages later.

stonier:
Shall I go ahead and set up a `move_base_msgs` ready for transfer to here?

mikeferguson:
@stonier please see https://github.com/ros-planning/navigation_msgs for navigation related messages for Jade+. I just created this as we needed to get map_msgs into a stable/maintained repository.

stonier:
Awesome :neckbeard:

PR #315, opened 2015-03-02
stops voxel_layer and obstacle_layer from deleting the same dynamic_reconfigure server

KaijenHsiao:
Without this fix, you get segfaults when destructing voxel_layer. 

A very basic test of voxel_layer that segfaults without the fix and doesn't with can be found here:
https://github.com/KaijenHsiao/navigation/tree/voxel_layer_tests

### DIFF
--- a/costmap_2d/include/costmap_2d/voxel_layer.h
+++ b/costmap_2d/include/costmap_2d/voxel_layer.h
@@ -94,7 +94,7 @@ private:
   virtual void raytraceFreespace(const costmap_2d::Observation& clearing_observation, double* min_x, double* min_y,
                                  double* max_x, double* max_y);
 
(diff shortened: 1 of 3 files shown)


mikeferguson:
I'm not convinced this is the right way to really fix this problem, since we really don't want to have two different dynamic reconfigure servers potentially getting setup. The simpler, and probably more correct fix, is to make sure that dsrv_ is set to NULL after it is deleted -- in both destructors. 

setupDynamicReconfigure() should also probably delete dsrv_ before creating it to avoid a memory leak but I'm not sure if initialize actually ever gets called more than once.

KaijenHsiao:
It only creates the voxel_layer dynamic reconfigure server if it is of type VoxelLayer.  The setup function for ObstacleLayer does not get used.  I actually tried setting it to NULL and somehow between the two destructors something in another thread (I didn't track down what) was changing it to not NULL much of the time.

PR #316, opened 2015-03-05
Accurate & Cached Bresenham raytracing implementation and clearing for corner cases

AlexReimann:
Changes / Fixes for #267 

Changes the bresenham algorithm to a more accurate version (before the starting points were at a fixed, incorrect position) based on:
http://graphics.idav.ucdavis.edu/education/GraphicsNotes/CAGDNotes/Bresenhams-Algorithm.pdf

Adds a cached version of the grid updating to speed up computation and also allow clearing corner cases as described in https://github.com/ros-planning/navigation/issues/267#issuecomment-61228679
Also allows now visualisation of which voxels get cleared.

Only tested with xtion yet. Please test throughly.

### DIFF
--- a/costmap_2d/cfg/VoxelPlugin.cfg
+++ b/costmap_2d/cfg/VoxelPlugin.cfg
@@ -19,5 +19,6 @@ combo_enum = gen.enum([ gen.const("Overwrite", int_t, 0, "b"),
 gen.add("combination_method", int_t, 0, "Method for combining two layers", 1, 0, 2, edit_method=combo_enum)
 gen.add("clear_corner_cases", bool_t, 0, "Whether to also clear corner cases or not, might over clear (but under clears without)", False)
 gen.add("accuracy_multiplier_bits", int_t, 0, "Number of bits which are used for increasing raytracing accuracy for clearing", 10, 0, 28)
+gen.add("use_cached_updating", bool_t, 0, "Whether to use a cached approach for updating which can be faster in case of 3d sensors and is needed for clear_corner_cases", False)
 
 exit(gen.generate("costmap_2d", "costmap_2d", "VoxelPlugin"))
--- a/costmap_2d/cfg/VoxelPlugin.cfg
+++ b/costmap_2d/cfg/VoxelPlugin.cfg
@@ -17,5 +17,7 @@ combo_enum = gen.enum([ gen.const("Overwrite", int_t, 0, "b"),
                       "Method for combining layers enum")
 
 gen.add("combination_method", int_t, 0, "Method for combining two layers", 1, 0, 2, edit_method=combo_enum)
+gen.add("clear_corner_cases", bool_t, 0, "Whether to also clear corner cases or not, might over clear (but under clears without)", False)
+gen.add("accuracy_multiplier_bits", int_t, 0, "Number of bits which are used for increasing raytracing accuracy for clearing", 10, 0, 28)
 
 exit(gen.generate("costmap_2d", "costmap_2d", "VoxelPlugin"))
--- a/costmap_2d/include/costmap_2d/obstacle_layer.h
+++ b/costmap_2d/include/costmap_2d/obstacle_layer.h
@@ -169,6 +169,8 @@ protected:
   
   int combination_method_;
 
+  double max_raytrace_range_;
+
 private:
   void reconfigureCB(costmap_2d::ObstaclePluginConfig &config, uint32_t level);
 };
@@ -176,3 +178,4 @@ private:
 }  // namespace costmap_2d
 
 #endif  // COSTMAP_2D_OBSTACLE_LAYER_H_
+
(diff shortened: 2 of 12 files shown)


mikeferguson:
@AlexReimann I will get testing this shortly.

AlexReimann:
@mikeferguson 
Still no time yet?

AlexReimann:
Is it better / will the process be faster if I do a pull request for Jade?

phil0stine:
@AlexReimann We have just run the code and do observe that obstacles are cleared as advertised, however we have also observed crashing after 5-10 minutes of operation. 

We have not debugged it yet other than looking at memory (which looks steady), but wanted to report. Our setup has on the order of 300k pts/sec, and the crash seems to occur for both cached and simple clear settings.

phil0stine:
Apparently we discovered an edge case. We still need to investigate how this occured from our data, but there were rare instances where the input `number_of_steps` to the 
[bresenham3DIncludingCorners](https://github.com/AlexReimann/navigation/blob/ce65f90ad8ee03205c2ea2704a53f15a2df567b2/voxel_grid/include/voxel_grid/voxel_grid.h#L392) function was zero (which then got decremented to start the loop).

There are a number of solutions for this, but a minimal patch is below. The function will still loop the same number of iterations but handle the edge case. We have tested it on our system with no crashes.

``` diff
diff --git a/voxel_grid/include/voxel_grid/voxel_grid.h b/voxel_grid/include/voxel_grid/voxel_grid.h
index b8c2b86..55dbfd6 100644
--- a/voxel_grid/include/voxel_grid/voxel_grid.h
+++ b/voxel_grid/include/voxel_grid/voxel_grid.h
@@ -398,7 +398,7 @@ private:

     //-1 because we don't want to clear the corners _after_ the last cell
     //last cell is cleared after the loop
-    for (unsigned int i = 0; i < number_of_steps - 1; ++i)
+    for (unsigned int i = 1; i < number_of_steps; ++i)
     {
       (*clearer)(offset, z_mask);
       off_a(offset_a);
```

AlexReimann:
I was not able to re-produce the error. For me it makes more sense to change then number_of_steps to a int. This will also stop the execution of the for loop in case of number_of_steps == 0.

Looking at the code the error should only occur if the ray start an end position is the same. Which should be a unlikely but valid thing. Especially because the [coordinates get scaled](https://github.com/ros-planning/navigation/blob/indigo-devel/costmap_2d/plugins/voxel_layer.cpp#L305) this can happen.

@phil0stine 
Do you have any feedback on the computation time?

phil0stine:
@AlexReimann 
Here is a runtime comparison of running this PR versus the jade-devel stack e8b14f0e667fbef19088b20a13701db15703ef1f . The runtimes profile the call to [raytraceFreeSpace](https://github.com/AlexReimann/navigation/blob/f653514118189d404b1101237da241a36da698f9/costmap_2d/plugins/voxel_layer.cpp#L164) in `voxel_layer.cpp`.

We have 3 sensors (2 laser scanners and a velodyne). We separated out the clearing runtimes for each sensor, and then combined them for the bottom plots. The takeaway is that it appears the cached clearing does not pay off for a small number of clearing points, but that it does for larger sized point clouds. 

For the laser scanners, the cached clearing ("clearing_fix") is slower, while for the velodyne with more points, the runtimes are the same. I have not tested with even larger point clouds to see if the caching offers a runtime improvement. 

In any case, its worth it to us to use the cached algorith since it clearly does a better job of clearing spurious voxels.

clearing_fix_time_comparison_per_sensor

AlexReimann:
I could switch the caching from caching "observations" one by one to actually cache all observations and than update the costmap. This should give a speedup when using multiple sensors with the same layer / using multiple buffered observations.

I could put that into another patch after this gets merged, if you would prefer that.

Besides that, should I update the code to resolve the conflicts?

AlexReimann:
I found two bugs, and will include the fixes soon™  and also update it, so it can be merged.

bradleypowers:
@AlexReimann: What is the status of these fixes?  We're noticing the same issues on our robot (which is using a laser, but with a Voxel Layer obstacle layer), with 1 cell obstacles remaining after clearing.  I'd be excited to help test/debug...

AlexReimann:
I currently don't have access to a robot / 3d sensor (and will not have for some time)
I copied the fixes over we used at yujin. But I also made some other bigger changes in that repo, so i am not sure if this does not lead to segfaults.
I'll see if I can get hands on a robot to test it this month..

AlexReimann:
Closed because out of date, conflicts and personally not longer working on it

nickvaras:
> @AlexReimann: What is the status of these fixes? We're noticing the same issues on our robot (which is using a laser, but with a Voxel Layer obstacle layer), with 1 cell obstacles remaining after clearing. I'd be excited to help test/debug...

@bradleypowers , were you ever able to solve this? I tried porting the improved Bresenham to Kinetic, and, as you say, some obstacles are cleared until a single voxel is left. It's hit or miss, sometimes things get cleared properly... I'd appreciate any updates on this.

PR #246, opened 2014-09-03
Static layer works with rolling window now

reinzor:
Fixes #197

PR #323, opened 2015-03-27
Static layer works in rolling window

mikeferguson:
This cherry-picks from #246, fixes an additional issue with the layeredcostmap calling matchSize() and making the static map all wrong, and applies some small style fixes for future maintainability of code.

Jade-targeted PR coming in a moment


mikeferguson:
@DLu this is tested against my current robot config -- and works well. Please review in case you see anything.

PR #329, opened 2015-04-23
Add a first_map_only parameter so we keep reusing the first received static map

jihoonl:
This is follow-up of #150 to use ros-navi stack for multi-master usecase.

### DIFF
--- a/costmap_2d/plugins/static_layer.cpp
+++ b/costmap_2d/plugins/static_layer.cpp
@@ -65,8 +65,8 @@ void StaticLayer::onInitialize()
 
   std::string map_topic;
   nh.param("map_topic", map_topic, std::string("map"));
+  nh.param("first_map_only", first_map_only_, false);
   nh.param("subscribe_to_updates", subscribe_to_updates_, false);
-
   nh.param("track_unknown_space", track_unknown_space_, true);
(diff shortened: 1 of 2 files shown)


costmap_2d/plugins/static_layer.cpp
```diff
@@ -150,6 +158,9 @@ unsigned char StaticLayer::interpretValue(unsigned char value)
 
 void StaticLayer::incomingMap(const nav_msgs::OccupancyGridConstPtr& new_map)
 {
+  if (first_map_only_)
+    first_map_.reset(new nav_msgs::OccupancyGrid(*new_map));
```
mikeferguson, line 162: this would seem to make every map the first map? Only by stopping and starting costmap updates would the map_sub_ be shutdown and the first_map actually be locked in

mikeferguson:
It would seem this could be made a lot simpler -- at the end of the map callback you could check if first_map_only_ was true, and if so, shutdown the subscriber and set a flag that map has been received. Then in the initialize, if first_map_only_ and the first map flag are set, you would simply mark has_updated_data_ as true rather than starting the subscriber (this would have UpdateBounds update properly on the next update cycle)

jihoonl:
Thanks. I will recreate the PR with updated code.

PR #330, opened 2015-04-27
proper locking during costmap update

mikeferguson:
This addresses possible empty costmap issue mentioned in https://github.com/ros-planning/navigation/issues/187#issuecomment-45051376

In order to properly hold costmap when calling resetMap, we need a recursive mutex. Since only costmap publisher was using a shared_lock, switching from shared_mutex to recursive should have no (runtime) downsides.

I've also changed getLock to getMutex -- since it actually returns a mutex, not a lock. In addition I have typedef'd the mutex type so that if we later change this again, it is less difficult to update everything properly.


DLu:
I have not tested this myself, but +1 because I don't actually know how to best write parallel C++.

PR #336, opened 2015-05-06
Add a first_map_only parameter so we keep reusing the first received static map

jihoonl:
It is to follow up #150 and #329

@corot's note from old commit.

We need this for a rather exotic case: when using https://github.com/robotics-in-concert/rocon_multimaster/tree/hydro-devel/rocon_gateway, the first time we subscribe we get the map, but not in subsequent subscriptions due to clearing costmaps.
An easier solution is just to call shutdown in the map subscriber, but we still have a big delay until receiving the map, so this flag is much better option.
I didn't verify that it doesn't interfere with map updates (because is the first time I see this feature). Appart from this, we have the change running happily in our robots.

### DIFF
--- a/costmap_2d/include/costmap_2d/static_layer.h
+++ b/costmap_2d/include/costmap_2d/static_layer.h
@@ -87,6 +87,7 @@ private:
   unsigned int x_,y_,width_,height_;
   bool track_unknown_space_;
   bool use_maximum_;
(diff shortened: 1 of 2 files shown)


mikeferguson:
This looks quite a bit better than the previous PR. For reference, please make future PRs against jade and then we can backport to indigo

jihoonl:
Alright I will. Thanks.

Issue #206, opened 2014-06-09
When using robot_radius, footprint is incorrectly set by costmap2d_ros


mikeferguson:
When using robot_radius, the footprint gets set by move_base to [] if the footprint parameter is not set. This means that the second time you launch the navigation stack, the footprint will be improper and an error appears:

```
[ERROR] [1402335344.906978220, 66.013000000]: You must specify at least three points for the robot footprint, reverting to previous footprint.
```

My current work around has been to add this to the launch file:

```
<rosparam command="delete" ns="move_base" />
```

mikeferguson:
fixed by #343 in indigo-devel, #344 in jade-devel

tutorgaming:
The bug does persist when you close the navigation stack but not the roscore, With the parameter "robot_radius" set only .  
- ROS Kinetic
  ![aa](https://cloud.githubusercontent.com/assets/5821937/19683147/46cc88e2-9adb-11e6-9c51-88beceab4069.png)

nitin5:
Is it solved?

DLu:
There was a fix that was not ported to kinetic/lunar. It is addressed by #663.

PR #343, opened 2015-05-16
Look for robot_radius when footprint is not set. #206

patrickcjh:


mikeferguson:
Looks good, tests pass. Thanks for fixing this very long open bug!

Issue #357, opened 2015-06-09
A* produces plans too close to walls


tavu:
A\* ignores the factor property of the expander. As a result it produces plans too close to walls and obstacles.

PR #358, opened 2015-06-09
multipling cost by a factor

tavu:
fixing issue #357

### DIFF
--- a/global_planner/include/global_planner/expander.h
+++ b/global_planner/include/global_planner/expander.h
@@ -81,7 +81,7 @@ class Expander {
                 int n = startCell+i+nx_*j;
                 if(potential[n]<POT_HIGH)
                     continue;
-                float c = costs[n]+neutral_cost_;
+                float c = factor_*costs[n]+neutral_cost_;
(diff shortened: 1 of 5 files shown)


mikeferguson:
@DLu please review

DLu:
This highlights some oversights in how the code was refactored, so thank you for that. However, it introduces a couple additional inconsistencies to where factor is applied. 

I'm not sure why I put it there, but the costs should go through the getCost function (https://github.com/ros-planning/navigation/blob/jade-devel/global_planner/include/global_planner/dijkstra.h#L86) that's in dijkstra.h. 

I'll add a few more comments later, and possibly a pull request of my own.

DLu:
Ok, I've used this to create two new branches. 

https://github.com/DLu/navigation/tree/fix357
This uses the getCost function more centrally to fix the places where the factor was not used. 

https://github.com/DLu/navigation/tree/floated_costs
This changes the header of the potential calculator to use floats. 

Do these two changes look to cover your use case?

DLu:
@tavu Likely going to close this if I don't hear back. Also since hydro has since closed up.

PR #362, opened 2015-06-26
Added CIRCUMSCRIBED_INFLATED_OBSTACLE constant

alexhenning:


DLu:
I'm guessing the change is motivated by the diagram here: http://wiki.ros.org/costmap_2d#Inflation

However, there isn't currently a way to fix the circumscribed cost, since all values less than INSCRIBED_INFLATED_OBSTACLE are dictated by the inscribed_radius and inflation_radius (https://github.com/ros-planning/navigation/blob/jade-devel/costmap_2d/include/costmap_2d/inflation_layer.h#L113). Thus the value will not always be 128. 

There's an argument for adding a way to figure out the circumscribed cost, but it will require more than this header change.

alexhenning:
@DLu You're right, I was basing it off that diagram. I've updated the description on the wiki to clarify that 128 is just an example.

Issue #289, opened 2014-11-29
Get Rid of footprint_layer


DLu:
Originally, the footprint clearing was done in its own layer, since it was a different kind of functionality than adding/clearing obstacles. 

However, having the footprint clearing in its own layer does not permanently remove the obstacle, which means that when the footprint moves, the obstacles reappear. 

This was temporarily fixed in 0e544b1276f26a8e96f0613a53103a136f11d957, but should be changed so it doesn't create the weird layer namespaces.

skaynama:
Is permanently clearing the obstacle necessarily a good solution? I'm just thinking what if the robot rolls over the inflated obstacle (e.g. due to cross track or localization error). Having the footprint layer temporarily clear the obstacle makes sense because otherwise the robot's position would be invalid and any attempt to replan using common planners would fail (bad initial condition). Removing the obstacle permanently in this case is obviously wrong and undesirable. What are your thoughts?

DLu:
The use case for the footprint layer is for imperfect sensor filtering of the robot itself. If your sensor happens to see a lethal obstacle within its footprint (such as its own hand, perhaps) then all resulting trajectories will be invalid because they will collide with that point. 

The inflation layer is handled AFTER the obstacle layer (and the footprint clearing), thus the inflated obstacle will not be removed. 

The motivation of removing the lethal points from the obstacle layer's costmap and not just the master costmap is so that when the robot moves to a new location, the imperfectly sensed point on the robot will not still be there.

skaynama:
What if the obstacle was a legitimate obstacle but it was sensed only because of bad localization (the robot thought an obstacle is within its footprint because the robot's position estimate had error). Say we clear that obstacle permanently, and move away and regain perfect localization. Then there would be a gap in place of the legitimate obstacle -- which is clearly undesirable. How would this scenario be handled?

DLu:
That's a plausible enough scenario. The gap would probably get filled in again though when it was resensed. And there's still a flag and plenty of other ways to customize the behavior if this sort of behavior doesn't work for you. This issue is primarily about getting rid of the footprint_layer object.

PR #364, opened 2015-06-28
Remove Footprint Layer (and organize footprint functions) 

DLu:
Rationale described here: #289

My apologies for the late inclusion for the jade release. 

This PR does three key things: 
1. Removes the independent footprint layer and integrates the functionality with the obstacle (and voxel) layer. 
2. Moves the footprint-specific functions from costmap_2d_ros to footprint.cpp
3. Moves the footprint publisher to the costmap_2d_ros object

One consequence of (3) is that I simplified the parameters slightly for the footprints, in that I've eliminated the footprint_param parameter and instead just had the incoming parameter named "footprint_spec" and the published footprint called "footprint".


costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -127,25 +126,11 @@ Costmap2DROS::Costmap2DROS(std::string name, tf::TransformListener& tf) :
...
   // subscribe to the footprint topic
-  std::string topic_param, topic;
-  if(!private_nh.searchParam("footprint_topic", topic_param))
-  {
-    topic_param = "footprint_topic";
-  }
-
-  private_nh.param(topic_param, topic, std::string("footprint"));
-  footprint_sub_ = private_nh.subscribe(topic, 1, &Costmap2DROS::setUnpaddedRobotFootprintPolygon, this);
+  footprint_sub_ = private_nh.subscribe("footprint_spec", 1, &Costmap2DROS::setUnpaddedRobotFootprintPolygon, this);
```
mikeferguson: the parameter here might be valuable -- what if you wanted to use different footprint topics for different costmaps within the same node?

costmap_2d/include/costmap_2d/footprint.h
```diff
@@ -40,18 +40,106 @@
...
+std::vector<geometry_msgs::Point> makeFootprintFromRadius( double radius );
+
+/** 
+ * @brief Make the footprint from the given string.
+ *
+ * Format should be bracketed array of arrays of floats, like so: [[1.0, 2.2], [3.3, 4.2], ...] 
+ * 
+ * If there's an error, the error_flag will be set to true. 
+ */
+std::vector<geometry_msgs::Point> makeFootprintFromString( const std::string& footprint_string, bool* error_flag );
```
mikeferguson: this seems an odd API -- and the usage below could have unintentened consequences. I would suggest that this be changed so that the vector of points is passed in by reference, the function returns bool, and then when used you can say if(makeFootprintFromString(string, footprint)) { setUnpaddedRobotFootprint(footprint) } else { print error message }

costmap_2d/include/costmap_2d/footprint.h
```diff
@@ -40,18 +40,106 @@
...
+ *
+ * @param footprint_xmlrpc should be an array of arrays, where the
+ * top-level array should have 3 or more elements, and the
+ * sub-arrays should all have exactly 2 elements (x and y
+ * coordinates).
+ *
+ * @param full_param_name this is the full name of the rosparam from
+ * which the footprint_xmlrpc value came.  It is used only for
+ * reporting errors. */
+std::vector<geometry_msgs::Point> makeFootprintFromXMLRPC( XmlRpc::XmlRpcValue& footprint_xmlrpc,
```
mikeferguson, line 134: should footprint_xmlrpc be const as well?

mikeferguson:
Looking at this change, I also wonder if it makes sense to backport to Indigo (we would of course have to make the parameter changes backwards-compatible). Given the lack of documentation about the footprint layer it seems that fixing this in Indigo would also be beneficial. It would break ABI, but then again, the costmap mutex fixes that I recently did (which were very much needed) also broke ABI.

One other note is that much of the added code does not follow the ROS style guide. In particular, there are a number places where there are extra spaces inside paranthesis, there are a couple "cuddled braces", and indents are not always 2 spaces. I realize large portions of the navigation stack do not conform to the style guide, but it would generally be nice to move towards the style guide with each new PR.

costmap_2d/include/costmap_2d/footprint.h
DLu, line 134: For some reason involving the ROS debug macros, adding const (`makeFootprintFromXMLRPC( const XmlRpc::XmlRpcValue& footprint_xmlrpc,`) results in a number of weird errors.

DLu:
I have updated the PR with fixes for the parameter searching and error API. Once we resolve #365, I'll fix the formatting on this one too. 

I am mostly fine porting to indigo. The largest thing would be that people's footprints would disappear in rviz, since that topic has changed.

mikeferguson:
@DLu linting changes are now merged -- this now needs to be rebased (and preferably squashed).

mikeferguson:
Replaced by new PR

PR #368, opened 2015-07-03
Remove Footprint Layer

DLu:
Rebased/squashed version of #364 

Fix for #289


mikeferguson:
I just ran the costmap_2d tests to make sure this would be ok -- seems there is a regression (or maybe a test needs to be updated):

```
[costmap_2d.rosunit-footprint_test/footprint_empty][FAILURE]--------------------
/home/fergs/tmp/nav_test/src/navigation/costmap_2d/test/footprint_tests.cpp:156
The difference between 0.0f and footprint[ 0 ].y is 0.0099999997764825821, which exceeds 0.0001, where
0.0f evaluates to 0,
footprint[ 0 ].y evaluates to 0.0099999997764825821, and
0.0001 evaluates to 0.0001.
```

DLu:
That was a legitimate broken test. Should be fine now.

mikeferguson:
Where is sign0() defined?

DLu:
costmap_math.h

mikeferguson:
huh -- github search didn't turn that up, thanks

PR #369, opened 2015-07-03
Remove Footprint Layer (Indigo)

DLu:
Fix for #289 backported for indigo. (jade version: #368)


mikeferguson:
@DLu does this need the same sign fix?

DLu:
Yes. Updated.

PR #371, opened 2015-07-08
More tolerant initial pose transform lookup [fake-localisation : indigo]

stonier:
If using the initial pose timestamp it will quite often be later than
the latest base -> map transform information (depends on how base -> odom
is being broadcasted). This happens with gazebo and in some robot setups.
In this situation, you get a future timestamp lookup error, e.g.

```
Failed to lookup dude transform!(Lookup would require extrapolation into
the future.  Requested time 8896.400000000 but the latest data is at time
8896.390000000, when looking up transform from frame [map] to frame
[base_footprint])
```

Note this is still not as complete as the amcl_node implementation, which
does a lookup to determine how much the robot has moved as well. However
it at least actually gets us somewhere.


mikeferguson:
@stonier thanks for opening PRs for both branches!

Issue #374, opened 2015-07-13
Deadlock on move_base's dynamic reconfigure of base_global_planner parameter


MartinPeris:
move_base freezes when dynamically changing the global planner plugin.

Steps to reproduce the problem:

1 - Start the navigation stack in your favorite robot.
2 - Bring up rqt_reconfigure gui: rosrun rqt_reconfigure rqt_reconfigure
3 - Change the parameter "move_base/base_global_planner" (you can use carrot_planner/CarrotPlanner for instance)
4 - Observe how move_base freezes until doom's day.

MartinPeris:
In the portion of code inside MoveBase::reconfigureCB that handles the change of base_global_planner parameter there is this:

 boost::unique_lockboost::mutex lock(planner_mutex_);
 ....
 resetState();
 ....
 lock.unlock();

The problem was that the call resetState() locks again the planner_mutex_ producing a deadlock condition.

I didn't have the time to deeply think about the use of mutex here, so I just unloked the mutex right before resetState() and locked it back again right after it.

Issue fixed.

PR #375, opened 2015-07-13
Solved deadlock on move_base dynamic reconfigure of base_global_plann…

MartinPeris:
…er parameter. This solves issue #374


mikeferguson:
https://github.com/ros-planning/navigation/pull/501 is a safer fix

Issue #333, opened 2015-04-29
Publishing Rolling Costmaps Incrementally


DLu:
[As seen here](http://answers.ros.org/question/207755/rviz-doesnt-update-rolling-window-costmap-origin/)

Using [OccupancyGridUpdate](https://github.com/ethz-asl/map_msgs/blob/master/msg/OccupancyGridUpdate.msg) will publish the data of the costmap incrementally, but there is currently no way using that message to update the [OccupancyGrid's origin pose](http://docs.ros.org/api/nav_msgs/html/msg/MapMetaData.html)

That data could also be published as a Pose message, but that makes a three-topic channel from what used to be one (bloated) channel.

DLu:
@stephanemagnenat Thoughts?

stephanemagnenat:
Updating the pose would imply shifting all data. What use case do you envision?

PR #377, opened 2015-07-16
Republish costmap if origin changes

DLu:
Addresses bug #333. 

Also fixes a small floating point error. Comparing the double resolution to the float32 msg.info.resolution failed in some cases because of the lack of precision (0.0250000004 vs. 0.0250000000). Rather than change all the resolution types, changing the resolution to a float before comparison fixes the problem.


mikeferguson:
LGTM -- can you also open a PR on indigo-devel

PR #388, opened 2015-08-24
No more ghosts in the inflation layer

stonier:
Previous bounds would fit the sensor measurements, and the inflation layer would clear
out to these, but leave 'ghosts' behind. These ghosts are from two sources - 1) the
inflation radius and 2) whole obstacles left behind as the robot has moved from the last point.

The modifications here remember the last bounds and set the new bounds so that a box at least
large enough to incorporate the old bounds plus the inflation radius is generated.

The image below is an example of the problem this fixes. The conditions of the scenario:
- Viewing global costmap in rviz
- Laser Scanner on a modified ObstacleLayer which clears the costmap every cycle.
- InflationLayer

leftovers

### DIFF
--- a/costmap_2d/plugins/inflation_layer.cpp
+++ b/costmap_2d/plugins/inflation_layer.cpp
@@ -35,6 +35,7 @@
  * Author: Eitan Marder-Eppstein
  *         David V. Lu!!
  *********************************************************************/
+#include <algorithm>
 #include <costmap_2d/inflation_layer.h>
 #include <costmap_2d/costmap_math.h>
 #include <costmap_2d/footprint.h>
@@ -58,6 +59,10 @@ InflationLayer::InflationLayer()
   , seen_(NULL)
   , cached_costs_(NULL)
   , cached_distances_(NULL)
+  , last_min_x_(-std::numeric_limits<float>::max())
+  , last_min_y_(-std::numeric_limits<float>::max())
+  , last_max_x_(std::numeric_limits<float>::max())
+  , last_max_y_(std::numeric_limits<float>::max())
 {
   access_ = new boost::shared_mutex();
 }
@@ -128,6 +133,10 @@ void InflationLayer::updateBounds(double robot_x, double robot_y, double robot_y
 {
   if (need_reinflation_)
   {
+    last_min_x_ = *min_x;
+    last_min_y_ = *min_y;
+    last_max_x_ = *max_x;
+    last_max_y_ = *max_y;
     // For some reason when I make these -<double>::max() it does not
     // work with Costmap2D::worldToMapEnforceBounds(), so I'm using
     // -<float>::max() instead.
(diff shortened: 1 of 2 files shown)


stonier:
Hold on this one for a sec, I need to check the ramifications of the proposed change under a couple of different use case scenarios that I didn't test.

stonier:
Ok, so I realised that our use case here was a bit different to the norm. It was a customised layer that clears itself every iteration, so the symptom is really obvious.

Nonetheless, ghost marks can still be left oustide the bounding box if for instance, the robot is at rest and someone walks along just inside the perimeter of the box. Ghost inflations will be left trailing along the outside of the box. This can eventually end up effecting a global planner. An example of this situation is shown below where an obstacle moves away from the robot with a scan sensor that has a narrow arc in front of it (about 60 degrees). This PR remedies it.

walking_along_the_perimeter

stonier:
Good to go.

costmap_2d/plugins/inflation_layer.cpp
```diff
@@ -126,8 +127,14 @@ void InflationLayer::matchSize()
 void InflationLayer::updateBounds(double robot_x, double robot_y, double robot_yaw, double* min_x,
                                            double* min_y, double* max_x, double* max_y)
 {
+  static double last_min_x =  -std::numeric_limits<float>::max();
```
mikeferguson: won't this variable be shared across all instances of the InflationLayer -- for instance, both the local and global. Maybe pluginlib saves us from the global/local planner sharing the same values, but this should be a class member variable rather than a static, right?

costmap_2d/plugins/inflation_layer.cpp
```diff
@@ -136,6 +143,17 @@ void InflationLayer::updateBounds(double robot_x, double robot_y, double robot_y
     *max_x = std::numeric_limits<float>::max();
     *max_y = std::numeric_limits<float>::max();
     need_reinflation_ = false;
+  } else {
```
mikeferguson: while this file is not 100% compatible with ROS style guide, we've been striving to make new code compatible. Therefore, opening and closing braces should be on their own lines

costmap_2d/plugins/inflation_layer.cpp
```diff
@@ -126,8 +127,14 @@ void InflationLayer::matchSize()
...
                                            double* min_y, double* max_x, double* max_y)
 {
+  static double last_min_x =  -std::numeric_limits<float>::max();
+  static double last_min_y = -std::numeric_limits<float>::max();
+  static double last_max_x = std::numeric_limits<float>::max();
+  static double last_max_y = std::numeric_limits<float>::max();
+
   if (need_reinflation_)
   {
+    last_min_x = *min_x; last_min_y = *min_y; last_max_x = *max_x; last_max_y = *max_y;
```
mikeferguson: each of these statements really should be on their own line

costmap_2d/plugins/inflation_layer.cpp
```diff
@@ -136,6 +143,17 @@ void InflationLayer::updateBounds(double robot_x, double robot_y, double robot_y
     *max_x = std::numeric_limits<float>::max();
     *max_y = std::numeric_limits<float>::max();
     need_reinflation_ = false;
+  } else {
+    double tmp_min_x = last_min_x; double tmp_min_y = last_min_y; double tmp_max_x = last_max_x; double tmp_max_y = last_max_y;
+    last_min_x = *min_x; last_min_y = *min_y; last_max_x = *max_x; last_max_y = *max_y;
```
mikeferguson: each of these statements really should be on their own line

stonier:
Hol, that was an embarrassing noobie thing to do. Thanks for catching it Mike.

mikeferguson:
@stonier if you could squash these two commits (on each PR), these look good for merge

stonier:
Squashed.

mikeferguson:
No idea why I never got back to merging these after the squash -- merging now

PR #389, opened 2015-08-24
No more ghosts in the inflation layer

stonier:
Previous bounds would fit the sensor measurements, and the inflation layer would clear
out to these, but leave 'ghosts' behind. These ghosts are from two sources - 1) the
inflation radius and 2) whole obstacles left behind as the robot has moved from the last point.

The modifications here remember the last bounds and set the new bounds so that a box at least
large enough to incorporate the old bounds plus the inflation radius is generated.

More detail in #388 for jade.

### DIFF
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -186,6 +186,7 @@ private:
 
   unsigned char** cached_costs_;
   double** cached_distances_;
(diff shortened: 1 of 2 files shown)


stonier:
Hold on this one for a sec, I need to check the ramifications of the proposed change under a couple of different use case scenarios that I didn't test.

stonier:
Good to go. See #388 for updated details.

mikeferguson:
See comments in #388

stonier:
Squashed.

PR #390, opened 2015-08-24
base_local_planner: some fixes in goal_functions

galou:
Prevent potential segmentation faults.


mikeferguson:
Good catch, thanks for the fixes!

PR #392, opened 2015-08-27
base_local_planner: reduce size of try statements

galou:


DLu:
@galou What is the motivation behind reducing the size of the try statements?

galou:
It's quite so time ago but I wanted to make the code clearer by putting only code that may generate one of the exception we want to catch inside the try statement. I may have missed some exception in the code that I transferred outside of the loop, though.

mikeferguson:
@galou thank you for your contribution. Unfortunately, it seems that none of us have had the time to fully ensure that we aren't missing any exceptions here. Given that, I'm going to suggest we accept that sometimes "tested code" can be better than "clean code", and close this PR. Any objections or new details about testing/review?

PR #395, opened 2015-08-29
pass in action types as references for costmap algorithms.

stonier:


mikeferguson:
See comments on #396

PR #396, opened 2015-08-29
Pass in action types as references for costmap algorithms.

stonier:

### DIFF
--- a/costmap_2d/include/costmap_2d/costmap_2d.h
+++ b/costmap_2d/include/costmap_2d/costmap_2d.h
@@ -357,7 +357,7 @@ protected:
    * @param  max_length The maximum desired length of the segment... allows you to not go all the way to the endpoint
    */
   template<class ActionType>
-    inline void raytraceLine(ActionType at, unsigned int x0, unsigned int y0, unsigned int x1, unsigned int y1,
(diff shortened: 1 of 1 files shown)


mikeferguson:
At first glance this seems like a good idea -- however, I think we need to dig into the API/ABI implications this might have (especially since much of the costmap stuff is done via plugins).

stonier:
Aye. Curious about this too.

For usage, the only point this would break that I can see is if someone passed a pointer in, but that would fail previously anyway due to the calls made on the object internally.

For ABI - how exactly do you go about testing this? Is there software that can verify it? Or is the best that can be done simply experimentally verifying an old executable against a newly built version of the navigation library?

Lots of discussion on google about ABI, but hard to find clear points on what breaks ABI and what doesn't. I imagine this is also quite dependant on c++ implementations.

At the end of the day, I'm not averse to having this go into a developmental branch, but be nice to have here and tested/used earlier rather than later if it isn't necessary to delay.

mikeferguson:
Daniel, I'm certainly not relagating this to a development branch -- I just want to closely look over the implications and have not yet had a chance to do that, I will try to before the end of the week.

stonier:
Ok, let me know if there's anything I can do.

mikeferguson:
@stonier -- After detailed review, I'm pretty sure this would cause an ABI change, so can you open a PR against Kinetic for this so we can get it in there (since indigo/jade are already ABI locked)

PR #403, opened 2015-10-16
recompute cluster stat when force_publication

lgeo3:
When force_publication is True, and no resample has been done the statistic
about the current clusters of particles are not recomputed. So amcl published
an outdated information.

This commit introduces a call to `pf_cluster_stats` when required (i.e no
resample but force_publication).


laurent-george:
seems to introducced bug (I have to check)

PR #407, opened 2015-12-10
Big version

cinchcircuits:


mikeferguson:
While I'm happy to look at new improvements, this is completely un-reviewable and un-mergeable (and also changes things like package versions, which cannot be merged). Please break out individual improvements into several smaller pull requests

cinchcircuits:
That should not have gone in, sorry about that.

PR #410, opened 2015-12-18
Fixes an issue with having multiple static layers

alexhenning:
If you have a static layer in both the local and global costmaps that
use the same map topic, there is a race condition that can cause the
static layer to get stuck after printing `Requesting map....`. This race
condition seems to be due to the call to shutdown in deactivate and how
the NodeHandle handles multiple subscribers under the hood.

This issue appears to happen about 1 in 1000 times in the setup I was
testing. This fix has never failed in over 1000000 tests. Instead of
calling activate and deactivate, the publisher is only recreated if the
topic has changed. Otherwise, it reuses the old setup.

@DLu can you take a look at this?

### DIFF
--- a/costmap_2d/plugins/static_layer.cpp
+++ b/costmap_2d/plugins/static_layer.cpp
@@ -79,25 +79,31 @@ void StaticLayer::onInitialize()
 
   lethal_threshold_ = std::max(std::min(temp_lethal_threshold, 100), 0);
   unknown_cost_value_ = temp_unknown_cost_value;
-  // we'll subscribe to the latched topic that the map server uses
-  ROS_INFO("Requesting the map...");
-  map_sub_ = g_nh.subscribe(map_topic, 1, &StaticLayer::incomingMap, this);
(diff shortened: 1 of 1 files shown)


mikeferguson:
@DLu we spent quite a bit of time discussing this in-house, and I can't come up with any use case this breaks -- but I figured we would circle you in for an opinion as well

mikeferguson:
@alexhenning please open a similar PR for jade-revel

alexhenning:
@mikeferguson jade PR is #415

palmieri:
Does this change mean that when a new map (published on the same initial "map" topic) is received, it is not properly updated? 
E.g. If we reset the static layer, the deactivate function will shutdown the initial subscriber. Meanwhile if the map server publishes a new map on the same topic there will be no subscriber that will read the new map. 

A solution for the original problem is to move the initialization of the map_received_  to false at the beginning of OnInitialize() method before the subscriber is reinitialized.

Best,

Luigi

PR #412, opened 2015-12-30
install wtf.py and add gps information to the status

k-okada:
- add execute bit to scripts/wtf.py
- add to install script/ directory
- robot_pose_ekf/src/odom_estimation_node.cpp: add to print gps sensor and used/not used in getStatus


robot_pose_ekf/CMakeLists.txt
```diff
@@ -66,6 +66,11 @@ install(
     DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
 )
 
+install(
+    DIRECTORY scripts
+    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
```
mikeferguson: this should actually be CATKIN_PACKAGE_BIN_DESTINATION

k-okada: if so, that will goes to lib/robot_pose_ekf/scripts/wtf.py, do you want to install to lib/robot_pose_ekf/wtf.py ?

mikeferguson:
LGTM

PR #413, opened 2016-01-02
move_base: Fix recursive deadlock.

patrickcjh:


DLu:
Can you describe what behavior you've seen and how this fixes it?

patrickcjh:
When switching the global planner using dynamic reconfigure, the program will acquire the _planner_mutex__ twice. When it acquires the mutex for the second time, it freezes indefinitely.

Later I found out that others have similar commits:
1. #375 
2. https://github.com/clearpathrobotics/navigation/commit/07f86292

mikeferguson:
I think https://github.com/ros-planning/navigation/pull/501 is probably a safer fix.

PR #414, opened 2016-01-04
Modified inflation layer so that it does not segfault when being reconfigured

levavakian:
A dynamic reconfigure of inflation radius parameters would sometimes
result in a segfault since the reconfigure callback did not lock the
access mutex before modifying the cached values. This pull request splits the inflation
radius reconfigure off into its own function and adds a mutex.


levavakian:
Matches #416

PR #416, opened 2016-01-05
Fixes the dynamic reconfigure segfault

levavakian:
Doing a dynamic reconfigure of the inflation radius recreates
the cached cost values without first locking a mutex, which causes
a segfault. This breaks the reconfigure of inflation parameters into
a separate function and adds a mutex lock.

Also renames the mutex to avoid future conflicts.


levavakian:
Matches #414

PR #419, opened 2016-02-01
Add optional PNG saving for map_server

mehwang:
Changes:
- Adds optional PNG saving feature enabled by use of rosparam
- Adds rosparam **/map_saver/save_file_type** to determine file type for map_saver
- Add external library dependencies for PNG processing:
  - libpng - C library for processing PNG files
  - png++ - C++ wrapper for libpng
- Reorganized map_saver into map_saver.cpp and map_generator.cpp/.h for ease of testing
- Adds unit test for saving PGM and PNG map files

Things of note that probably need changing:
- The define guards work, but they might not be the right way to do things
- The PNG dependencies are test_depends for a couple of reasons:
  - They aren't necessary for building/running if the functionality isn't desired
  - The tests don't have define guards
- The CMakeLists.txt works, but the formatting might be off
  - libpng is a required package at the moment, which might interfere with the above test_depend assumptions
- Should I switch the rosparam to a namespaced one? I kept it hard coded since it would've required a node handle and none of the other map_server stuff seems to namespace their parameters
- Documentation? Not much code was changed but if more comments are needed I can add them in

I'll also have to update the ROS wiki with the rosparam/functionality changes if they get in.

### DIFF
--- a/map_server/CMakeLists.txt
+++ b/map_server/CMakeLists.txt
@@ -7,14 +7,24 @@ find_package(catkin REQUIRED
             tf
             nav_msgs
         )
+find_package(PNG REQUIRED)
+find_path(png++_INCLUDE_DIR
+  NAMES png++/png.hpp)
+
 
 find_package(Boost REQUIRED COMPONENTS system)
 
 find_package(PkgConfig)
 pkg_check_modules(NEW_YAMLCPP yaml-cpp>=0.5)
 if(NEW_YAMLCPP_FOUND)
-add_definitions(-DHAVE_NEW_YAMLCPP)
+  add_definitions(-DHAVE_NEW_YAMLCPP)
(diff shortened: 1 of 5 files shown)


map_server/CMakeLists.txt
```diff
@@ -60,10 +100,20 @@ if(CATKIN_ENABLE_TESTING)
   catkin_add_gtest(${PROJECT_NAME}_utest test/utest.cpp test/test_constants.cpp)
   target_link_libraries(${PROJECT_NAME}_utest image_loader SDL SDL_image)
 
-  add_executable(rtest test/rtest.cpp test/test_constants.cpp)
+  add_executable(rtest
```
mikeferguson: while we are at it -- we should make this name more specific by prefacing with map_server_

PR #422, opened 2016-02-16
Adding map server raw mode implemented by DLu

aaronhoy:
Originally implemented by DLu (and I've added some minor cleanup), this adds a mode to the map server that simply passes pixel values straight through, allowing more flexibility for how maps are used.

### DIFF
--- a/map_server/src/image_loader.cpp
+++ b/map_server/src/image_loader.cpp
@@ -124,13 +124,13 @@ loadMapFromFile(nav_msgs::GetMap::Response* resp,
 
       if(negate)
         color_avg = 255 - color_avg;
-        
+
       if(mode==RAW){
           value = color_avg;
           resp->map.data[MAP_IDX(resp->map.info.width,i,resp->map.info.height - j - 1)] = value;
           continue;
       }
-        
+
 
       // If negate is true, we consider blacker pixels free, and whiter
       // pixels free.  Otherwise, it's vice versa.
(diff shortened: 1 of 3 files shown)


aaronhoy:
I tested both the new mode and the default behavior (trinary mode) and everything seems to be working as expected.

map_server/src/image_loader.cpp
```diff
@@ -99,7 +99,7 @@ loadMapFromFile(nav_msgs::GetMap::Response* resp,
   rowstride = img->pitch;
   n_channels = img->format->BytesPerPixel;
 
-  if (trinary || n_channels == 1)
+  if (n_channels <= 3)
```
mikeferguson: are we sure on this change?

mikeferguson: I think this should actually be "mode==Trinary || n_channels == 1" -- I'm not even sure why mode==Trinary is there, but let's not change the functionality of trinary mode

aaronhoy:
Correction pushed to address comment from @mikeferguson .
The change now properly checks for the presence of an alpha channel, but still allows trinary mode to override that so as to preserve legacy behavior.

mikeferguson:
@aaronhoy can you open a similar PR against jade-devel

aaronhoy:
@mikeferguson, I opened #428

DLu:
Thanks for following up on this @aaronhoy, and for catching my bugs.

PR #425, opened 2016-02-23
Virtual computeCost function and protected members for inflation layer

AlexReimann:
Currently child classes of the inflation layer do not have access to the resolution, inflation / inscribed radius and the weight (they are private.
So if you want to use a different cost function for the inflation you currently can't simply inherent and change the cost function.

So this changes the members needed for cost calculation to protected and makes the computeCost function virtual so it is possible to override it.

### DIFF
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -111,7 +111,7 @@ public:
   /** @brief  Given a distance, compute a cost.
    * @param  distance The distance from an obstacle in cells
    * @return A cost value for the distance */
-  inline unsigned char computeCost(double distance) const
+  virtual inline unsigned char computeCost(double distance) const
   {
     unsigned char cost = 0;
     if (distance == 0)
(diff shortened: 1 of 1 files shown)


mikeferguson:
I'm pretty sure this breaks ABI on a plugin -- which is very much not desirable. Although I have to admit, I'm not 100% sure as this was inline...

AlexReimann:
Well, it runs with the extension plugin I made so far.
But maybe I just got lucky. The inline keyword is after all just a hint for the compiler and maybe it breaks on different settings.

But as far as I understand from the [gcc documentation (3rd last paragraph](https://gcc.gnu.org/onlinedocs/gcc/Inline.html) it seems like non-static functions (at least in gcc) don't get inlined anyways?:
"When an inline function is not static, then the compiler must assume that there may be calls from other source files; since a global symbol can be defined only once in any program, the function must not be defined in the other source files, so the calls therein cannot be integrated. Therefore, a non-static inline function is always compiled on its own in the usual fashion. "

Remove the inlining? 
I'll try to force an inlining and see what happens >:P

DLu:
Other than the potential ABI breakage, I am fine with this. There was no particular logic between what was protected/private.

AlexReimann:
Checked the ABI breakage: Yes it breaks stuff.
I guess not getting this in.

mikeferguson:
@AlexReimann -- please open against kinetic -- we're still accepting ABI breaks there (since most dependent packages haven't been released yet)

PR #426, opened 2016-02-24
Fixes #424 for indigo: New test and fix for proper interpretation of a delayed initial pose.

### LINKED ISSUE #424: tf lookup timing issues in AmclNode::handleInitialPoseMessage // ("Failed to transform initial pose in time")
In `AmclNode::handleInitialPoseMessage` (see [here](https://github.com/ros-planning/navigation/blob/1f970f25ab1e6689c4c9d841e95ecdee65d0803d/amcl/src/amcl_node.cpp#L1328-L1343)), the tf lookup uses `ros::Time::now()` to get the robot motion since the initial pose's time stamp.
Within this function, `tf2::BufferCore::lookupTransform(base_frame_id_, global_frame_id_, ros::Time::now())` is called. That last call fails often if the odometry update rate is low as it needs to have updated the tf from `global_frame_id_` to `base_frame_id_` since the call to `ros::Time::now()` to succeed.

As a solution I would propose to add a `waitForTransform()` before the lookup call.

stwirth:
See commit messages for more details.

I'll open a PR to jade-devel with this fix if it gets accepted.

### DIFF
--- a/amcl/CMakeLists.txt
+++ b/amcl/CMakeLists.txt
@@ -121,6 +121,7 @@ if(CATKIN_ENABLE_TESTING)
 
   # Tests
   add_rostest(test/set_initial_pose.xml)
+  add_rostest(test/set_initial_pose_delayed.xml)
   add_rostest(test/basic_localization_stage.xml)
   add_rostest(test/small_loop_prf.xml)
   add_rostest(test/small_loop_crazy_driving_prg.xml)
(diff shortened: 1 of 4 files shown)


mikeferguson:
@dbking77 can you take a look at this?

dbking77:
The patch works for me, the previous version does not work correctly

I increased the tf listener to buffer 60 seconds, and set the initial pose after long delays and lots of motion.

mikeferguson:
Thanks! @stwirth can you please open that jade-devel PR and we can merge that too

stwirth:
@mikeferguson PR for jade: https://github.com/ros-planning/navigation/pull/427
@dbking77 thanks for the review!

Issue #432, opened 2016-03-12
Global Costmap is not static anymore (although static is enabled)


ssrselvamraju:
After the upgrade to Navstack 1.12.7 from 1.12.4, we're now experiencing the following issues:
1. In 1.12.4 I didn't have to specify the width and height of the global costmap, and it was automatically picking up to be the size of the actual "map". After the 1.12.7 upgrade seems like I'm having to set the height and width (discovered this, when the global costmap was defaulting to size of 10x10 after the navstack upgrade)
2. Although the height and width is set to 10,000x10,000 the size seems to be shrinking based on the distance the robot is travelling, so if there's a goal position set to a different point in the map which is now not in the global costmap, the robot is unable to plan there.

Global costmap with navstack 1.12.7 (and height X width = 10,000 x 10,000)

![image](https://cloud.githubusercontent.com/assets/3400029/13720086/e8f3aaa4-e7b4-11e5-96c4-fdf76c19d23a.png)

Global costmap with navstack 1.12.4 (and height X width is not explicitly set anywhere, but default params are 10x10. Also publish frequency is 0, since this is a static layer and doesn't have to update)

![image](https://cloud.githubusercontent.com/assets/3400029/13720060/810edeae-e7b4-11e5-9e24-29a0b603bb75.png)

Any suggestions/solutions for me to try?

mirzashah:
I believe you're experiencing same issue I had #430 . I fell back to 1.12.5, the code breaks in 1.12.6

mikeferguson:
Do either of you have a link to your navigation configuration? We're not seeing this issue in house, so I'm hoping to narrow down what you might be using that is different from our config.

mikeferguson:
Also, a trace of the ROSOUT would be useful, in particular, if you see "Requesting the map..." or "Received a map.."

ssrselvamraju:
$ cat costmap_common_params.yaml 

```
obstacle_range: 3.0
raytrace_range: 5.0
robot_radius: 0.5
inflation_radius: 0.5

observation_sources: velodyne_scan laser_scan_sensor laser_scan_sensor2 #laser_scan_sensor3

velodyne_scan: {sensor_frame: velodyne, data_type: LaserScan, topic: velodyne_scan, marking: true, clearing: true}


laser_scan_sensor: {sensor_frame: LMS151, data_type: LaserScan, topic: bottom_scan, marking: true, clearing: true}

[...]

Issue #430, opened 2016-03-04
Nav stack release 1.12.6 breaks repopulation of static layer on costmap clear


mirzashah:
@alexhenning @mikeferguson 

The changes in 1.12.6 created a regression for us. When the costmap is cleared using the clear costmap service exposed by move_base, the static layer does not repopulate with the current map. This resulted in some interesting behavior where the robot wants to drive through walls.

https://github.com/ros-planning/navigation/commit/31accb2e35f0506c3dae08315da1583bb6761403

Reverting back to 1.12.5 appears to fix this.

RaymondKYLiu:
This condition block it.
`if (map_sub_.getTopic() != ros::names::resolve(map_topic))`

corot:
Same problem for me. An easy workaround is to set first_map_only static layer parameter as true. This makes the static layer keep a copy of the first map received and use it on resets. It shutdowns the map subscriber to ensure it only uses this stored copy.

This parameter (and many others) is not documented on the wiki since current layer pages date from hydro! (I opened an issue #439)

alexhenning:
@mirzashah Could you post the full output of `rosparam get /move_base`

mirzashah:
@alexhenning is it ok if the output comes from 1.12.5 or do you need it specifically from 1.12.6?

alexhenning:
@mirzashah I assume you're using the same parameters with both? If the output differs, then that should be unrelated to the source change.

alexhenning:
@mirzashah https://github.com/ros-planning/navigation/pull/446 should fix this bug and be in the next release. Can you verify that it resolves your problem.

mikeferguson:
446 is released into 1.12.8 -- we've not gotten a yes/no from original reporters -- but I'm going to close this. Please reopen if you find that the fix does not work and can provide additional info

corot:
Fixed!   thanks

mirzashah:
@mikeferguson @alexhenning  Thanks so much for the fix --  just tested with 1.12.10, regression seems to be fixed

mikeferguson:
@mirzashah -- awesome -- and apologies on that -- we really did test the crap out of this stuff and somehow never managed to get it into that state.

PR #434, opened 2016-03-24
revert #410, add optional timeout to avoid deadlock

mikeferguson:


mikeferguson:
Addresses #432 #430

@mirzashah @ssrselvamraju I'm unable to figure out why this causes what you're reporting -- I have something else occurring on our robots that is maybe related, but I'm unable to reproduce your exact bug described. I'm going to roll back the change and add an optional param to limit the time we can get stuck if someone is using two static costmap layers

ssrselvamraju:
@mikeferguson I can send you some bag files if you'd like to review.

mirzashah:
@mikeferguson sorry i've been out of the loop and trying to catch up on this -- I'll make a note to try this version out and see if it causes the same issue

ssrselvamraju:
@mikeferguson Just a heads up - I tested this commit on one of our robots, and didn't seem to have fixed the issue. I'm testing more on other machines, with different configurations. Will update details soon.

mikeferguson:
closed in favor of #446

Issue #327, opened 2015-04-02
ObstacleCostFunction has an empty footprint when the first goal is an in-place rotation


ex-ratt:
There seems to be a problem with setting the footprint to the ObstacleCostFunction when using dwa_local_planner. The footprint is set whenever DWAPlanner::findBestPath is called (line 298 in https://github.com/ros-planning/navigation/blob/indigo-devel/dwa_local_planner/src/dwa_planner.cpp) and nowhere else (to the best of my knowledge). In case the goal given to the robot is at its position, so it only has to rotate to match the orientation, DWAPlanner::findBestPath is not called - instead, the LatchedStopRotateController only calls DWAPlanner::checkTrajectory to check for a valid rotation command, which does not set the footprint to the ObstacleCostFunction, though it needs the footprint to check for valid rotation.

If now the very first goal to the robot after starting the move_base node results in such an in-place rotation, then the ObstacleCostFunction cannot check the rotation command and instead prints an error to the log / console (see line 78ff in https://github.com/ros-planning/navigation/blob/indigo-devel/base_local_planner/src/obstacle_cost_function.cpp). If another command was given beforehand, where the robot had to move, then this error does not appear, as the footprint was set. This problem occurred in a question: http://answers.ros.org/question/204740/dwa_local_planner-in-place-rotation-first-goal-does-not-work/?answer=205955#post-id-205955

I wonder if it were possible to set the footprint at construction time. The DWAPlanner knows the costmap and calls the ObstacleCostFunction's constructor with that costmap and the costmap should know the footprint (I'm just not sure if it knows at that moment). So it may be possible to retrieve and set the footprint in the constructor of ObstacleCostFunction. If this is not possible, then it has to be set in DWAPlanner::checkTrajectory similar to DWAPlanner::findBestPath, but that would mean that quite some other objects have to now about the footprint, which to me does not seem that desirable.

Swan-Baigne:
Hello,
i have the same problem, and i wanted to know if any modification was planned, or if i have to rebuild the dwa_planner package myself

ZdenekM:
I'm also getting "Footprint spec is empty, maybe missing call to setFootprint?" right now so I vote +1 for this issue...

askkl:
I have the same problem, too

bradleypowers:
@DLu Any plans to fix this?  We're seeing the same issue as well.  We may be able to take a look at a fix, but bandwidth is limited here atm.

Swan-Baigne:
[...]

PR #435, opened 2016-04-05
Fix issue #327: ObstacleCostFunction has an empty footprint when the …

corot:
…first goal is an in-place rotation.

This is a simplistic fix, without deep study of the problem, so please give it a try before merging.


mikeferguson:
Jade no longer supported -- and #681 is tracking this with a more detailed solution

Issue #437, opened 2016-04-08
[Costmap2DROS] movementCB's hardcoded timer period too short for low-end systems


spmaniato:
@DLu, @mikeferguson, I believe that the hardcoded period (0.1 sec) of `movementCB`'s timer ([costmap_2d_ros.cpp#L160](https://github.com/ros-planning/navigation/blob/jade-devel/costmap_2d/src/costmap_2d_ros.cpp#L160)):

``` cpp
timer_ = private_nh.createTimer(ros::Duration(.1), &Costmap2DROS::movementCB, this);
```

might be too short for very low-end systems, such as many Single-Board Computers. 

**TL;DR**: Would a pull request that makes this timer's period parametrizable make sense? Edit: See #440 😁 

---

The timer's period being so short (for a low-end computer) leads to `tf`-related methods such as `getRobotPose` being called very frequently. In turn, this results in errors and warnings such as:
- `[WARN] Could not get robot pose, cancelling reconfiguration`
- `[ERROR] Extrapolation Error looking up robot pose: [...]` 
- `[WARN] Costmap2DROS transform timeout. [...]`

Many people have been experiencing related issues on real robots with low-end computers, e.g.,
- http://answers.ros.org/question/199631
- http://answers.ros.org/question/215256

I'm not claiming that increasing this timer's period will fix all of these. Other methods call `getRobotPose` too. But it will provide another way to trade-off computational cost for accuracy and vice versa. 

I was thinking of adding a new parameter along the lines of `transform_threshold` and `update_frequency`. For example, `pose_update_period`, with a default of 0.1, or `pose_update_frequency` with a default of 10. Thoughts?

Thanks for your time!

spmaniato:
Addressed in #440, which has been merged. Thanks @DLu 😄

PR #440, opened 2016-04-13
[Costmap2DROS] Parametrize movementCB timer's period

spmaniato:
This change addresses the concern I expressed in #437

The default value of the new parameter, `pose_update_frequency`, results in a timer period of `0.1`, which is the same as the previously hardcoded value.

I didn't think it was important / useful to make the new parameter dynamically reconfigurable. But I can do that too if it's desirable, e.g., for consistency.

Thank you for your time  😃


DLu:
Thanks!

PR #442, opened 2016-04-14
Core 3611 Zero planner patience to avoid loop of death detection failure due to cached paths

chandel01:
In addition to changing planner patience to zero, added >= to the check to ensure after loop of death detection recovery is invoked.


chandel01:
Can't merge myself, I don't have write access rights. @skaynama, @jasonimercer, @p6chen

p6chen:
You are trying to merge this to upstream...

PR #406, opened 2015-11-14
[proposed for Kinetic Kame] navfn: change API

hrnr:
Currently it is not possible to make use of navfn::NavfnROS without costmap_2d::Costmap2DROS. This means that you are forced to publish costmap every time you want to run planner on costmap.

This PR adds another constructor for navfn::NavfnROS to allow running navfn::NavfnROS on arbitrary costmap2d. This makes navfnROS usable in local scenarios, where you don't want to publish costmap, it doesn't make sense to publish costmap, or costmap is published without costmap_2d::Costmap2DROS.

navfn::NavfnROS:
    \* remove direct dependency on costmap_2d::Costmap2DROS
    \* add constructor for barebone costmap_2d::Costmap2D (user must provide also global_frame)
    \* NavfnROS::initialize() follows constructor semantics

nav_core::BaseGlobalPlanner interface unchanged

PR #444, opened 2016-04-26
[proposed for Kinetic Kame] navfn: change API

hrnr:
this unbounds `NavfnROS` from `Costmap2DROS` allowing its use with just `Costmap2D` without all the `Costmap2DROS` machinery.

Reproposal of #406.

### DIFF
--- a/navfn/include/navfn/navfn_ros.h
+++ b/navfn/include/navfn/navfn_ros.h
@@ -70,11 +70,27 @@ namespace navfn {
       NavfnROS(std::string name, costmap_2d::Costmap2DROS* costmap_ros);
 
       /**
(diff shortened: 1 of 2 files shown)


hrnr:
Hi, how is it going with Kinetic? Is this good to go in?

mikeferguson:
@alexhenning -- can you give your thoughts on this (note that it is Kinetic+ only)

alexhenning:
@mikeferguson SRTM

mikeferguson:
Once previous PR merges pass build, I will merge

mikeferguson:
Current mainline is not building due to https://github.com/ros-planning/navigation/issues/456 -- once we get that resolved this will get in

PR #322, opened 2015-03-26
Switch to tf2 for Jade

vrabaud:
Same as https://github.com/ros-planning/navigation/pull/317 but for Jade.
Honestly, this merge request is only useful because it can free you from PCL later. So you might as well drop the indigo version (you won't gain performance or anything else).

PR #458, opened 2016-05-07
Switch to tf2 for Kinetic

vrabaud:
This a cherry-pick from https://github.com/ros-planning/navigation/pull/322/files


vrabaud:
I think you had a problem with that PR at some point, a year or two ago. Tests in Kinetic apparently got tighter and helped uncover a but on my end in AMCL. You should probably give it a new try if you have the time.

vrabaud:
@dbking77 , enjoy

vrabaud:
rebased, tests are passing: it shows as failed because there was a GCC warning (that I cannot narrow down to).

mikeferguson:
Just tried this out on my Kinetic simulation -- and it seems that Fetch cannot turn (except in the recovery behaviors)... going to try and look into it more later

vrabaud:
Thx for not giving up on that. Remember it is the first step to get rid of PCL.
I rebased: tests pass.

mikeferguson:
Yeah, I had rebased on top of head of kinetic for testing, I'm not sure why it doesn't work at runtime -- I do plan to look into it now that I have a 16.04 setup, but ran out of time the other day

corot:
:+1:  for this, but... out of curiosity: given tf1 indigo implementation wraps a tf2 listener for backward compatibility, can we expect any performance improvement with this?
I ask because I'm experiencing weird things with transforms delays (see #517), and I wonder if the tf wrapping of tf2 has something to do with that

vrabaud:
@corot, no idea about performance, I don't use that code.
Now, I just rebased it and it takes an insane amount of time to compile (actually link the binaries in costmap_2d) because PCL now brings in a lot of VTK dependencies. This PR is still the first step to minimize the PCL dependency.

vrabaud:
Sooooo, how do we move with this one? I understand it is scary and things were not working for you.
How about I split it into what is only math, and what also deals with pub/sub ?
Also, do you care that some public API is broken? For which packages? Is it ok for packages nobody depends upon except navigation? (like dwa_local_planner or carrot_planner, http://rosindex.github.io/p/dwa_local_planner/#kinetic-deps)

lucbettaieb:
If this is passing...  Could be nice!

vrabaud:
I can cherry-pick to Lunar, please advise.

mikeferguson:
#755 is the new #458

PR #459, opened 2016-05-09
Pass in action types as references for costmap algorithms.

stonier:

Initial pull request for indigo (redirected here to kinetic to avoid ABI changes) in #396


stonier:
Closing this - it will suffer from the same ABI change problem now. I don't have a need for this any longer, so if anyone new wants to push it along, they can recreate a PR for it.

PR #462, opened 2016-05-23
[amcl_laser] Revert the measurement computation to what is in Probabilistic Robotics textbook.

130s:
Originally raised [in this QandA](http://answers.ros.org/question/30711/amcl-observation-model/).

I'm also wondering how this ad-hoc weight was developed. I understand that this ad-hoc value has been functioning for the purpose, but ideally should better be to be documented with a mathematical rationale considering it's based on a textbook.

I don't expect this to be merged as it is, or even understand if rejected; if the rationale for the current ad-hoc modification is provided, I think that would be fine. And we might want to keep using a long-standing logic rather than modifying it, unless new logic is tested thoroughly or gone through a heavy review process.

### DIFF
--- a/amcl/src/amcl/sensors/amcl_laser.cpp
+++ b/amcl/src/amcl/sensors/amcl_laser.cpp
@@ -197,10 +197,11 @@ double AMCLLaser::BeamModel(AMCLLaserData *data, pf_sample_set_t* set)
 
       assert(pz <= 1.0);
       assert(pz >= 0.0);
-      //      p *= pz;
+      // Instead of using the ad-hoc weight in a few lines later, this is what the textbook defines.
+      p *= pz;
       // here we have an ad-hoc weighting scheme for combining beam probs
       // works well, though...
-      p += pz*pz*pz;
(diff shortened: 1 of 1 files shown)


mikeferguson:
Note that this has been this way since at least 2009 based on blame

lucbettaieb:
@130s Have you tested this change and seen it working effectively?  I tried making the change in some stuff that I've been working on and it resulted in scoring not working correctly.

130s:
@lucbettaieb Not tested exhaustively. As mentioned I created this PR more for raising an attention.

But if you have a `.bag` file that you can share publicly, that'll be great (not just for me. We might be able to integrate that into the system test).

lucbettaieb:
I'm not able to release what I've done, sorry. :disappointed: 
But, I understand the point you've raised and I appreciate it.

mikeferguson:
I'm pretty sure if we were to change this, everyone would get pretty angry. If you really want to get something along these lines in, I would suggest adding a new parameterized approach that can be enabled by choice and opening a new PR against K or L.

PR #463, opened 2016-05-24
Fixed an issue where vtheta_samples parameter was ignored

cwitting:
In the documentation the parameter for the number of vtheta samples is called vtheta_samples, in the config file it was different. This caused the vtheta_samples parameter to have no effect, and it would always result in the default value of 20 samples.
It is probably wrong in the other branches as well.


mikeferguson:
This has been around for a very long time now -- and I would presume that most people have either:

 * Tuned their robot using vth_samples after realizing that it wasn't doing what they wanted, or:
 * Tuned their robot to work properly with the default vth_samples, and might have a completely crazy vtheta_samples value that causes things to break.

The least breaking changing would be to update the documentation to note that the parameter name is actually vth_samples. Unfortunately that will mean the parameter names are not the same for base_local_planner and dwa_local_planner, but there are other params that are not the same, so I think that is not too high a cost to pay for existing configurations to work.

I have updated http://wiki.ros.org/dwa_local_planner -- if you're aware of other docs, please update those. Thanks for noticing this!

PR #464, opened 2016-05-25
Made costmap publishing truly lazy

alexhenning:
This prevents the message from being created if there are no subscribers.

@mikeferguson Can you take a look at it? If you're happy with it, I'll cherry-pick it and make matching jade and kinetic PRs.


mikeferguson:
LGTM

PR #469, opened 2016-05-27
Fixes issue with costmaps shearing

alexhenning:
A regression that appears to have been introduced in https://github.com/ros-planning/navigation/commit/1e01019f837a751f3d52cd8c0989693541e8220e causes the costmap to be sheared.

screenshot - 05272016 - 11 51 01 am

This commit resolves that problem in our testing.


costmap_2d/plugins/inflation_layer.cpp
```diff
@@ -154,10 +154,10 @@ void InflationLayer::updateBounds(double robot_x, double robot_y, double robot_y
     // box by the amount of the cell_inflation_radius_.  Cells
```
mikeferguson: this comment probably needs updating

alexhenning: Addressed this.

PR #479, opened 2016-06-08
remove link to devel job

mikeferguson:


mikeferguson:
@alexhenning -- this did pass, so we must have some regression in your PR

PR #478, opened 2016-06-07
Fixed bug with inflation layer that caused underinflation

alexhenning:
When marking before adding to the priority queue, it was possible to
underestimate the cost of a cell. This is both dangerous and can lead to
unintended side-effects with navigation.

PR #481, opened 2016-06-09
Fixed bug with inflation layer that caused underinflation

alexhenning:
When marking before adding to the priority queue, it was possible to
underestimate the cost of a cell. This is both dangerous and can lead to
unintended side-effects with navigation.
#478 for kinetic


mikeferguson:
Build failure is same macro warning, which I haven't been able to track down yet

Issue #485, opened 2016-06-23
Release 1.12.11 breaks local costmap - It doesn't always / fully update


spmaniato:
On my Trusty/Indigo machine, (Debian package `ros-indigo-navigation` `1.12.8`), the local costmap works as expected.

On my macOS/Indigo laptop (navigation `1.12.11-0.tar.gz` built from source), I'm seeing weird local costmap behavior. In a nutshell, it's not updating properly in the presence of obstacles / laser scans within it's width and height. Please see screenshots below. (The global costmap looked fined, but I didn't pay close attention to be honest.)

~~There's a very slight chance this a macOS ROS installation artifact, although I doubt it~~ See below

Relevant configuration (which is the same on both machines):

``` yaml
global_frame: odom # Problem also appears with map as the global frame

update_frequency:  5.0
publish_frequency: 2.0

static_map: false
rolling_window: true

width: 1.5
height: 1.5
resolution: 0.05

plugins:
  - {name: obstacle_layer,  type: "costmap_2d::ObstacleLayer"}
  - {name: inflation_layer, type: "costmap_2d::InflationLayer"}
```

Please let me know if I should provide more information, any debug logs, etc. In the meantime, I'll try building `1.12.8` from source in macOS and `1.12.11` from source in Ubuntu to see what happens:
- **Ubuntu**
  - `1.12.11` from source: The exact same problem manifests ❌ 
  - `1.12.10` from source: `TODO`
  - `1.12.9` from source: `TODO`
  - `1.12.8` Debian package: Works as expected ✅
- **macOS**
  - `1.12.11` from source: This is where I happened to observe the problem ❌
  - `1.12.10` from source: Works as expected ✅
  - `1.12.9` from source: Works as expected ✅ 
  - `1.12.8` from source: Works as expected ✅
## Screenshots

<img width="1550" alt="screen shot 2016-06-23 at 10 23 15 am" src="https://cloud.githubusercontent.com/assets/2285973/16310397/fda942a0-3931-11e6-996b-2c61da6e3923.png">
<img width="1550" alt="screen shot 2016-06-23 at 10 25 10 am" src="https://cloud.githubusercontent.com/assets/2285973/16310394/fda59538-3931-11e6-8693-0046e836f6ce.png">
<img width="1550" alt="screen shot 2016-06-23 at 10 26 42 am" src="https://cloud.githubusercontent.com/assets/2285973/16310395/fda6949c-3931-11e6-8bc2-d801d1eed73f.png">
[...]

PR #486, opened 2016-06-24
#485 inflation fix

alexhenning:
Fixes a sign error in the inflation layer and adds a warning if there is ever another similar bug in any plugin.

Closes https://github.com/ros-planning/navigation/issues/485


mikeferguson:
@alexhenning -- please open PRs for jade/kinetic as well

PR #495, opened 2016-07-06
Fix #494: prevent zero-velocity commands during recovery behaviors

### LINKED ISSUE #494: cmd_vel flooded with zero-velocity commands during recovery behaviors
This is something new, seems to me that coming with [this commit](https://github.com/ros-planning/navigation/commit/0a686c90c6f188a2731f6f88c2c08610f0ec907e). Happens when the global planner fails to produce a plan: move_base keeps publishing zero-velocity messages. This can hamper recovery behaviors that want to move the robot.
@gjc13, do you think we can remove the publishZeroVelocity(); part from your fix?

corot:
Partially reverts 0a686c90c6f188a2731f6f88c2c08610f0ec907e

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -529,7 +529,6 @@ namespace move_base {
     //if the planner fails or returns a zero length plan, planning failed
     if(!planner_->makePlan(start, goal, plan) || plan.empty()){
       ROS_DEBUG_NAMED("move_base","Failed to find a  plan to point (%.2f, %.2f)", goal.pose.position.x, goal.pose.position.y);
-      publishZeroVelocity();
       return false;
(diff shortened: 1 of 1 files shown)


mikeferguson:
I'm not sure why the CI failed -- I will look into it shortly

corot:
mmm...  change is terribly simple; cannot imagine how can make anything fail!

I bet all tests will pass if run again; could you do so, @mikeferguson, assuming you are not on holidays?

Many thanks!

mikeferguson:
I think this might need to be ported to J/K as well?

corot:
Seems like there's no problem on J/K. I didn't tried the code, but the line I removed on indigo is not there

Issue #496, opened 2016-07-11
Can we make move_base fail faster if the global planner cannot find a path?


corot:
Now move_base keeps calling global planner up to 'planner_patience' seconds regardless how many times the global planner fails. In our particular case, we allocate 2 seconds (rather short time) to the SBPL lattice planner, so we set planner_patience as 2.5 seconds. This is a notable waste of time because the planner has time to fail many many times, while after 4 or five attempts planner can do in some tens of milliseconds, it seems clear he will never succeed.

Afaik, there's no solution for this without patching move_base. Would it make sense to replace planner patience by an alternative planner_max_calls parameter? Or better keep both mechanism for different use cases.

mikeferguson:
removing planner_patience would probably be too drastic of an API change -- but adding an additional "max_planning_attempts" is an acceptable API change as long as the default is "disabled" (probably represented as either 0 or -1)

corot:
Thanks! Also acceptable to release on indigo?

mikeferguson:
Yes -- we'll have to extensively review it (so the shorter the PR, the better), but we should be able to put this into I/J/K as long as we maintain backwards-compatability

corot:
I tried to be quirurgical...

PR #498, opened 2016-07-12
Issue #496 implementation

corot:
Add a max_planning_retries parameter as an alternative to planner_patience to limit the failed calls to global planner.

The new parameter is -1 by default, what means disabled and so the old good behavior based on planner patience.

### DIFF
--- a/move_base/cfg/MoveBase.cfg
+++ b/move_base/cfg/MoveBase.cfg
@@ -2,7 +2,7 @@
 
 PACKAGE = 'move_base'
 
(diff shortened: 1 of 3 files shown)


corot:
@mikeferguson,  looks good for merging?  I would like to have it merged asap, whenever possible

corot:
Sorry for insisting.... @mikeferguson, can we merge this? do you prefer a different approach?
Thanks!

mikeferguson:
LGTM, please open a PR against J/K branches as well

corot:
PR on kinetic

PR #500, opened 2016-07-14
Add custom Rosie odom model

spmaniato:


spmaniato:
Oops, wrong target remote! Sorry!

PR #501, opened 2016-07-19
Fixed deadlock when changing global planner

selvasamuel:
Changing the global planner through dynamic_reconfigure causes a deadlock. The deadlock is caused because of the attempt to lock planner_mutex_ twice during the execution of reconfigureCB() in move_base/src/move_base.cpp. When setting the global planner, planner_mutex_ is first locked on line 266. After this, resetState() is called on line 272. In resetState(), planner_mutex_ is locked again on line 1158 resulting in a deadlock. This fix changes planner_mutex_ to a boost::recursive_mutex.

### DIFF
--- a/move_base/include/move_base/move_base.h
+++ b/move_base/include/move_base/move_base.h
@@ -212,8 +212,8 @@ namespace move_base {
 
       //set up the planner's thread
       bool runPlanner_;
-      boost::mutex planner_mutex_;
-      boost::condition_variable planner_cond_;
+      boost::recursive_mutex planner_mutex_;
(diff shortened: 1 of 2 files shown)


mikeferguson:
trying to retrigger tests

selvasamuel:
Now that the tests are passing, will this be merged with the base branch?

mikeferguson:
I apparently reviewed this ages ago, closed a bunch of competing pull requests, and then never got around to actually merging after CI finished. Apologies on that one.

mikeferguson:
Cherry-picked into K&L

PR #504, opened 2016-07-25
[jade] move_base: Add move_base_msgs to find_package.

de-vri-es:
The `CMakeLists.txt` pf `move_base` was missing `move_base_msgs` in `find_package(catkin REQUIRED COMPONENTS ...)`, causing `catkin_make_isolated` builds to fail.

This PR should fix the problem by adding the `move_base_msgs` to the list of packages. It is already listed in `package.xml`.

### DIFF
--- a/move_base/CMakeLists.txt
+++ b/move_base/CMakeLists.txt
@@ -9,6 +9,7 @@ find_package(catkin REQUIRED
         actionlib
         dynamic_reconfigure
         message_generation
+        move_base_msgs
         nav_core
         tf
 )


de-vri-es:
Note that I based the PR on `jade-devel` because that is the default branch. I can base it on `kinetic-devel` instead if preferred.

mikeferguson:
This looks good, I'll wait for the CI to pass before merging -- if you could also open PRs against indigo-devel and kinetic-devel as well since this probably affects all branches. Thanks!

de-vri-es:
Opened two more PRs for kinetic and indigo :)

PR #509, opened 2016-08-10
Fixed race condition with costmaps

alexhenning:
Modifying minx_, miny_, maxx_, and maxy_ without locking the mutex is
unsafe and is a race condition if updateMap is called from multiple
threads.


mikeferguson:
Please open similar against J/K

PR #513, opened 2016-08-16
Add legacy signature with MapMode as bool

### LINKED ISSUE #474: [map_server] preserve backwards compatibility of map_server::loadMapFromFile
The recent change from `trinary` to `map_mode` in map_server::loadMapFromFile breaks backwards compatibility.
Instead of just changing the signature, a new signature should be provided and the "old" one with the `trinary` boolean should be preserved for backwards compatibility. The users of the old one should be instructed (e.g. by a ROS_WARN) to change their code to use the new interface.

mryellow:
fixes #474


map_server/src/image_loader.cpp
```diff
@@ -54,6 +57,34 @@ namespace map_server
...
+loadMapFromFile(nav_msgs::GetMap::Response* resp,
+                const char* fname, double res, bool negate,
+                double occ_th, double free_th, double* origin,
+                bool trinary)
+{
+  ROS_WARN(
+    "map_server::loadMapFromFile with trinary as bool has been deprecated.",
+    "Please update your code to use MapMode enumerable."
+  );
+  MapMode mode = (trinary)?TRINARY:RAW;
```
stwirth: I think this needs to be SCALE instead of RAW. The RAW option did not exist before the MapMode became an enum.

mikeferguson:
I've commented in the original ticket as to whether this is still relevant a year later, or if all downstream libraries have already updated. If it is relevant, we will need to do some fixups to cmake. Also, if we are going to fix in Kinetic, we should also have a patch in Indigo, so that the first release that no longer supports the classic API would be Lunar (would make no sense to remove in Indigo, re-add in Kinetic, and remove in Lunar)

mikeferguson:
(And BTW, with the exception of the branch conflicts, this LGTM)

DLu:
@mryellow Can you resolve the conflicts?

mryellow:
Don't have a kinetic setup running which I can test on.

This commit stepped on the same area, looks like newer `kinetic-devel` is probably the correct one.
https://github.com/ros-planning/navigation/commit/18356eaa4da59f8d1f25230dbfa27209121d69bf#diff-17895ab0833c9808dd833590ac8871beR32

This other `std::string(fname) + std::string("\": ") + IMG_GetError();` conflict in `image_loader.cpp`, not sure where that comes in. Looks to just be formatting related on https://github.com/ros-planning/navigation/pull/513/commits/ed2f1830b3c72b18bfee0213ea0cd8026d0cf02b#diff-131b6b4d9e11b01840556ffeeb9acfe0R89

mikeferguson:

Issue #517, opened 2016-08-24
Can failure of costmap to meet its update rate make getRobotPose fail?


corot:
Ok, this is something rather mysterious I cannot understand, probably not a bug, but is bugging me quite a lot:

First, I have noticed that on all move_base components robot pose is get by calling Costmap2DROS::getRobotPose. That's fine, but... could somehow this method fail if the costmap is strugling to keep its update rate? I know this don't make sense because the map is updated in a separated thread, and the tf listener used to transform the robot pose into costmap frame has also his own thread.... but the fact is that I have this problem with global costmap / global planner:

```
[/move_base/move_base  WARN 1470136638.285557852]: Map update loop missed its desired rate of 1.0000Hz... the loop actually took 1.573 seconds
[/move_base/move_base  WARN 1470136638.285557852]: Unable to get starting pose of robot, unable to create global plan
     .........many more..................
[/move_base/move_base  WARN 1470136638.285557852]: Unable to get starting pose of robot, unable to create global plan
[/move_base/move_base  WARN 1470136638.285584956]: Unable to get starting pose of robot, unable to create global plan
[/move_base/move_base  WARN 1470136638.285606660]: Unable to get starting pose of robot, unable to create global plan
[/move_base/move_base  WARN 1470136638.285620901]: Unable to get starting pose of robot, unable to create global plan
[/move_base/move_base  WARN 1470136638.285633567]: Unable to get starting pose of robot, unable to create global plan
[/move_base/move_base  WARN 1470136638.496970706]: Could not get robot pose, cancelling pose reconfiguration
[/move_base/move_base  WARN 1470136643.497352621]: Costmap2DROS transform timeout. Current time: 1470136643.4973, global_pose stamp: 1470136643.1959, tolerance: 0.3000
[/move_base/move_base  WARN 1470136643.497717177]: Could not get robot pose, cancelling pose reconfiguration
[/move_base/move_base  WARN 1470136648.997126127]: Costmap2DROS transform timeout. Current time: 1470136648.9971, global_pose stamp: 1470136648.6959, tolerance: 0.3000
[/move_base/move_base  WARN 1470136648.997166981]: Could not get robot pose, cancelling pose reconfiguration
[/move_base/move_base  WARN 1470136651.843847579]: Costmap2DROS transform timeout. Current time: 1470136651.8438, global_pose stamp: 1470136651.5292, tolerance: 0.3000
[...]

Issue #441, opened 2016-04-14
Move locks from move_base into individual planners


AlexReimann:
During the [execute cycle move base locks the local costmap in the controlling state](https://github.com/ros-planning/navigation/blob/ada542feae238cc85eeab9944cc39e37eee93481/move_base/src/move_base.cpp#L879) without actually accessing the costmap.

What's the reasoning behind this?

It seems kind of to be done in the case the local planner wants to do something with the local costmap in the computeVelocityCommands function. This seems to be overkill and maybe restricting the performance.

In our case the computeVelocityCmmands sometimes takes a while, but we don't keep the updating of the local costmap locked during that time.

So.. could it be removed?

AlexReimann:
fyi @stonier

corot:
Same with the [global planner](https://github.com/ros-planning/navigation/blob/ada542feae238cc85eeab9944cc39e37eee93481/move_base/src/move_base.cpp#L450). I think it makes sense to lock the costmaps when about to read them in the planners, so you don't read them right when they are cleaned (I have seen this happens often), but that should be the responsibility of the planners themselves, done in the right moment to minimize the blocking time.

stonier:
+1

mikeferguson:
To a large extent, the default planners all assume a consistent world state throughout a planning cycle. I concur that it probably *should* be in the planner itself, but that is made slightly difficult because while we can update all the default planners, we don't have the ability to update all the planners out there that may be relying on our costmap locking in move_base.

I would suggest that if we want to change this, we only do it in lunar and above if we do it very soon, or we do it in a later release if we cannot fix it ASAP. That limits the impact of the change on operational systems.

mikeferguson:
Adding help wanted to note that maintainers do not currently plan to tackle this, but PRs are welcome!

corot:
FYI, on the new [move base flex](https://github.com/magazino/move_base_flex/issues/4), we plan to drop this locking and advice plugin developers to adapt their code.
As a showcase of why we think this is a good thing, the sbpl_lattice_planner can take an arbitrary time to compute a plan (configurable), while it just need some tens of milliseconds to make his own copy of the global costmap.

mikeferguson:
I've changed the title of this ticket to the actionable fix we want to implement

iluetkeb:
Can I ask for some clarification on the motivation here, please?

[...]

PR #520, opened 2016-09-22
Fix for #517: create a getRobotPose method on move_base

corot:
instead of using that on the costmaps.

Solves (I would say surprisingly) the 

```
"Unable to get starting pose of robot, unable to create global plan"
```

problem.

 Change is very simple and hardly harmful. But I cannot tell what is exactly fixing nor why...

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -85,7 +85,7 @@ namespace move_base {
     //set up the planner's thread
     planner_thread_ = new boost::thread(boost::bind(&MoveBase::planThread, this));
 
-    //for comanding the base
+    //for commanding the base
     vel_pub_ = nh.advertise<geometry_msgs::Twist>("cmd_vel", 1);
     current_goal_pub_ = private_nh.advertise<geometry_msgs::PoseStamped>("current_goal", 0 );
 
@@ -332,7 +332,7 @@ namespace move_base {
     tf::Stamped<tf::Pose> global_pose;
 
     //clear the planner's costmap
-    planner_costmap_ros_->getRobotPose(global_pose);
+    getRobotPose(global_pose, planner_costmap_ros_);
 
     std::vector<geometry_msgs::Point> clear_poly;
     double x = global_pose.getOrigin().x();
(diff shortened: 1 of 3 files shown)


move_base/src/move_base.cpp
```diff
@@ -1181,4 +1181,46 @@ namespace move_base {
       controller_costmap_ros_->stop();
     }
   }
+
+  bool MoveBase::getRobotPose(tf::Stamped<tf::Pose>& global_pose, costmap_2d::Costmap2DROS* costmap)
```
mikeferguson, line 1185: ABI in MoveBase doesn't matter, for Indigo+Kinetic -- let's add the name & transform_tolerance to this function signature, and then hard code them above (not great, but avoids ABI break). For Lunar -- let's go ahead and put this PR just this way (with the Costmap2DROS ABI change)

mikeferguson (changes requested):
I like this PR -- but I don't like the breaking of ABI in the Costmap2dROS class (which will cause all sorts of hell on people with out-of-sync plugins, which is tough to track down). Here's my proposal:

 * I will cherry-pick this as-is into Lunar. There are far fewer users and we've still been fixing build issues over there, so stability should not yet be expected.

 * We update this PR to hard-code name+tolerance as parameters to MoveBase::getRobotPose() since ABI of MoveBase class does not matter.

 * Once we merge this updated version into indigo, I will cherry-pick to Kinetic

move_base/src/move_base.cpp
corot: :+1: 

Btw, I think I understand better now the problem I was solving with this PR: As we look the global costmap before calling the global planner (https://github.com/ros-planning/navigation/blob/ada542feae238cc85eeab9944cc39e37eee93481/move_base/src/move_base.cpp#L450), it stops updating and get outdated, so we get the 'Costmap2DROS transform timeout. Current time: 1470136643.4973, global_pose stamp: 1470136643.1959, tolerance: 0.3000' WARN.
As I mention on #441, I think this costmap locking should be responsibility of the planners, and last for the shorter possible time.

corot: btw @mikeferguson, do you want me to apply the requested changes? Or will you do by yourself?

DLu:
This was merged into lunar in #601 and is a part of melodic and noetic as well.

Issue #523, opened 2016-09-29
Global planner called much more times than expected


corot:
After merging #498, if max_planning_retries is set to 0 or bigger, global planner should be called 1 + max_planning_retries times. If not, it should be call for up to planner_patiente seconds.
But it's not the case; it gets called tens or even hundreds of times!

This is discussed in this [ROS answer](http://answers.ros.org/question/242629/global-planner-behavior-upon-planning-failure/)

ledmonster:
+1

asimay:
+1.
I'm writing another global planner demo now, and find this issue also. Original global planner's makePlan function is called back much more times than we expected, even though we send relative "goal" msg only once.

asimay:
I can test this in these days. will give you feedback later. Thanks!

asimay:
hi, dear @corot , I test the new changes.
And it works well in normal case.
but if the robot stuck in recovery mode, the "makePlan" still works too much time.  

Backgroud: if the goal set too near the inflation layer or in inflation layer, this is made deliberated, the robot will stuck in recovery mode, and oscillate over and over again, and this made the max_retries invalid.

please see screenshot as belows: (green line is the plan path, we can see at least 3 lines here)

![screenshot from 2017-02-09 13_01_30](https://cloud.githubusercontent.com/assets/8703449/22770493/ef98e990-eecb-11e6-8f69-573442b9232d.png)

and short log files as belows:

> [ INFO] [1486616322.730474749, 11.900000000]: Got new plan
[ INFO] [1486616322.763460791, 11.900000000]: Got new plan
[ WARN] [1486616397.623080855, 86.800000000]: DWA planner failed to produce path.
[ INFO] [1486616397.635274320, 86.800000000]: Got new plan
[ WARN] [1486616406.830471817, 96.000000000]: Clearing costmap to unstuck robot (3.000000m).
[ INFO] [1486616406.845336966, 96.000000000]: Got new plan
[ WARN] [1486616416.930792445, 106.100000000]: Rotate recovery behavior started.
[ INFO] [1486616419.399480928, 108.600000000]: Got new plan
[ WARN] [1486616429.511433472, 118.700000000]: Clearing costmap to unstuck robot (1.840000m).
[ INFO] [1486616429.530771139, 118.700000000]: Got new plan
[ WARN] [1486616439.629621073, 128.800000000]: Rotate recovery behavior started.
[...]

PR #524, opened 2016-09-29
Fix for #523: global planner called much more times than expected

corot:
So now planner is called 1 + max_planning_retries in a first try, and then 1 + max_planning_retries after every recovery behavior. Or up to  planner_patiente seconds after every recovery. I think this is the expected operation.

Seems to work fine now, but move_base.cpp code is quite messy, so probably needs at least one independent tester.

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -619,54 +619,59 @@ namespace move_base {
         wait_for_wake = false;
       }
       ros::Time start_time = ros::Time::now();
-
-      //time to plan! get a copy of the goal and unlock the mutex
-      geometry_msgs::PoseStamped temp_goal = planner_goal_;
       lock.unlock();
-      ROS_DEBUG_NAMED("move_base_plan_thread","Planning...");
 
-      //run planner
-      planner_plan_->clear();
-      bool gotPlan = n.ok() && makePlan(temp_goal, *planner_plan_);
+      if(state_==PLANNING){
+        //time to plan! get a copy of the goal and unlock the mutex
+        geometry_msgs::PoseStamped temp_goal = planner_goal_;
+        ROS_DEBUG_NAMED("move_base_plan_thread","Planning...");
 
-      if(gotPlan){
-        ROS_DEBUG_NAMED("move_base_plan_thread","Got Plan with %zu points!", planner_plan_->size());
-        //pointer swap the plans under mutex (the controller will pull from latest_plan_)
-        std::vector<geometry_msgs::PoseStamped>* temp_plan = planner_plan_;
+        //run planner
+        planner_plan_->clear();
+        bool gotPlan = n.ok() && makePlan(temp_goal, *planner_plan_);
 
-        lock.lock();
-        planner_plan_ = latest_plan_;
-        latest_plan_ = temp_plan;
-        last_valid_plan_ = ros::Time::now();
-        planning_retries_ = 0;
-        new_global_plan_ = true;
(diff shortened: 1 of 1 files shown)


ivan-gavran:
@corot , would you mind explaining the fix? From the diff, it seems as some combination with locks, but I don't understand it completely.

I tested the behavior, and it is as described: the move base is being called only 1+max_planning_retries for each recovery behavior.

corot:
Hi @gergia, sorry for not being clear. All the issue is better explained in issue #523 and this [ROS answer](http://answers.ros.org/question/242629/global-planner-behavior-upon-planning-failure/).

The change is quite minimal, once whitespaces are ignored: Without the change, note that there's nothing preventing to keep [calling the planner](https://github.com/ros-planning/navigation/pull/524/files#diff-d6afbf7b921c7c246a6ee2454dc05be4L630) even if we are not in PLANNING state. The change makes that planning thread only calls the planner while current state is PLANNING, up to max_planning_retries+1 times or during the time allowed by planner_patiente. If not, it is blocked in the wait_for_wake planner_cond_.wait(lock); (saving CPU)

Hope is clearer now

corot:
Hi all, anything preventing this PR to get merged?

corot:
PR also on kinetic

move_base/src/move_base.cpp
```diff
@@ -619,54 +619,59 @@ namespace move_base {
         wait_for_wake = false;
       }
       ros::Time start_time = ros::Time::now();
-
-      //time to plan! get a copy of the goal and unlock the mutex
-      geometry_msgs::PoseStamped temp_goal = planner_goal_;
       lock.unlock();
```
mikeferguson: This lock.unlock() call should really move with the temp_goal declaration. We should get the goal WHILE under lock. Of course, that means we need to add an "else{ lock.unlock() }" to the "if (state_==PLANNING)" below, which isn't ideal (it's convoluted) but it is at least thread-safe

move_base/src/move_base.cpp
```diff
@@ -619,54 +619,59 @@ namespace move_base {
...
+            state_ = CLEARING;
+            wait_for_wake = true;
+            publishZeroVelocity();
+            recovery_trigger_ = PLANNING_R;
+          }
+
+          lock.unlock();
         }
-        lock.unlock();
       }
```
mikeferguson: This is where we would add "else { lock.unlock() }".

I did test this lock change, and it looks all good (also tested the whole feature in general).

mikeferguson (changes requested):
@corot -- apologies for the slow responses on these, we should really keep the same locking structure as before, which requires a pair of simple changes. Would be good to also apply these same changes to #539.

mikeferguson:
Thanks! LGTM -- I'll wait for CI to complete again, and then squash-merge for ease of cherry-picking into K+L.

mikeferguson (approved):

DLu:
I think this PR breaks the ability to plan with positive `planner_frequency`, since it forces the planner to only run when `state=PLANNING`, but the intended behavior of move_base is to be able to replan continuously, even when in `state=CONTROLLING`. Was this tested with positive `planner_frequency`?

corot:
I think you are right :see_no_evil: 

Will try to fix it asap; sorry for the gaffe

corot:
PR #622
Ok, hope this time all works as intended!  @DLu, can you give a try to the PR? Thanks & thanks for detecting my pitfall!

Issue #468, opened 2016-05-26
Map update &  Control loop loop missed its desired rate.


asimay:
with new code and old code, it occurs many loop warning like below....
please check the screenshot below, thanks!

![2222](https://cloud.githubusercontent.com/assets/8703449/15561502/1d9facda-2329-11e6-8dcd-41f3339b4d27.png)

And after check code I think I find out a temporary solution, maybe it is also the root cause:
in costmap_2d_ros.cpp file:
ROS_WARN("Map update loop missed its desired rate of %.4fHz... the loop actually took %.4f seconds", frequency,
at line 409,
r.sleep();
if (r.cycleTime() > ros::Duration(1 / frequency)) , the loop time must have some issues, maybe not correct. 

I test the map loop time in function, most of them is actually very small, and can satisfy our needs.(based on small map 10*10m, not test large map.)
if we comment the r.sleep();
the output is very good, and almost no loop warning output. so the cycleTime's calculation must be not correct.
if no wrong, why we comment the sleep(), the function act good and loop warning disapper?

and with the same method, we can remove the controller loop warning in the same way.
it acts good finally.

could you please check above bug?  thanks!

alexhenning:
If you look at how [`sleep`](http://docs.ros.org/api/rostime/html/rate_8cpp_source.html#l00056) and [`cycleTime`](http://docs.ros.org/api/rostime/html/rate_8cpp_source.html#l00098) are implemented, you'll see that the cycle time only gets updated when sleep is called. Not calling sleep may avoid the prints, but it is unlikely that it is actually helping your performance.

mikeferguson:
It appears you are in simulation -- is the parameter "use_sim_time" actually set to true?

asimay:
Dear @alexhenning ,yes, i look through the sleep prototype of ros. but it indeed help me on not print the loop warning info if i comment the sleep code.  so from the code logic, we can verify that the cycle time is not exceed the frequency time, so the loop warning info not be printed out.

Dear @mikeferguson , the parameter "use_sim_time" is set to true, all the costmap's parameter is tuned well according to tutorials. and from #432 we also see the author also has same loop warning info printed out.
and also please these old info:
http://answers.ros.org/question/194980/someone-has-this-error-map-update-loop-missed-its-desired-rate-of-50hz/
http://answers.ros.org/question/195523/warning-map-update-loop-missed-its-desired-rate-keeps-on-comming/

please see below screenshot:

[...]

PR #525, opened 2016-10-05
Speedup (~60%) inflation layer update

corot:
Following [this discussion](https://github.com/ros-planning/navigation/issues/468), the pop in the priority_queue used for sorting the cells to inflate takes a lot of CPU. I replaced it by a map containing cells to inflate split by distance from the closest obstacle. 

I have played around with the changes for a while, and it seems to work fine, also scaling well to higher resolutions and inflation radius. After profiling my changes, the new biggest CPU eater is the map lookup for distances on enqueue method. Maybe someone with better understanding of STL containers can give an extra speedup, knowing this.

I require C++11. Hope it's not an issue for indigo code.

### DIFF
--- a/costmap_2d/CMakeLists.txt
+++ b/costmap_2d/CMakeLists.txt
@@ -1,9 +1,6 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(costmap_2d)
 
-# Required to use is_sorted and all_of algorithms
-add_definitions("-std=c++11")
-
 find_package(catkin REQUIRED
         COMPONENTS
             cmake_modules
--- a/costmap_2d/CMakeLists.txt
+++ b/costmap_2d/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(costmap_2d)
 
+# Required to use is_sorted and all_of algorithms
+add_definitions("-std=c++11")
+
 find_package(catkin REQUIRED
         COMPONENTS
             cmake_modules
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -44,7 +44,6 @@
 #include <costmap_2d/InflationPluginConfig.h>
 #include <dynamic_reconfigure/server.h>
 #include <boost/thread.hpp>
-#include <queue>
 
 namespace costmap_2d
 {
@@ -57,7 +56,6 @@ class CellData
 public:
   /**
    * @brief  Constructor for a CellData objects
-   * @param  d The distance to the nearest obstacle, used for ordering in the priority queue
    * @param  i The index of the cell in the cost map
    * @param  x The x coordinate of the cell in the cost map
    * @param  y The y coordinate of the cell in the cost map
@@ -65,24 +63,14 @@ public:
    * @param  sy The y coordinate of the closest obstacle cell in the costmap
    * @return
    */
-  CellData(double d, double i, unsigned int x, unsigned int y, unsigned int sx, unsigned int sy) :
-      distance_(d), index_(i), x_(x), y_(y), src_x_(sx), src_y_(sy)
+  CellData(double i, unsigned int x, unsigned int y, unsigned int sx, unsigned int sy) :
+      index_(i), x_(x), y_(y), src_x_(sx), src_y_(sy)
   {
   }
-  double distance_;
   unsigned int index_;
   unsigned int x_, y_;
   unsigned int src_x_, src_y_;
 };
-
-/**
- * @brief Provide an ordering between CellData objects in the priority queue
- * @return We want the lowest distance to have the highest priority... so this returns true if a has higher priority than b
- */
-inline bool operator<(const CellData &a, const CellData &b)
-{
-  return a.distance_ > b.distance_;
-}
 
 class InflationLayer : public Layer
 {
@@ -185,7 +173,7 @@ private:
   double inflation_radius_, inscribed_radius_, weight_;
   unsigned int cell_inflation_radius_;
   unsigned int cached_cell_inflation_radius_;
-  std::priority_queue<CellData> inflation_queue_;
+  std::map<double, std::vector<CellData>> inflation_cells_;
 
   double resolution_;
 
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -173,7 +173,7 @@ private:
   double inflation_radius_, inscribed_radius_, weight_;
   unsigned int cell_inflation_radius_;
   unsigned int cached_cell_inflation_radius_;
-  std::map<double, std::vector<CellData>> inflation_cells_;
+  std::map<double, std::vector<CellData> > inflation_cells_;
 
   double resolution_;
 
--- a/costmap_2d/plugins/inflation_layer.cpp
+++ b/costmap_2d/plugins/inflation_layer.cpp
@@ -232,10 +232,12 @@ void InflationLayer::updateCosts(costmap_2d::Costmap2D& master_grid, int min_i, 
   std::map<double, std::vector<CellData> >::iterator bin;
   for (bin = inflation_cells_.begin(); bin != inflation_cells_.end(); ++bin)
   {
-    for (std::vector<CellData>::iterator cell = bin->second.begin(); cell != bin->second.end(); ++cell)
+    for (int i = 0; i < bin->second.size(); ++i)
     {
       // process all cells at distance dist_bin.first
-      unsigned int index = cell->index_;
+      const CellData& cell = bin->second[i];
+
+      unsigned int index = cell.index_;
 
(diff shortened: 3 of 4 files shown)


corot:
Any problem on merging this? I think I fixed the failed check in the second commit (doesn't a new commit re-trigger the testing?
Actually, this PR is quite valuable (imho), so I think many people will appreciate to get it merged asap

procopiostein:
it should re-trigger testing, I believe. Looking at the compilation output I found this error, which should be the cause of the failure:
 /tmp/catkin_workspace/install_isolated/include/costmap_2d/inflation_layer.h:176:40: error: ‘>>’ should be ‘> >’ within a nested template argument list
08:59:11    std::map<double, std::vector<CellData>> inflation_cells_;

corot:
Thanks @procopiostein. I added c++11 to the CMakeLists.txt, so using '>>' on templates should be fine. Maybe this header file used somewhere else where it is compiled with c++98?
I use catkin build, so I build already isolated without any problem.

Anyway, I will just modify the PR by now so it gets merged.

jasonimercer:
Confirmed speedup with VTune. Very nice approach. 

inflation_vtune_markup_sanitized

corot:
Thanks @jasonimercer, nice analysis. 
Btw, is the C++11 requirement preventing this to get merged? I think I can get rid of it easily. @AlexReimann  pointed in issue #468 that indigo public packages must use C++03

mikeferguson:
C++11 is definitely going to hold up merging this -- if you can get rid of that, I can prioritize testing and merging

AlexReimann:
Also [breaks the ABI compatibility probably](http://www.ros.org/reps/rep-0009.html#what-s-not-allowed). I don't think you can get around that. :/

corot:
@mikeferguson, back to C++03

AlexReimann:
That last commit f8d5aa4 is a bugfix for a bug which is also present in c++11 version?

jasonimercer:
Yes. Confirmed with a little test program:

```
#include <vector>
#include <iostream>

int main()
{
  std::vector<int> v;

  v.push_back(0);
  v.push_back(1);
  v.push_back(2);

  for (std::vector<int>::iterator it = v.begin(); it != v.end(); it++)
  {
    if (*it == 1)
    {
      v.push_back(5);
      v.push_back(5);
      v.push_back(5);
    }

    std::cout << *it << std::endl;
  }

  return 0;
}
```

jasonimercer:
and for the C++11 version of the above snippet:

```
#include <vector>
#include <iostream>

int main()
{
  std::vector<int> v;

  v.push_back(0);
  v.push_back(1);
  v.push_back(2);

  for (auto& i: v)
  {
    if (i == 1)
    {
      v.push_back(5);
      v.push_back(5);
      v.push_back(5);
    }

    std::cout << i << std::endl;
  }

  return 0;
}
```

it again gives unexpected (but different unexpected) results. 

The differences seem to be related to how the termination condition is hoisted.

corot:
@jasonimercer, thanks for the analysis. Interestingly enough, the code using both iterators and C++11 vector traversing worked, but using iterators makes test crash, hence the second commit. Till now I thought that inserting in a vector doesn't invalidate iterators, but clearly that's not the case.
However, using iterators to traverse the map seems to be fine, even if adding new keys.

jasonimercer:
The above snippets actually work if you only push back a single value as that does not trigger a resize (updated size still fits in vactor capacity). Valgrind doesn't even complain.

When a resize happens the iterator is now stomping through memory no longer related to the vector. If I had a .reserve(big_number) then the iterator version works as intended (until you exceed big_number). The C++11 foreach syntactic sugar still hoists the termination condition and outputs 0 1 2 rather than 0 1 2 5 5 5.

Either way, good catch on the bug fix.

corot:
Something else blocking merge?

mikeferguson:
@corot -- I'm currently out of the office, but I will look into testing this early next week.

AravindaDP:
any updates on status of this PR?

corot:
We have been using costmaps with this change for several months without problem, and the speedup is really welcomed, so I would encourage merging this PR asap

jasonimercer:
+1
This is solid code now with > 1M hours run time on our robots.

DLu:
This looks good to me, although I'm not sure whether the CellData changes break ABI, and/or whether we should care. :+1:

PPokorski:
I'd have another suggestion. [Here](https://github.com/ros-planning/navigation/pull/525/files#diff-6c22d6c267e1d4b67c00b5ec6fa00902R225) you could, instead of using push_back (which is a really costly operation) iterate beforehand to determine the size of the vector and then use reserve() or resize(), like:

```
int number = 0;
for(int j = min_j; j < max_j ; j++) 
   for(int i = min_i; i < max_i; i ++)
       number++;

obs_bin.resize(number)
```

Btw. How about merging this, anyway?

jasonimercer:
Iterating over the memory twice also has a price and adds more lines of code for someone new to reason about. This can be sped up down the road when C++11 is allowed and we can use move semantics. I'd suggest either merging this PR as-is or closing it. It's high quality, free work by the community given to ROS. ROS should not string contributors along for 9 months.

procopiostein:
@mikeferguson , is there anything stopping this merge? Is there anything we can do to help?

mikeferguson:
@procopiostein I really just need to sit down and test it. Should have time to do that over the next week, trying to do a big scrub of open issues/PRs (I know they've been piling up for a while)

At the end of the day, we're not trying to "string contributors along" -- but we do have day jobs that don't always allow a bunch of time for testing/development. While I'd love to merge lots of new features/improvements quickly, we also have to weigh the fact that when something in navigation breaks (particularly costmaps), real world damage can occur (robots hit stuff). This particular PR drastically changes a core function, and is targeted at an LTS release that was already 2+ years into that 5 year life cycle when the PR was created.

procopiostein:
@mikeferguson thanks for the feedback. I do understand that this is a very delicate PR and it takes a lot of your free time to check it properly. Sorry if it felt I was pressing you in any way, that was not my intention.

mikeferguson:
@procopiostein -- I should have been more clear there -- my first paragraph was responding to your question, but my second paragraph was more general towards the set of the comments that came before yours.

I also didn't specifically answer your question though: one thing you can do to help this PR along would be if you have tested this specific PR on a particular robot platform (real or simulated) please let us know by posting a comment here. It would help to know which robot and which navigation configuration package (if publicly available, please post a link to it). There are many robots running the navigation stack these days, and far more parameters than we can possibly test all combinations, so knowing that something has been tested on a wide set of platforms helps.

corot:
Thanks for the explanation, @mikeferguson. I really understand how busy core maintainers are.

If this helps, I can PR this on kinetic and/or lunar, so we don't risk our venerable indigo. In any case, we have been using the modified costmaps for almost a year without problems, and in fact got rid off issue #584 that used to happen once in a while before.

mikeferguson:
@corot -- there are still a ton of robots running Indigo -- and it sounds like we do have a bunch of testing going on around this, so I think the community would really like it to go into indigo. We've let it sit here for 9 months -- let me dig into it closer over the next couple days.

PPokorski:
@mikeferguson - As a general remark, I understand how hard it is to find time for maintenance like this, but maybe you could use some help 'hiring' more people for that purpose? It would be a great help for the whole community.

procopiostein:
I did some testing with a real PMB2 robot, running ros indigo and ubuntu 14.04, comparing **corot:faster_costmap** branch against version **1.12.13**.
Everything looks good. 
Peaking at the cpu usage, the big difference is while the robot is not navigating, with the cpu at about  40% for corot's branch compared to 120% in 1.12.13.
While moving there was no clear difference in CPU usage.

jasonimercer:
@mikeferguson, sorry I spoke harshly. I know you guys are also contributing your time to manage these repos and keeping the ROS stack stable. I really appreciate all the work you put into it. I just need to recalibrate for upstream and realise that a 2 day old PR isn't "getting old" : )

mikeferguson:
Ok, tested in several hardware configurations here, code looks good. Going to squash-merge for ease of forward port to K+L

HappySamuel:
Hi all

Has this speedup on inflation layer update implemented in ROS Noetic navigation stack deb?

Best,
Samuel

wanghuohuo0716:
> Hi all
> 
> Has this speedup on inflation layer update implemented in ROS Noetic navigation stack deb?
> 
> Best, Samuel

hello samuel， I have the same question in ros melodic version!!!

anyone can reply?

HappySamuel:
Hi @wanghuohuo0716 

Currently these features only available on ROS2.

Best,
Samuel

corot:
> Hi all
> 
> Has this speedup on inflation layer update implemented in ROS Noetic navigation stack deb?

Th initial 60% yes, both in melodic and noetic. The additional 25% only in ROS2, as @HappySamuel says
But you are compiling ROS nav from source, try simple to rebase the branch into yours, or cherry-pick the commits

wanghuohuo0716:
ok, thx!!! I will try it~

PR #528, opened 2016-10-19
renamed targets for message generation (gencpp -> generate_messages_c…

StephanOpfer:
…pp) in order to avoid warnings for non-existing target dependencies


StephanOpfer:
Hm, what does it mean that the checks failed? I think this pull request fixes some warning, but not all.

mikeferguson:
Disregard the failed checks -- they are actually CMake warnings

PR #529, opened 2016-10-23
Fix CMake warnings

vrabaud:
Thix is a follow-up on https://github.com/ros-planning/navigation/pull/528/files that will help me fix https://github.com/ros-planning/navigation/pull/458 on the farm.

### DIFF
--- a/base_local_planner/CMakeLists.txt
+++ b/base_local_planner/CMakeLists.txt
@@ -21,15 +21,15 @@ find_package(Boost REQUIRED
         thread
         )
 
-find_package(Eigen REQUIRED)
+find_package(Eigen3 REQUIRED)
 find_package(PCL REQUIRED)
 include_directories(
     include
     ${catkin_INCLUDE_DIRS}
-    ${EIGEN_INCLUDE_DIRS}
+    ${EIGEN3_INCLUDE_DIRS}
     ${PCL_INCLUDE_DIRS}
 )
-add_definitions(${EIGEN_DEFINITIONS})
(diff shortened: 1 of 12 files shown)


vrabaud:
@dirk-thomas , @mikeferguson , any idea on what is causing the GCC warning? 
http://build.ros.org/job/Kpr__navigation__ubuntu_xenial_amd64/29/warnings18Result/
I could not reproduce locally but from what I could guess, it's something like

```
gcc -D"FOO;"
```

and the ";" should not be there. But that's on the PR script side and I don't know where that is. Thx.
(maybe we could run it once in verbose mode ?)

dirk-thomas:
You can set `CMAKE_VERBOSE_MAKEFILE` to `TRUE` in your branch to make the PR job run in verbose mode.

vrabaud:
@dirk-thomas thx.
And all is fixed, we probably want this in asap to get other PR green too.

base_local_planner/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ find_package(Boost REQUIRED
 
 find_package(Eigen3 REQUIRED)
 find_package(PCL REQUIRED)
+remove_definitions(-DDISABLE_LIBUSB-1.0)
```
mikeferguson, line 26: is this coming in from PCL, or something else?

mikeferguson:
Thanks @vrabaud!

vrabaud:
yes, PCL brings that in but it's not present in their headers that are installed by default.

PR #531, opened 2016-10-24
Move some packages to tf2

vrabaud:
This is the first split of https://github.com/ros-planning/navigation/pull/458

### DIFF
--- a/base_local_planner/CMakeLists.txt
+++ b/base_local_planner/CMakeLists.txt
@@ -18,6 +18,7 @@ find_package(catkin REQUIRED
             rospy
             std_msgs
             tf
+            tf2_geometry_msgs
             voxel_grid
         )
 
--- a/base_local_planner/package.xml
+++ b/base_local_planner/package.xml
@@ -33,6 +33,7 @@
     <build_depend>eigen</build_depend>
     <build_depend>dynamic_reconfigure</build_depend>
     <build_depend>message_generation</build_depend>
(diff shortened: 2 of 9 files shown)


vrabaud:
@wjwwood , please review, this is a first to move to tf2. It uses getYaw from tf2 instead of tf1.

mikeferguson:
@vrabaud -- I'm adding this to my plate for review sometime tomorrow.

One note: we merged some updates for Lunar to use package.xml v2. So, we'll need to do an actual forward port on this (cherry-pick will have a decent number of conflicts)

vrabaud:
Thx ! Would you rather have me make the pull requests on Lunar ? or both ? It all depends on your plans for the different branches (which ones you consider as maintenance and which one you'd rather switch to tf2 one day).

This PR and a few other ones I'll make from #458 do not impact the API and are just based on tf2 math so they should be rather safe (I still believe #458 is not working for you because of some prefix handling behavior change).

Thx for having a look !

mikeferguson:
@vrabaud these all look correct -- I need to still do a runtime test to be 100% sure.

I think we can merge this part in Kinetic, we'll just have to create a slightly different PR for Lunar.

moriarty:
I cherry-picked this onto melodic, fixed a few conflicts and tested it on a robot. PR #752

oscar-lima:
I confirm it works for us in real robot and simulation

PR #533, opened 2016-10-26
Set frame_id and stamp on Path message even if path is not found.

ledmonster:
When path planning is failed, global planner publishes nav_msgs::Path without frame_id and stamp now. It causes error on rviz. Fix it to publish nav_msgs::Path with frame_id and stamp even if path planning was failed.

Before:

```
header: 
  seq: 24
  stamp: 
    secs: 0
    nsecs:         0
  frame_id: ''
poses: []
```

After:

```
header: 
  seq: 122
  stamp: 
    secs: 1477468563
    nsecs: 495041114
  frame_id: map
poses: []

```

### DIFF
--- a/global_planner/src/planner_core.cpp
+++ b/global_planner/src/planner_core.cpp
@@ -337,10 +337,8 @@ void GlobalPlanner::publishPlan(const std::vector<geometry_msgs::PoseStamped>& p
     nav_msgs::Path gui_path;
     gui_path.poses.resize(path.size());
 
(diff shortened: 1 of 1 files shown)


mikeferguson:
I can't see any downside to this, but @DLu is there a reason that this initially had that if() block?

DLu:
If there was, I don't remember it anymore. 
+1

ledmonster:
Thank you!

PR #534, opened 2016-10-31
Inflation Layer protected members and virtual computeCost

AlexReimann:
Brought over from indigo ([original PR](https://github.com/ros-planning/navigation/pull/425)):  

> Currently child classes of the inflation layer do not have access to the resolution, inflation / inscribed radius and the weight (they are private.
> So if you want to use a different cost function for the inflation you currently can't simply inherent and change the cost function.
> 
> So this changes the members needed for cost calculation to protected and makes the computeCost function virtual so it is possible to override it.

This is ABI breaking.


mikeferguson:
Since this ABI breaking, it really shouldn't go into Kinetic (especially now, but even back in October). I've retargeted this at Lunar so we finally get it in (I know this one has been brought up a few times). #602

PR #535, opened 2016-11-01
Make BaseGlobalPlanner::makePlan overload return. [Kinetic]

tappan-at-git:
The [BaseGlobalPlanner::makePlan](https://github.com/locusrobotics/navigation/blob/kinetic-devel/nav_core/include/nav_core/base_global_planner.h#L68) overload which returns calculated cost does not actually return anything by default, causing compiler warnings and potential undefined behavior.

PR for Jade: #536

### DIFF
--- a/nav_core/include/nav_core/base_global_planner.h
+++ b/nav_core/include/nav_core/base_global_planner.h
@@ -70,7 +70,7 @@ namespace nav_core {
                             double& cost)
       {
         cost = 0;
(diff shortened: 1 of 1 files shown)


mikeferguson:
Great catch -- thanks for the patch!

tappan-at-git:
No problem! Credit for the catch should go to @ayrton04 for noticing the bug in the first place.

PR #539, opened 2016-11-09
Add a max_planning_retries parameter to move_base [kinetic]

corot:
It's an alternative to planner_patience to limit the failed calls to global planner, much better suited for planners that take long to compute a plan (and so you cannot set a low patiente) but short to fail, e.g. sbpl lattice planner.

This PR combines #498 (merged on indigo) and #523 (still open) for Kinetic. I included #523 because it fixes move_base in a way that makes #498 fully effective (explanation in the open PR)

### DIFF
--- a/move_base/cfg/MoveBase.cfg
+++ b/move_base/cfg/MoveBase.cfg
@@ -2,7 +2,7 @@
 
 PACKAGE = 'move_base'
 
-from dynamic_reconfigure.parameter_generator_catkin import ParameterGenerator, str_t, double_t, bool_t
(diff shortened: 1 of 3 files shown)


ledmonster:
+1

mikeferguson (changes requested):
Requesting same lock fixes as #524

mikeferguson:
Cherry-picked to lunar (https://github.com/ros-planning/navigation/commit/7e6fa602017bbee9b756e26b4097a4ea3ab75eda)

PR #543, opened 2016-11-17
Set default name for uncomplete path

Jntzko:
In the case we run:
`rosrun map_server map_saver -f /tmp/`
The map will be created as '.pgm' and '.yaml'. With this fix the map
will be created as 'map.pgm' and 'map.yaml' as in the case no -f parameter is set.

### DIFF
--- a/map_server/src/map_saver.cpp
+++ b/map_server/src/map_saver.cpp
@@ -153,6 +153,9 @@ int main(int argc, char** argv)
       return 1;
     }
   }
+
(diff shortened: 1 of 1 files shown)


DLu:
LGTM. @mikeferguson Do you know why the checks haven't completed yet?

mikeferguson:
@DLu -- not sure why it didn't work -- I've closed/reopened which seems to have retriggered

PR #545, opened 2016-11-25
Voxel clearing corners

phil0stine:
This could just as well apply to `kinetic-devel` but we have only tested with `indigo` at the moment.

This is the distillation of #267 to a pull request that has been tested on a robot with 3 sensors. Much of the initial development was done by @AlexReimann, with design refinements and bugfixes added. 

Many of the changes (ie, `AbstractGridUpdater`) were written to support the "simple" default clearer, which should be _exactly_ the same as the original (current) implementation. Now, using a parameter flag (`clear_corner_cases`), the new algorithm can be applied instead, which creates a volumetric cache around each clearing sensor to apply to the costmap.

The cache is used to speed up the implementation, and should have larger impact with more points to clear. We profiled the initial implementation, shown [here](https://github.com/ros-planning/navigation/pull/316#issuecomment-123719787), but have not profiled the latest code. However, we have used it in practice on a live robot and there are no qualitative speed issues.

One overall issue with such a commit such as this is that there is currently no simple way to perform unit/integration tests to establish its correctness (this applies to many components of the nav stack). One suggestion would agree upon a test bagfile, as well as test launch and parameter files, so that all components of the execution could be properly controlled. This seems like a larger issue that would need more work/consensus, but I think it would be worth the effort.

I know this is a large change but I encourage testing and consideration into the main branches.


safijari:
Heya,

During some debugging of our robot that uses the Intel Realsense rgbd sensor I ran into an issue that I "think" is related to this PR. Our system seeminglyrefuses to clear a cell in the voxel costmap if the marked voxel is neighboring a cell in the "unknown" space. In the screenshot below the blue voxels are this "unknown area" and the red voxels are marked as obstacles. These were marked as I was entering or leaving the robot's field of view. Reading the attached issue it sounds like this is the so called "corner case" but I wanna confirm. Do you think my problem could be fixed with this PR? Thanks :)

image

AlexReimann:
Yes, sounds exactly like what this should be solving.

safijari:
Okay cool. We will test this out some time in the next week and post back :)

phil0stine:
Just a heads up @safijari that this PR should work in practice but is currently failing a test due to some refactoring that wasn't retroactively applied to the test, so only use this as a temporary fix. I will update the PR so that it passes tests shortly.

davidbsp:
@phil0stine, this never made it to the official release? I am facing the "clear corner case" again, with an Orbbec Astra this time (I am also using an RPLIDAR laser as a second observation source without issues).

I tested your commit c83b69e and noticed a clear computation performance drop. My local map cannot keep the update rate at 5Hz, and my control loop also fails to send cmd_vels to the base at 10Hz. As expected, the clearance issues and ghost pixels seem to go away.

phil0stine:
@davidbsp I have not updated my original PR yet, but I would be interested to hear more about the details of your setup. We are using reasonably powerful computing so might not have noticed as great a drop, but we are using a comparable sensor set.

phil0stine:
I have recently been back at this and have (finally) produced an update that passes all tests and has been verified on a real system, though not with rigorous experiments. I also included new tests to verify that the corner clearing works and does not break existing functionality if turned off, which is the default.

Also, @davidbsp is correct about a performance hit, which led us to downsample our dense point cloud sensor with a VoxelGrid nodelet (to 0.05m voxels), and to also limit the clearing range.

AlexReimann:
Try increasing the `accuracy_multiplier_bits` via dynamic reconfigure or try the `clear_corner_cases` option.
There are still limits to the the max accuracy though.

Maybe also take a look at the `Use of the Spatio-Temporal Voxel Layer` talk from [RosCon2018](https://roscon.ros.org/2018/#program) as an alternative

mikeferguson:
Targets a dead release (indigo). If this is still of interest, please open a PR to noetic-devel branch

PR #546, opened 2016-12-01
Fix to increment 'cycle' in while loop

ledmonster:
'cycle' should be incremented in while loop.


mikeferguson:
@DLu could you look this one over -- you're more familiar with global_planner than I am

ledmonster:
Thank you !

PR #547, opened 2016-12-01
Fix to check index value before accessing to element of potential array

ledmonster:
`next_i` can be out of range, and it may cause segmentation fault.


mikeferguson:
@DLu could you look this one over -- you're more familiar with global_planner than I am

PR #560, opened 2017-03-03
fix cmake use of yaml-cpp and sdl / sdl-image in map_server

wjwwood:
While hacking on a port of this for ROS 2 (for a demo) I found that the cmake, especially with respect to finding and using dependencies, was more fragile than it has to be. So I open this pr, which:

- removes `find_package()` of boost (it's not used directly)
- changes the logic to find yaml-cpp and actually use the result
- change the compiler definition which indicates yaml-cpp >= 5.0.0 to be more self describing
- actually does `find_package()` the existing dep `SDL-image` and the implicit dep on `SDL`
  - It did not compile for me on `macOS` because `SDL-image` headers assume they are installed to the same folder as `SDL` headers, which is not the case with Homebrew.
- adds a `package.xml` dependency on `SDL` (https://github.com/ros/rosdistro/blob/7249ca98502a9efef4d96ef9c0feb09e9c0fc62c/rosdep/base.yaml#L3756-L3761)

Also, I would have used `target_include_directories`, but I think I need CMake 3.0.2 for that and I didn't want to mess with the minimum CMake version, but I might propose something like that in the future.

### DIFF
--- a/map_server/CMakeLists.txt
+++ b/map_server/CMakeLists.txt
@@ -8,33 +8,49 @@ find_package(catkin REQUIRED
             nav_msgs
         )
 
-find_package(Boost REQUIRED COMPONENTS system)
+find_package(SDL REQUIRED)
+find_package(SDL_image REQUIRED)
 
-find_package(PkgConfig)
-pkg_check_modules(NEW_YAMLCPP yaml-cpp>=0.5)
-if(NEW_YAMLCPP_FOUND)
-add_definitions(-DHAVE_NEW_YAMLCPP)
-endif(NEW_YAMLCPP_FOUND)
+find_package(PkgConfig REQUIRED)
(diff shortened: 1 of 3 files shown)


wjwwood:
Also, the SDL and SDL_image CMake Modules ship with CMake:

- https://cmake.org/cmake/help/latest/module/FindSDL.html
- https://cmake.org/cmake/help/latest/module/FindSDL_image.html

I couldn't find where they were initially introduced, but I think it was a long time ago. The build farm will tell if they were recent additions.

mikaelarguedas:
For reference: these 2 files are present on the 2.8.3 branch of the cmake repo:
https://gitlab.kitware.com/cmake/cmake/blob/v2.8.3/Modules/FindSDL.cmake
https://gitlab.kitware.com/cmake/cmake/blob/v2.8.3/Modules/FindSDL_image.cmake

PR #561, opened 2017-03-03
refactor to not use tf version 1

wjwwood:
I was going to convert this to `tf2`, but it turns out that it really only needs to convert some Euler angles to Quaternion, and @tfoote's recommendation was to just use bullet directly:

https://github.com/ros2/navigation/pull/1#discussion_r104065561

So that's what's happening here instead.

- Bullet's CMake module: https://cmake.org/cmake/help/v3.0/module/FindBullet.html
- Bullet's rosdep rule: https://github.com/ros/rosdistro/blob/7249ca98502a9efef4d96ef9c0feb09e9c0fc62c/rosdep/base.yaml#L393-L434
- Bullet API being used: http://bulletphysics.org/Bullet/BulletFull/classbtQuaternion.html#a808b430136b4a8d1e7120134af14a1ff


tfoote (approved):

mikeferguson:
Seems reasonable -- I presume you've actually loaded a map and tested this?

wjwwood:
I have used this patch in the ROS 2 demo work we've been doing. I haven't tried it in ROS 1 yet.

I just resolved the conflicts too.

vrabaud:
I probably need to rebase by now but there is a full tf2 port of navigation since hydro: https://github.com/ros-planning/navigation/pull/531

wjwwood:
@vrabaud that would be great to get in, but I avoided using tf2's internal transform in favor of bullets in this pr at @tfoote's recommendation. Your pr does the former I think:

https://github.com/ros-planning/navigation/pull/531/files#diff-131b6b4d9e11b01840556ffeeb9acfe0

So if we close this in favor of #531, then I think we should update it to fix `image_loader` and `map_saver` like this pr does.

vrabaud:
Your PR is smaller so go with it, I'll adapt. Then again, there is still some tf1 in there you need to also fix in your current PR:
https://github.com/ros-planning/navigation/blob/kinetic-devel/map_server/src/map_saver.cpp#L35

wjwwood:
@vrabaud that's true. I guess I missed it from my port in our ROS 2 demo, but I just used the `tf2` version of that header:

https://github.com/ros2/navigation/blob/ros2/map_server/src/map_saver.cpp#L38

vrabaud:
Still not working. From my experience when switching to tf2: just do a grep on "tf/" or "tf::". You still have a "tf::".
Please refer to: https://github.com/ros-planning/navigation/pull/531/files#diff-11fba1ba2b27983920dd7eab46a366dd

wjwwood:
@vrabaud I was being lazy, and then remembered that pr testing is currently down. I'll fix it up asap.

Also we could roll this into your pr and close this one. Wouldn't matter to me.

vrabaud:
Naah, go with this one. The smaller the better.

vrabaud:
@vrabaud, that's not what she said !

wjwwood:
Haha, ok I'll fix this one up asap.

nuclearsandwich:
@ros-pull-request-builder retest this please.

vrabaud:
LGTM, feel free to comment on https://github.com/ros-planning/navigation/pull/531

wjwwood:
The only test failures was in the costmap tests, I'm thinking it might just be flakey. I'm going to rerun the tests to see.

wjwwood:
Ok, CI and stuff looks good. I still haven't tried this locally with ROS 1 as a sanity check.

vrabaud:
@wjwwood , I have rebased #458 and #531 and used your changes. But let's merge this PR first.

wjwwood:
@vrabaud that's fine by me, but I'm not the maintainer. I'll let one of them decide on when/if to merge this.

mikeferguson:
Ok -- tested on live ROS1 system. Going to squash-merge for ease of pulling into lunar

PR #562, opened 2017-03-21
Add "lethal" costmap update method

hiddewie:
### Use scenario

I have a use scenario where I use a costmap for exploring the environment. A pixel contains `NO_INFORMATION` if it has not been explored. Explored pixels are marked as `FREE_SPACE`. The static layer is used to process the available map to mark some pixels as `LETHAL_OBSTACLE` on the costmap, for buffering them (inflation layer) and avoiding obstacles as explore target. Unknown pixels must be ignored when updating the costmap.

### Problem

At this moment the static layer has two options for updating the costmap from the map information, namely using the `maximum` or `overwrite` method. This is controlled by the option `use_maximum`. A third option is required to solve the use scenario.

### Solution

Add an extra update method in the costmap code which implements the desired behaviour, and use this update method for the static costmap, if the user desires it. This requires changing the static costmap layer option from a boolean `use_maximum` to a string `update_method` where each possible update method is a named option.

### Changes

- Added costmap update method `CostmapLayer::updateLethalIgnoreUnknown`.
- Changed parameter `use_maximum` (`boolean`) to `update_method` (`std::string`)

---

If required, I can re-add the old parameter `use_maximum` which adds backward compatibility.

### DIFF
--- a/costmap_2d/include/costmap_2d/costmap_layer.h
+++ b/costmap_2d/include/costmap_2d/costmap_layer.h
@@ -103,6 +103,16 @@ protected:
    * Updates the master_grid within the specified
    * bounding box using this layer's values.
    *
+   * Overwrites the master's grid values with the this layer's values, if
+   * this layer's value is not NO_INFORMATION. If the master's value is
+   * NO_INFORMATION it is only updated if this layer's value is LETHAL_OBSTACLE.
+   */
+  void updateLethalIgnoreUnknown(costmap_2d::Costmap2D& master_grid, int min_i, int min_j, int max_i, int max_j);
+
+  /*
+   * Updates the master_grid within the specified
+   * bounding box using this layer's values.
+   *
    * Sets the new value to the sum of the master grid's value
    * and this layer's value. If the master value is NO_INFORMATION,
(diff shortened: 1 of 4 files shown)


mikeferguson:
Backwards compatibility is extremely important, to even consider this the parameters would have to retain "use_maximum". However, this seems like a very specific use case. I'd suggest simply creating your own version of the StaticLayer that carries out the logic you want, because:

 * String comparison in a loop (for the rolling costmap update) is going to be terribly slow
 * I'm guessing you haven't tested this on a rolling costmap -- because the "rolling" looks like it probably won't work, since you don't have any else statement (you're only updating the costmap when it meets the condition for copy, you don't handle the other case)
 * We made costmaps into layers specifically so that people could customize the functionality without adding every possible functionality to the primary codebase. Every set of parameters that gets added increases the testing/maintenance burden.

hiddewie:
Thank you for your comments. You are right, this is a specific use case. 

No problem that the PR has been closed. The most important thing for me is that you, and possible other authors of the package, have the chance to use the code and make it available for other users if you choose to do so.

PR #563, opened 2017-03-23
Make recovery behaviors work by default

llach:
We added two small changes to make the recovery behaviors work when the robot gets stuck.
They were tested almost a year on our indigo branch.


DLu:
There's a couple of things conflated here. It is easier to review if you make the pull requests for one fix at a time.

PR #564, opened 2017-03-31
Added parameter for allowing inflation in unknown cells

hiddewie:
### Problem

Unknown space might have a cost moving over it, especially if the used sensor has a small range. To have the global or local planner make better paths inflation costs are required for unknown space. 

### Current situation 

Only lethal costs are inflated over unknown areas. 

### Solution

I've added a parameter `inflate_unknown` where *any* cost can be inflated over unknown areas. The default value is `false` resulting in the previous behavior for backwards compatibility.

### Screenshots 

Screenshots from `rviz`. The difference is in the top right corner: before no inflation is done over the unknown area, while after the inflated area (red and dark blue) is extended into the unknown area.

Before:
inflate-known

After:
inflate-unknown

### DIFF
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -120,8 +120,9 @@ public:
    * @brief Change the values of the inflation radius parameters
    * @param inflation_radius The new inflation radius
    * @param cost_scaling_factor The new weight
+   * @param inflate_unknown Whether to inflate unknown cells
    */
-  void setInflationParameters(double inflation_radius, double cost_scaling_factor);
+  void setInflationParameters(double inflation_radius, double cost_scaling_factor, bool inflate_unknown);
 
 protected:
   virtual void onFootprintChanged();
@@ -171,6 +172,7 @@ private:
                       unsigned int src_x, unsigned int src_y);
 
   double inflation_radius_, inscribed_radius_, weight_;
+  bool inflate_unknown_;
   unsigned int cell_inflation_radius_;
   unsigned int cached_cell_inflation_radius_;
   std::map<double, std::vector<CellData> > inflation_cells_;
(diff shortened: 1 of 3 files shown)


costmap_2d/include/costmap_2d/inflation_layer.h
```diff
@@ -132,8 +132,9 @@ class InflationLayer : public Layer
    * @brief Change the values of the inflation radius parameters
    * @param inflation_radius The new inflation radius
    * @param cost_scaling_factor The new weight
+   * @param inflate_unknown Whether to inflate unknown cells
    */
-  void setInflationParameters(double inflation_radius, double cost_scaling_factor);
+  void setInflationParameters(double inflation_radius, double cost_scaling_factor, bool inflate_unknown);
```
DLu: The reason that these two parameters were separated is because they are key in computing the cached values. Since inflate_unknown does not affect the cached values, it would be simpler to leave this function definition as it was. Otherwise, it will change the binary compatibility of the object.

costmap_2d/plugins/inflation_layer.cpp
```diff
@@ -99,7 +100,7 @@ void InflationLayer::onInitialize()
 
 void InflationLayer::reconfigureCB(costmap_2d::InflationPluginConfig &config, uint32_t level)
 {
-  setInflationParameters(config.inflation_radius, config.cost_scaling_factor);
+  setInflationParameters(config.inflation_radius, config.cost_scaling_factor, config.inflate_unknown);
 
   if (enabled_ != config.enabled) {
```
DLu: Add to the above line to force reinflation if the inflate_unknown value changes.

DLu (changes requested):
This looks solid, although I haven't tested it yet. A couple comments:
1) Changing the `setInflationParameters` function definition changes the binary compatibility, which makes it harder to patch. Fortunately, it makes more sense not to change it (although I can understand why you did initially).
2) I'm a little unclear exactly what is happening in the pictures. Can you explain a) what elements are present in the visualization b) what the specific change is between the two

hiddewie:
I've added the requested changes. Good idea.

I have also explained the images a bit more to clarify the contents.

mikeferguson:
I've done a quick rebase of this since we merged some other costmap changes. Once CI passes, we'll get this merged. Thanks!

hiddewie:
Thank you, great!

PR #565, opened 2017-04-05
make rostest in CMakeLists optional (ros/rosdistro#3010)

bulwahn:
It is minor change that might be worthwhile to be back-ported to the other branches. There is no need to backport for my use case, though.


mikeferguson:
Thanks!

Note for posterity: cherry-picked to jade/indigo

PR #570, opened 2017-04-20
Add cost function to prevent unnecessary spinning

codebot:
Adds a cost function and configuration parameter to penalize changes in robot heading. This isn't relevant for diff-drive robots, where heading changes are typically necessary to move towards goals. However, this is useful for having control of the 'twirling' behavior of holonomic robots. Setting the twirling_scale parameter to a high value, say 50, causes the dwa planner to prefer to 'strafe' towards
goals rather than (eventually) aiming the robot at the goal while driving. For some applications this is much preferred. This parameter defaults to zero, which maintains the existing behavior unless it is explicitly set to a non-zero value.

I'm happy to iterate on this as needed to get it ready to merge :chicken:


mikeferguson:
LGTM, going to cherry-pick to lunar

PR #461, opened 2016-05-23
Added change_map service to map_server

sloretz:
This PR adds a "change_map" service to map_server including a change_map rostest. It also resolves #279.

The change_map service expects a path to a yaml file, and responds with a Boolean indicating if the map was successfully loaded. The new map and metadata are published to the appropriate topics. 

If change_map is called with the currently loaded map, the current map is reloaded as if it were a brand new map, resolving #279.

PR #575, opened 2017-05-09
Add runtime reload map feature

wiseware-ros:
subscribed topic: 
        ~/reload (std_msgs/String) 
Provides a non-action interface to map_server for reloading the map information while map_server 
is running. 
Message data is a string with the file name.
If message data is empty, map_server reloads the map from the last specified file.


DLu:
A) Are you still interested in this feature? 
B) Would a service like the one in #461 be sufficient?
C) This PR has a lot of additional changes outside the scope of the proposed change, which makes it difficult to review.

PR #576, opened 2017-05-12
Update gradient_path.cpp

ljy1992:
correct an obvious error at line 285


DLu:
Please update this to fix the equivalent line in navfn too

ljy1992:
I've changed the same place to the navfn code. I'm new to git, so is this the correct operation to add one more commits?

mikeferguson:
@DLu -- is your comment an approval of this fix? If so, I will merge and forward port to K+L

PR #581, opened 2017-05-30
Indigo to Jade

DLu:
Forward porting a couple of PRs (#440, #470, #543, #498) to Jade


mikeferguson:
Seems reasonable, probably won't get a Debian release since Jade is EOL

PR #582, opened 2017-06-01
Fixed planner race condition

PPokorski:
There was a race condition inside planner's makePlan() function, which was especially dangerous for high frequencies (for me: 15 Hz for planner, 10 Hz for costmap). 

Actually it would be better to ensure that **all** costmap's functions are thread-safe, but this hotfix seems to work for me.

Cheers and thanks for your work :)

### DIFF
--- a/global_planner/src/planner_core.cpp
+++ b/global_planner/src/planner_core.cpp
@@ -253,6 +253,8 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
     unsigned int start_x_i, start_y_i, goal_x_i, goal_y_i;
     double start_x, start_y, goal_x, goal_y;
 
+    boost::unique_lock<costmap_2d::Costmap2D::mutex_t> costmap_lock(*costmap_->getMutex());
+
     if (!costmap_->worldToMap(wx, wy, start_x_i, start_y_i)) {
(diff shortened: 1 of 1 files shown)


mikeferguson:
I'm not sure how this isn't redundant:
 * the planThread calls MoveBase::makePlan() (https://github.com/ros-planning/navigation/blob/kinetic-devel/move_base/src/move_base.cpp#L570)
 * in MoveBase::makePlan() the first think we do is lock the costmaps (https://github.com/ros-planning/navigation/blob/kinetic-devel/move_base/src/move_base.cpp#L450) before calling planner_->makePlan()

PPokorski:
Ok, that's interesting. Actually I'm not using move_base, just plain global_planner, that's why I stumbled across this problem. The question now is: should the global_planner or move_base lock  the costmap?

mikeferguson:
That makes total sense. Going to close this and point at the ticket that discusses where we should lock -- https://github.com/ros-planning/navigation/issues/441 -- I'd like to make all of those changes in a single PR.

The discussion in that ticket points out that we would like to move logs into the planners, but since we don't have access to all the planners, we should really only do it in a recent platform (changing indigo/kinetic is pretty dangerous to others who might be relying on that lock)

Issue #586, opened 2017-06-16
call of overloaded 'abs(unsigned int)' is ambiguous in amcl with gcc-7.1


bulwahn:
When compiling amcl with gcc-7.1 in the cross-compiliation setup of meta-ros, amcl fails with:
```
navigation-1.12.13/amcl/src/amcl/map/map_cspace.cpp:98:34: error: call of overloaded 'abs(unsigned int)' is ambiguous
|    unsigned int di = abs(i - src_i);
|                                   ^
```

The reason for gcc raising this error is also described in the gcc porting page (https://gcc.gnu.org/gcc-6/porting_to.html#overloaded-abs).

A simple solution is to avoid to use the `abs` function, as already suggested in 
https://github.com/ros-planning/navigation/issues/518.

The full error message during compiling is:
```
[...]

Issue #518, opened 2016-08-31
amcl segfault in libamcl_map.so


nxdefiant:
amcl segfaults when compiled from source in current debian testing (gcc 6.1.1, boost 1.61, kinetic-devel)

ros debug output is attached in amcl.txt:

[amcl.txt](https://github.com/ros-planning/navigation/files/448150/amcl.txt)

nxdefiant:
The crash is in map/map_cspace.cpp, enqueue():

```
unsigned int di = abs(i - src_i);
unsigned int dj = abs(j - src_j);
```

(i, j, src_i, src_j also unsigned int)

with e.g. i=997 and src_i=998 on g+ 6.1.1 it results in 4294967295 instead of 1 because g++ now has its own abs() in math.h. The values should probably cast to signed or the lines converted to something like

`unsigned int di = i > src_i ? i - src_i : src_i - i;`

mikeferguson:
Likely fixed by #587

PR #587, opened 2017-06-16
amcl: fix compilation with gcc v7

rojkov:
When building amcl with recent enough gcc v7 compilation fails with
the error
```
src/amcl/map/map_cspace.cpp: In function 'void enqueue(map_t*, unsigned int, unsigned int, unsigned int, unsigned int, std::priority_queue<CellData>&, CachedDistanceMap*, unsigned char*)':
src/amcl/map/map_cspace.cpp:98:34: error: call of overloaded 'abs(unsigned int)' is ambiguous
   unsigned int di = abs(i - src_i);
```
Use `int abs(int)` flavour of the abs() function.

### DIFF
--- a/amcl/src/amcl/map/map_cspace.cpp
+++ b/amcl/src/amcl/map/map_cspace.cpp
@@ -86,8 +86,8 @@ get_distance_map(double scale, double max_dist)
   return cdm;
 }
 
-void enqueue(map_t* map, unsigned int i, unsigned int j, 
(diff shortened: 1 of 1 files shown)


bulwahn:
This commit resolves #586, and probably also #518.

mikeferguson:
Cherry-picked to lunar as well (not going to back port to indigo unless somebody speaks up pointing out we're using new enough compilers with indigo)

PR #590, opened 2017-07-16
Cleanup PCL imports: fixes compilation on zesty and stretch

mikaelarguedas:
This is a more elegant alternative to #578 , this fixes compilation errors on Zesty and Stretch.
It also allow to reduce the amount of PCL libraries these packages link against and add missing dependencies on pcl for packages using it.

### DIFF
--- a/base_local_planner/CMakeLists.txt
+++ b/base_local_planner/CMakeLists.txt
@@ -21,9 +21,7 @@ find_package(Boost REQUIRED
         )
 
 find_package(Eigen3 REQUIRED)
-find_package(PCL REQUIRED)
-# For debian: https://github.com/ros-perception/perception_pcl/issues/139
-find_package(Qt5Widgets QUIET)
+find_package(PCL REQUIRED COMPONENTS common)
 remove_definitions(-DDISABLE_LIBUSB-1.0)
 include_directories(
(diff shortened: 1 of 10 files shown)


mikeferguson:
Triggering CI

mikeferguson:
LGTM -- waiting for CI to complete, and then I'd like to merge this BEFORE #591 (which will then require a fixup)

mikeferguson:
I'm not sure how that test failure would occur with build change... but this is changing the related package, so I think we need to look into that

mikaelarguedas:
I tried to reproduce the test failure locally without success (running a prerelease job with this branch). Not sure why the test didnt start properly on the PR job.

> I'd like to merge this BEFORE #591 (which will then require a fixup)

:+1: happy to adapt #591 accordingly once this one gets merged

mikaelarguedas:
Closed involuntarily this PR. The new CI launched when reopening seems to be green, so I dont think that the failure seen in http://build.ros.org/job/Lpr__navigation__ubuntu_xenial_amd64/5/ is related to this PR

mikeferguson:
Thanks for retriggering again -- looking over things it looks like that test is failing intermittently.

Issue #124, opened 2013-09-05
Obstacle layer ignores range_min / range_max


ZdenekM:
Obstacle layer should deal somehow with range_min / range_max values from LaserScan message.

ZdenekM:
Well, range_max is quite clear. In case of range_min my proposal is to start raytracing on clearing operation not at sensor origin but in range_min distance from sensor. In case of marking - it might be configurable (mark when dist < range_min vs. don't mark).

procopiostein:
+1

PR #594, opened 2017-07-31
Adding "start clearing range" to the Obstacle Layer. 

adrienbarral:
Start clearing range is the minimal distance at which a sensor can see an obstacle. This is useful for all stereo based sensor (like RGBD and stereo cameras). Obstacles  will not be cleared between the sensor and this range.

I code this patch because I encounter the follwing problem. I have a mobile robot with an RGBD sensor moving in a place where there is a lot of shelfs (or low table). When the robot is far of this objects, there is no problem, they are seend as obstacles. But if the robot go close of this objects, they are cleared as the robot move...

If you want further explanation ask me, I can create a video With and Without my patch.

### DIFF
--- a/costmap_2d/include/costmap_2d/observation.h
+++ b/costmap_2d/include/costmap_2d/observation.h
@@ -51,7 +51,7 @@ public:
    * @brief  Creates an empty observation
    */
   Observation() :
-    cloud_(new pcl::PointCloud<pcl::PointXYZ>()), obstacle_range_(0.0), raytrace_range_(0.0), start_clearing_range_(0.0)
+    cloud_(new pcl::PointCloud<pcl::PointXYZ>()), obstacle_range_(0.0), max_raytrace_range_(0.0), min_raytrace_range_(0.0)
   {
   }
 
@@ -61,16 +61,34 @@ public:
   }
 
   /**
+   * @brief  Creates an observation from an origin point and a point cloud.
+   * This constructor is deprecated, use the constructor with min and max raytrace range now.
+   * @param origin The origin point of the observation
+   * @param cloud The point cloud of the observation
+   * @param obstacle_range The range out to which an observation should be able to insert obstacles
+   * @param max_raytrace_range The range out to which an observation should be able to clear via raytracing
+   */
+  ROS_DEPRECATED
+  Observation(geometry_msgs::Point& origin, pcl::PointCloud<pcl::PointXYZ> cloud,
+              double obstacle_range, double max_raytrace_range) :
+      origin_(origin), cloud_(new pcl::PointCloud<pcl::PointXYZ>(cloud)),
+      obstacle_range_(obstacle_range), min_raytrace_range_(0.0), max_raytrace_range_(max_raytrace_range)
+  {
+  }
+
+  /**
    * @brief  Creates an observation from an origin point and a point cloud
    * @param origin The origin point of the observation
    * @param cloud The point cloud of the observation
    * @param obstacle_range The range out to which an observation should be able to insert obstacles
-   * @param raytrace_range The range out to which an observation should be able to clear via raytracing
+   * @param min_raytrace_range The range from which an observation should be able to clear via raytracing
+   * @param max_raytrace_range The range out to which an observation should be able to clear via raytracing
    */
   Observation(geometry_msgs::Point& origin, pcl::PointCloud<pcl::PointXYZ> cloud,
-              double obstacle_range, double raytrace_range, double start_clearing_range=0.0) :
-      origin_(origin), cloud_(new pcl::PointCloud<pcl::PointXYZ>(cloud)),
-      obstacle_range_(obstacle_range), raytrace_range_(raytrace_range),start_clearing_range_(start_clearing_range)
+              double obstacle_range, double min_raytrace_range, double max_raytrace_range) :
+          origin_(origin), cloud_(new pcl::PointCloud<pcl::PointXYZ>(cloud)),
+          obstacle_range_(obstacle_range), min_raytrace_range_(min_raytrace_range),
+          max_raytrace_range_(max_raytrace_range)
   {
   }
 
@@ -80,8 +98,8 @@ public:
    */
   Observation(const Observation& obs) :
       origin_(obs.origin_), cloud_(new pcl::PointCloud<pcl::PointXYZ>(*(obs.cloud_))),
-      obstacle_range_(obs.obstacle_range_), raytrace_range_(obs.raytrace_range_),
-      start_clearing_range_(obs.start_clearing_range_)
+      obstacle_range_(obs.obstacle_range_), min_raytrace_range_(obs.min_raytrace_range_),
+      max_raytrace_range_(obs.max_raytrace_range_)
   {
   }
 
@@ -91,16 +109,14 @@ public:
    * @param obstacle_range The range out to which an observation should be able to insert obstacles
    */
   Observation(pcl::PointCloud<pcl::PointXYZ> cloud, double obstacle_range) :
-      cloud_(new pcl::PointCloud<pcl::PointXYZ>(cloud)), obstacle_range_(obstacle_range), raytrace_range_(0.0),
-      start_clearing_range_(0.0)
+      cloud_(new pcl::PointCloud<pcl::PointXYZ>(cloud)), obstacle_range_(obstacle_range), min_raytrace_range_(0.0),
+      max_raytrace_range_(0.0)
   {
   }
 
   geometry_msgs::Point origin_;
   pcl::PointCloud<pcl::PointXYZ>* cloud_;
-  double obstacle_range_, raytrace_range_;
-  double start_clearing_range_;
-
+  double obstacle_range_, min_raytrace_range_, max_raytrace_range_;
 };
 
 }  // namespace costmap_2d
--- a/costmap_2d/include/costmap_2d/observation.h
+++ b/costmap_2d/include/costmap_2d/observation.h
@@ -51,7 +51,7 @@ public:
    * @brief  Creates an empty observation
    */
   Observation() :
-    cloud_(new pcl::PointCloud<pcl::PointXYZ>()), obstacle_range_(0.0), max_raytrace_range_(0.0), min_raytrace_range_(0.0)
+    cloud_(new pcl::PointCloud<pcl::PointXYZ>()), max_obstacle_range_(0.0), max_raytrace_range_(0.0), min_raytrace_range_(0.0)
   {
   }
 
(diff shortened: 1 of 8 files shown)


mikeferguson:
This is half of https://github.com/ros-planning/navigation/issues/124

costmap_2d/include/costmap_2d/observation_buffer.h
```diff
@@ -70,15 +70,16 @@ class ObservationBuffer
    * @param  max_obstacle_height The minimum height of a hitpoint to be considered legal
    * @param  obstacle_range The range to which the sensor should be trusted for inserting obstacles
    * @param  raytrace_range The range to which the sensor should be trusted for raytracing to clear out space
+   * @param  start_clearing_range The range to wich the sensor start to be trusted for clearing out space.
```
romainreignier: "to wHich"

adrienbarral:
Yes this look like to be the half of #124. But my patch works for all data source (not only LaserScan). I think this is important to introduce a new parameter in the source properties rather than use the min_rage data of LaserScan. Because, using this new parameter "start_clearing_range", we can work with all data sources (not only LaserScan).

procopiostein:
I agree that your implementation is more generic than of  #124.
But I would propose a different nomeclature, to keep some consistency with the other parameters, for example, naming it "min_raytrace_range" which defaults to 0. 
Also there could be another parameter "min_obstacle_range" that would do the other half of #124, controlling from what distance the obstacles should be marked.

adrienbarral:
Ok, I can rename my parameter to min_raytrace_range. And I will implement the second half of #124 with this "min_obstacle_range".
To be sure that I well understood what "min_obstacle_range" must do. If we have an hit (a point sensed) at a range of 2 meters and that min_obstacle_range is 1.5m, this point must be discarded (no clearing, and no point set). Isn't it  ?
Or should I clear between min_raytrace_range and min_obstacle_range, and not set the point as a lethal obstacle ? Or something else ?

I think I can post seomething before the end of this week.

procopiostein:
> To be sure that I well understood what "min_obstacle_range" must do. If we have an hit (a point sensed) at a range of 2 meters and that min_obstacle_range is 1.5m, this point must be discarded (no clearing, and no point set). Isn't it ?

Yes, that how I would interpret  #124 second part in a more generic form.
Anyway, maybe you should wait for some feedback from @DLu and @mikeferguson on that.

thanks!

costmap_2d/include/costmap_2d/observation.h
```diff
@@ -51,7 +51,7 @@ class Observation
    * @brief  Creates an empty observation
    */
   Observation() :
-    cloud_(new pcl::PointCloud<pcl::PointXYZ>()), obstacle_range_(0.0), raytrace_range_(0.0)
+    cloud_(new pcl::PointCloud<pcl::PointXYZ>()), obstacle_range_(0.0), raytrace_range_(0.0), start_clearing_range_(0.0)
```
mikeferguson: Let's call this new parameter "min_raytrace_range_". Let's also update raytrace_range_ to be max_raytrace_range_

costmap_2d/include/costmap_2d/observation.h
```diff
@@ -68,9 +68,9 @@ class Observation
    * @param raytrace_range The range out to which an observation should be able to clear via raytracing
    */
   Observation(geometry_msgs::Point& origin, pcl::PointCloud<pcl::PointXYZ> cloud,
-              double obstacle_range, double raytrace_range) :
+              double obstacle_range, double raytrace_range, double start_clearing_range=0.0) :
       origin_(origin), cloud_(new pcl::PointCloud<pcl::PointXYZ>(cloud)),
-      obstacle_range_(obstacle_range), raytrace_range_(raytrace_range)
+      obstacle_range_(obstacle_range), raytrace_range_(raytrace_range),start_clearing_range_(start_clearing_range)
```
mikeferguson: let's leave this function signature without change (you won't be able to set min through it), but default min_raytrace_range_ to 0.0, and add a deprecation on this constructor. Let's then create a new constructor which has min_range BEFORE max_range.

This way, code people have will still compile (albeit with a deprecation warning) but new code that sets mins will look more "normal" by having the min parameter before the max parameter in the parameter list.

costmap_2d/include/costmap_2d/observation.h
```diff
@@ -90,13 +91,16 @@ class Observation
...
+      cloud_(new pcl::PointCloud<pcl::PointXYZ>(cloud)), obstacle_range_(obstacle_range), raytrace_range_(0.0),
+      start_clearing_range_(0.0)
   {
   }
 
   geometry_msgs::Point origin_;
   pcl::PointCloud<pcl::PointXYZ>* cloud_;
   double obstacle_range_, raytrace_range_;
+  double start_clearing_range_;
+
```
mikeferguson: remove extra newline, uneeded change

costmap_2d/include/costmap_2d/observation_buffer.h
```diff
@@ -70,15 +70,16 @@ class ObservationBuffer
...
    * @param  global_frame The frame to transform PointClouds into
    * @param  sensor_frame The frame of the origin of the sensor, can be left blank to be read from the messages
    * @param  tf_tolerance The amount of time to wait for a transform to be available when setting a new global frame
    */
   ObservationBuffer(std::string topic_name, double observation_keep_time, double expected_update_rate,
                     double min_obstacle_height, double max_obstacle_height, double obstacle_range,
-                    double raytrace_range, tf::TransformListener& tf, std::string global_frame,
-                    std::string sensor_frame, double tf_tolerance);
+                    double raytrace_range, double start_clearing_range, tf::TransformListener& tf,
+                    std::string global_frame, std::string sensor_frame, double tf_tolerance);
```
mikeferguson: same comments as we did for Observation -- leave this constructer, default min_raytrace_range to 0.0, deprecate constructor. Add new constructor with min before max. Since we'll rename raytrace range to max_raytrace_range, let's also make sure the doxygen gets updated

costmap_2d/include/costmap_2d/observation_buffer.h
```diff
@@ -159,6 +160,7 @@ class ObservationBuffer
   boost::recursive_mutex lock_;  ///< @brief A lock for accessing data in callbacks safely
   double obstacle_range_, raytrace_range_;
   double tf_tolerance_;
+  double start_clearing_range_;
```
mikeferguson: reaname to min_raytrace_range_, update raytrace_range to be max_raytrace_range_, and let's declare all of them on the same line above: "double obstacle_range_, min_raytrace_range_, max_raytrace_range_".

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -114,7 +114,7 @@ void ObstacleLayer::onInitialize()
       throw std::runtime_error("Only topics that use point clouds or laser scans are currently supported");
     }
 
-    std::string raytrace_range_param_name, obstacle_range_param_name;
+    std::string raytrace_range_param_name, obstacle_range_param_name, start_clearing_range_param_name;
```
mikeferguson: i would update these to be min/max_raytrace_range_param_name

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -130,15 +130,21 @@ void ObstacleLayer::onInitialize()
       source_node.getParam(raytrace_range_param_name, raytrace_range);
     }
 
+    double start_clearing_range = 0.0;
+    if (source_node.searchParam("start_clearing_range", start_clearing_range_param_name))
```
mikeferguson: min_raytrace_range

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -130,15 +130,21 @@ void ObstacleLayer::onInitialize()
       source_node.getParam(raytrace_range_param_name, raytrace_range);
```
mikeferguson: We should update this to search for max_raytrace_range, if not found, it looks for raytrace_range, but prints a warning that raytrace_range is deprecated (and yes, this starts to leak into the config for each robot, but we will give several/many release cycles before we remove the deprecated part)

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -525,6 +531,31 @@ void ObstacleLayer::raytraceFreespace(const Observation& clearing_observation, d
   {
     double wx = cloud.points[i].x;
     double wy = cloud.points[i].y;
+    if(clearing_observation.start_clearing_range_ > 0)
```
mikeferguson: space after if

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -525,6 +531,31 @@ void ObstacleLayer::raytraceFreespace(const Observation& clearing_observation, d
...
+    {
+      // If this sensor start seeing obstacle at a certain range, we must move x0,y0 at this
+      // range from the sensor (following the line going from the sensor to the point)
+      double obstacle_range = hypot(wx-ox, wy-oy);
+      if(obstacle_range < clearing_observation.start_clearing_range_)
```
mikeferguson: space after if

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -525,6 +531,31 @@ void ObstacleLayer::raytraceFreespace(const Observation& clearing_observation, d
...
+      {
+        // This may happens if parameter start_clearing_range was set to a range lesser than the minimal sensor range.
+        // In this particular case, we just don't set free spaces (after displaying a warning).
+        ROS_WARN_THROTTLE(
+                1.0, "Your start_clearing_range parameter seems to be too short because an observation was returned by"
+                "your sensor in this range. Cells between sensor and this observation was not cleared");
```
mikeferguson: I'm not a fan of this log message -- there are certainly sensors that report 0 when they cannot see something, and this would spam the crap out of the console/logs. I would suggest we either remove this, or make it a ROS_DEBUG_NAMED_THROTTLE and increase the throttle to something like 60.0.

costmap_2d/plugins/obstacle_layer.cpp
```diff
@@ -525,6 +531,31 @@ void ObstacleLayer::raytraceFreespace(const Observation& clearing_observation, d
...
+        continue;
+      }
+      double scale_factor = clearing_observation.start_clearing_range_/obstacle_range;
+      double x_start_clearing = (wx-ox)*scale_factor;
+      double y_start_clearing = (wy-oy)*scale_factor;
```
mikeferguson: you aren't actually using these below in the raytrace -- so none of this computation is yet changing anything in your costmap!

mikeferguson: Also, x_start_clearing needs to be added to ox at some point (same for y) to get an actual point. Your worldToMap check is checking the wrong points.

mikeferguson (changes requested):
I like @procopiostein suggestion that we also implement min_obstacle_range -- I would suggest we go all the way with this and also rename obstacle_range to max_obstacle_range.

mikeferguson:
For clarity:

 * Let us define D as the distance from the sensor origin to the measured point (x,y) seen by our sensor.
 * Let us define R as a ray/line from the sensor origin to the measured point (x,y). Note: the ray does not include the actual voxel in which the measured point falls.

And here is how the parameters should probably work: 

 * min_obstacle_range - if D < min_obstacle_range, we will not mark x,y as an obstacle
 * max_obstacle_range (currently just obstacle_range) - if D > max_obstacle_range, we will not mark x,y as an obstacle.
 * min_raytrace_range - if D > min_raytrace_range, then we will start clearing along ray R from a distance min_raytrace_range from the sensor origin.
 * max_raytrace_range (currently just raytrace_range) - if D > max_raytrace_range, then we will stop clearing along ray R at a distance max_raytrace_range from the origin.

Should min_raytrace_range = 0, and the ray R falls entirely within the map_bounds, and D < min_raytrace_range, then we will raytrace all of ray R.

costmap_2d/plugins/obstacle_layer.cpp
adrienbarral: About your first comment : In the next line, I call
`worldToMap(x_start_clearing, y_start_clearing, x0, y0)`
So this method will update x0, and y0. Both variables are then passed in argument to the method "raytraceLine". So I think that this have an impact in my costmap.

About your second comment : I can't find any emote icon to describe all my shame !!!!

costmap_2d/plugins/obstacle_layer.cpp
adrienbarral: And what if I display this message only if obstacle_range > 0 (and < to min_raytrace_range) ? This will handle the particular case of sensor that report 0 when they cannot see.

mikeferguson:

PR #597, opened 2017-08-01
don't update costs if inflation radius is zero

### LINKED ISSUE #556: bug in inflation_radius .
bug as follows:
if deliberately set ` inflation_radius=0.0` in common_cost.yaml will cause move_base hang down.
as belows.  the code should protect the value's scale settings.

inflation

mikeferguson:
fixes #556


mikeferguson:
Cherry-picked to K&L

PR #599, opened 2017-08-01
recompute cluster stat when force_publication

mikeferguson:
When force_publication is True, and no resample has been done the statistic
about the current clusters of particles are not recomputed. So amcl published
an outdated information.

This commit introduces a call to `pf_cluster_stats` when required (i.e no
resample but force_publication).


mikeferguson:
Forward port of #403

PR #601, opened 2017-08-01
#520 for lunar [ABI Breaking]

mikeferguson:
Per review, I'm going to merge this as-is into Lunar. For I+K, we need to avoid the ABI break of costmap2dROS


tfoote:

PR #602, opened 2017-08-01
Inflation Layer protected members and virtual computeCost [ABI BREAKING]

mikeferguson:
This is #534 but targeting lunar


tfoote:

Issue #592, opened 2017-07-20
AMCL doesn't throw error when it receives maps on the correct topic but not matching the global_frame_id


vik748:
Currently,
AMCL seems to proceed as if every thing is good if it receives maps on the specified map_topic even if the frame_id on the map doesn't match global_frame_id.  In this case AMCL essentially is unable to scan map to the map being published and thus the error on the particles keep growing.  A simple fix should be able to warn users that the map being published is not being used by AMCL because of the mismatch.

Cheers,
Vik

mikeferguson:
Seems reasonable -- pull request for proposed simple fix would be accepted.

vik748:
@mikeferguson I am very new to github.  What is the best way for me to submit the changes?  Do I created a new forked repo and then submit that or do I create a new branch in this repo and request that to be merged?

Thanks,
Vik

mikeferguson:
@vik748 new forked repo (you won't have permissions to create branches on this particular repo). Ideally a feature branch within your repo (so you can easily pull in updates to your fork). Then open a PR against this repo. Thanks!

vik748:
Pull request #604 submitted.

mikeferguson:
Fixed in kinetic & lunar

PR #604, opened 2017-08-01
Reference Issue #592 Added warning to AMCL when map is published on ...

vik748:
...correct topic but its frame_id doesn't match the global_frame_id

### DIFF
--- a/amcl/src/amcl_node.cpp
+++ b/amcl/src/amcl_node.cpp
@@ -802,6 +802,11 @@ AmclNode::handleMapMessage(const nav_msgs::OccupancyGrid& msg)
            msg.info.width,
            msg.info.height,
            msg.info.resolution);
+  
+  if(msg.header.frame_id != global_frame_id_)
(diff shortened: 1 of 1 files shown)


amcl/src/amcl_node.cpp
```diff
@@ -802,6 +802,11 @@ AmclNode::handleMapMessage(const nav_msgs::OccupancyGrid& msg)
            msg.info.width,
            msg.info.height,
            msg.info.resolution);
+  
+  if(msg.header.frame_id != global_frame_id_)
+  		ROS_WARN("Frame_id of map received:'%s' doesn't match global_frame_id:'%s;'. This could cause issues with reading published topics",
```
mikeferguson: this PR has a mix of tabs + spaces -- which is why github is rendering it odd. The amcl code base isn't perfect on style guide, but I do try to ask new code be mostly compatible. Could you update your changes to use ONLY spaces, and in particular 2 spaces indent for each new block (you can just push a new commit and I will squash later). Thanks!

vik748: Done.

mikeferguson:
Cherry picked to lunar

PR #605, opened 2017-08-02
move_base_flex API support - more flexible navigation - BPI Change

spuetz:
Changes for Move Base Flex. We want plugins / planners which can be used with move_base as well as with Move Base Flex in Lunar. We'll present Move Base Flex at the ROSCon 2017 in Vancouver.  So this pull request is a bit urgent. @mikeferguson 
See  https://github.com/magazino/move_base_flex

About MoveBaseFlex:
Move Base Flex (MBF) is a backwards-compatible replacement for move_base. MBF can use existing plugins for move_base, and provides an enhanced version of the same ROS interface. It exposes action servers for planning, controlling and recovering, providing detailed information of the current state and the plugin's feedback. An external executive logic can use MBF and its actions to perform smart and flexible navigation strategies. For example, Magazino has successfully deployed MBF at customer facilities to control TORU robots in highly dynamical environments. Furthermore, MBF enables the use of other map representations, e.g. meshes. The core features are:

1. Fully backwards-compatible with current ROS navigation.
2. Actions for the submodules planning, controlling and recovering, and services to query the costmaps are provided. This interface allows external executives, e.g. SMACH, or Behavior Trees, to run highly flexible and complex navigation strategies.
3. Comprehensive result and feedback information on all actions, including error codes and messages from the loaded plugins. For users still relying on a unique navigation interface, we have extended move_base action with detailed result and feedback information (though we still provide the current one).
4. Separation between an abstract navigation framework and concrete implementations, allowing faster development of new applications, e.g. 3D navigation.

### DIFF
--- a/nav_core/include/nav_core/abstract_global_planner.h
+++ b/nav_core/include/nav_core/abstract_global_planner.h
@@ -1,0 +1,91 @@
+/*
+ *  Copyright 2017, Sebastian Pütz
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *  1. Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *
+ *  2. Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *
+ *  3. Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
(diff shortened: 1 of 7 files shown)


corot:
@mikeferguson, We have noticed today that the first ROS navigation Lunar release is on its way! I hope we are still on time with this PR, before it makes its way to the official repository.

mikeferguson:
@corot -- just as a note, I'm actually working on the 2nd release into Lunar. @DLu did a release a few weeks ago, which is building on all platforms except debian. This second release will fix the debian builds, and also includes a number of outstanding PRs. Some of those already merged PRs are ABI/API breaking, so I'm not concerned  yet about API/ABI in Lunar, but we do need to move fast on any additional ABI/API changes.

mikeferguson:
A couple of thoughts/questions before I really dig in review this:

 * Why add new abstract classes, if you are deprecating the entire non-abstract interface? Why not just add the new functions to the existing base classes and then deprecate the older interfaces over time?

 * If we are going to deprecate the older interfaces, we should have a planned timeline on when nav_core/etc will be migrated to new interfaces. I'm not a fan of blasting a bunch of deprecation warnings at people, IMO deprecation warnings are there for other folks outside the navigation stack who might be using the interfaces so that they have time to migrate their code.

corot:
Great to hear we still have time to integrate this change. @mikeferguson, or anyone else, please feel free to criticize or propose changes for whatever you dislike. Current PR is an attempt to overcome the limitation of move_base calls to plugins (with just a true/false result) while minimizing impact on the BPI. And so being is far from perfect! We had also considered to replace the error_code / msg arguments with new messages to be added to nav_msgs: nav_msgs/LocalPlannerInfo, and nav_msgs/GlobalPlannerInfo, passed as reference. But we didn't cause we though it will make this PR more cumbersome to integrate.

corot:
Answering your questions...
- One of the ideas behind move base flex is to decouple an abstract scaffolding for a navigation system from particular implementations (e.g. move_base). But @spuetz can explain you better this point.
- Honestly, didn't think on this! Probably makes sense to remove on next LTS ROS version

spuetz:
The abstract classes are made for a more abstract navigation stack, which is also able to do navigation tasks without the costmap_2d stuff. The costmap_2d stuff is bound in the BaseLocalPlanner, BaseGlobalPlanner and the RecoveryBehavior.  MoveBaseFlex is templated in the choice of the abstract base class. Due to that, the BaseGlobalPlanner, BaseLocalPlanner and the RecoveryBehavior represent the way for planning with costmap_2d, and it is - on the one hand - still compatible with move_base and - on the other hand - provides a richer interface for move_base_flex. Here all planners are working with move_base_flex and with move_base in the current state (as they are right now, but compiled with the new nav_core). 
Currently we are on writing an interface for GridMap (https://github.com/ethz-asl/grid_map) planners. That interface also inherits the abstract base classes. While the planner execution logic and the server logic is still the same in Move Base Flex.

Deprecating the old version is not our main plan. The developers should still be able to use the current implementations with both move_base and Move Base Flex. But we want them to have a lock for move_base_flex, which can solve many problems.

corot:
Hey @mikeferguson, do you need any additional information or details? We would really like to make this merged on next Lunar release (or for sure before the next API/ABI-freezing release), so we are more than ready to discuss and or modify the PR.
Btw, @spuetz has updated the [mbf wiki](http://wiki.ros.org/move_base_flex) with more details about the architecture, in case you are interested.

corot:
Strategy change: we have created an alternative to nav_core named [move_base_flex_core](https://github.com/magazino/move_base_flex/tree/master/move_base_flex_core) that provides interfaces valid for both old nav_core-based and new MBF-ready plugins.

- Current plugins are wrapped so they can run within MBF
- New plugins extend move_base_flex_core base classes to profit of MBF extended interface

PR #609, opened 2017-08-05
switch fake_localization to tf2.

vrabaud:


guohefu:
> _No description provided._
Hi, vrabaud
When I was using the fake localization(kinetic-devel 1.14.4)， I found that the rviz 2D pose estimate was not working.  rostopic echo /initialpose  can receive the data，but the coordinates of the robot can't change.

DLu:
@guohefu Please do not cross-post your questions. #806

PR #613, opened 2017-08-10
Use occupancy values when saving a map

allenh1:
This is to allow cartographer's maps to be saved.

The occupancy limits are comments within the source, but they are not enforced by the logic they are next to.

### DIFF
--- a/map_server/src/map_saver.cpp
+++ b/map_server/src/map_saver.cpp
@@ -73,11 +73,11 @@ class MapGenerator
       for(unsigned int y = 0; y < map->info.height; y++) {
         for(unsigned int x = 0; x < map->info.width; x++) {
           unsigned int i = x + (map->info.height - y - 1) * map->info.width;
-          if (map->data[i] == 0) { //occ [0,0.1)
+          if (map->data[i] >= 0 && map->data[i] < 25) { //occ [0,0.25)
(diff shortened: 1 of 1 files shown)


tfoote (approved):

BrannonKing:
I think you probably want 35 ticks on the low end as well; make it symmetric. I'd go `< 35` for line 76.

mikeferguson:
Has anyone tested this against gmapping and karto? How different does the map look?

allenh1:
> I think you probably want 35 ticks on the low end as well; make it symmetric. I'd go < 35 for line 76.

I've not tested this, but I don't know how that would end up. I'll test with 35 later.

> Has anyone tested this against gmapping and karto? How different does the map look?

The maps are not quite what I'm used to for cartographer, but I'm not seeing much of a difference at all from gmapping. In fact, this makes sense if you look at how GMapping publishes its maps.

https://github.com/ros-perception/slam_gmapping/blob/hydro-devel/gmapping/src/slam_gmapping.cpp#L742-L760

allenh1:
Update: maps seem fine with GMapping as well. No noticeable difference.

Should I bumpt the lower bound to be < 35 as suggested above?

allenh1:
@mikeferguson ping

PR #619, opened 2017-08-23
Adding "cutoff distance" as param to Obstacle Layer

ipa-fke:
Priviously laser scan measurements with a distances value greater than the sensors max range are always excluded from the costmap clearing. A parameter let's the user decide whether distance measurements exceeding the max range can be used for clearing. Therefore the `cutoff_distance` parameter has to be set to a threshold distance value corresponding to the limit of trustworthiness. All measurements exceeding this threshold are discarded.

### DIFF
--- a/costmap_2d/include/costmap_2d/observation_buffer.h
+++ b/costmap_2d/include/costmap_2d/observation_buffer.h
@@ -141,6 +141,12 @@ public:
    */
   void resetLastUpdated();
 
+  /**
+   * @brief Used to gain access to the raytrace range
+   * @return the raytrace_range_ attribute
+   */
+  double getRaytraceRange();
+
 private:
   /**
    * @brief  Removes any stale observations from the buffer list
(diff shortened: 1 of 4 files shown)


DLu:
In practice, how is this different than `obstacle_range/raytrace_range`?

ipa-fke:
The `obstacle_range/raytrace_range` only effects the clearing behavior, if  `obstacle_range/raytrace_range < max_range`. All measured points which are in beyond `max_range` are allways excluded, doesn't matter which value for `raytrace_range` is set.

But because the functionality of these params is nearly the same, I see the point that an additional check on `raytrace_range` can solve the problem as well instead of using a new param `cutoff_distance`.

ipa-fke:
@DLu ping

DLu:
Just so I'm clear, I believe you only want this to affect ray tracing, and right now the main code works fine assuming that `raytrace_range` is less than `scan.range_max`.

The end behavior that you want is if the length of a ray is greater than `scan.max_range`, it should clear out to that range. Is that correct? 

If so, I speculate that you might be able to get the behavior you want playing with the `inf_is_valid` flag, but I'm not positive.

mikeferguson:
So, here's my thinking: if you want to override range_max -- then your range_max is probably set wrong. That sounds like an issue for your laser driver, not the navigation stack. This change also contradicts [REP-117](http://www.ros.org/reps/rep-0117.html).

PR #621, opened 2017-08-25
Reconfigure min obstacle height

AlexReimann:
Add reconfigure for min obstacle height for obstacle layer. Max was already there.


AlexReimann:
Oops, wrong repo sorry~~~

PR #622, opened 2017-08-31
Better fix for #523 respecting planner_frequency intended behavior

corot:
Essentially I'm reverting 679ee2baa9af78b202e3d49bc3003c98ca6b74b9 but adding this line: https://github.com/ros-planning/navigation/compare/kinetic-devel...corot:better_fix_for_523?expand=1#diff-d6afbf7b921c7c246a6ee2454dc05be4R609

Now seems to work as expected: the new parameter max_retries does as expected, the planner stops while recovering and keeps replanning while moving if planner_frequency > 0

Not thoroughly tested, so please, give it a good try!

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -562,62 +562,55 @@ namespace move_base {
       }
       ros::Time start_time = ros::Time::now();
 
-      if(state_==PLANNING){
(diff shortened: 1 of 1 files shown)


DLu:
LGTM, in my use case. Thanks for the quick fix.

PR #628, opened 2017-10-09
Only do a getRobotPose when no start pose is given

mintar:
Omit the unnecessary call to getRobotPose when the start pose was
already given, so that move_base can also generate a path in
situations where getRobotPose would fail.

This is actually to work around an issue of getRobotPose randomly
failing.


DLu:
Thanks!

PR #631, opened 2017-10-27
Print SDL error on IMG_Load failure in server_map

Alabate:
When `IMG_Load` from SDL fail to load an image, the node stop without any message because a `std::runtime_error` is thrown without being catch.

This PR expose the SDL Error in the exception and then catch it properly, to print it as a `ROS_ERROR`.

For instance, if the image is too big, we get this log before the node stop:

```
[ INFO] [1509093324.779196221]: Loading map from image "map.png"
[ERROR] [1509093324.779879685]: failed to open image file "map.png": Out of memory
```


DLu:
Thanks!

PR #632, opened 2017-11-01
Fix bounds bug in Costmap2D

samkhal:
Ostensibly, Costmap2D::worldToMapEnforceBounds takes an x and y in world coordinates and converts it to valid cell coordinates, clamping to the map bounds if the input point is outside the map. However, the upper bound has a bug: given an map of `resolution_ = 1, size_x_ = 10, origin_x_ = 0`, an input of `wx=9.5` produces `mx=9`,  an input of `wx=10.5` produces `mx=9`, but an input of `wx=10` produces `mx=10`, an invalid index. This PR fixes that.


DLu:
Can you please add a test for this?

DLu:
Replaced by #871

PR #633, opened 2017-11-07
Feature/zero weight obstacle

dantwinkler:


DLu:
This PR seems to be on the wrong branch, and also is still being worked on. Please reopen against a different branch when you have finished work.

PR #641, opened 2017-11-30
Obstacle layer name is obstacle_layer

LeroyR:
changed clearable_layers_default to obstacle_layer as this is the name used by move_base costmaps

https://github.com/ros-planning/navigation/blob/kinetic-devel/costmap_2d/src/costmap_2d_ros.cpp#L216

### DIFF
--- a/clear_costmap_recovery/src/clear_costmap_recovery.cpp
+++ b/clear_costmap_recovery/src/clear_costmap_recovery.cpp
@@ -61,7 +61,7 @@ void ClearCostmapRecovery::initialize(std::string name, tf::TransformListener* t
     private_nh.param("reset_distance", reset_distance_, 3.0);
     
     std::vector<std::string> clearable_layers_default, clearable_layers;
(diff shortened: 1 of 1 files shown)


DLu:
Thanks for finding this bug. I haven't tested it myself yet, but I'd rather fix this by changing the backwards-compatible name in costmaps than changing the recovery behavior, since other people may rely on the default configuration of the recovery behavior. What do you think?

LeroyR:
I don't know how many people use either of the current configuration. Could be that other people interact with the obstacle_layer with different plugins or other custom applications.

How about adding both layers to the clearing?

DLu:
I proposed an alternative here: https://github.com/ros-planning/navigation/pull/1005

LeroyR:
#1005 supersedes this, closing this request.

*insert rant about fragmentation of forks after 3 year merge window*

😄

DLu:
The rant would be well deserved.

PR #642, opened 2017-11-30
dont add unreacheable target pose

LeroyR:
we are already in tolerance region.
unnecessary burden to local planner and possible unsafe to reach region

This can result in the robot getting stuck for additional time (on blocked goal)

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -395,7 +395,7 @@ namespace move_base {
 
                     //adding the (unreachable) original goal to the end of the global plan, in case the local planner can get you there
                     //(the reachable goal should have been added by the global planner)
-                    global_plan.push_back(req.goal);
+                    //global_plan.push_back(req.goal);
 
                     found_legal = true;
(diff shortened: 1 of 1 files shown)


DLu:
This is a fundamental change to how move_base works that is not desirable in all situations. If you would like to merge this into this repo, I'd recommend hiding it by a flag/parameter.

As it stands, this might result in some unstable behavior, when `found_legal` is set to true. It would probably be better to skip that whole block of code than just comment out the one line.

LeroyR:
It is a legal goal as in the defined constraints: `req.tolerance`, it should set the found_legal flag

I which situation should the local planner try to drive to unreachable goals if the caller approves a goal by setting the tolerance?

DLu:
If it was localization error that made it so that the original goal was unreachable, and then the local planner moves closer and finds the original goal  reachable, then it would probably be preferable to go to the original goal.

mikeferguson:
I concur with DLu -- if we want to add this, it should be parameter-based, and default to the old behavior. The use of the local planner and the tolerances set there, is probably quite pervasive in dealing with mislocalization and sensor noise errors.

mikeferguson:
I'm going to close this since there has been no further discussion -- if you decide to add a parameter based version, please reopen. Thanks!

PR #646, opened 2017-12-07
Remove Dead Code [Lunar]

DLu:
Following up on #639, this PR removes a few instances of dead code from amcl and navfn.


aaronhoy (approved):
LGTM

PR #656, opened 2018-02-09
fix: clear_costmap_recovery default clear layer name

tyuownu:
obstacles -> obstacle_layer


mikeferguson:
Duplicate of #641 (still under discussion there)

PR #658, opened 2018-02-12
CostmapModel: Make lineCost and pointCost public

luator:
The CostmapModel class of the base_local_planner package contains two methods `lineCost` and `pointCost`. They can be quite useful, however they are currently private, so they cannot be used outside the class.

To avoid the need for reimplementing the exact same functionality, I propose this pull request to simply make them public. Both methods do not change the class instance (I also made them `const` to ensure and emphasise this), so I do not think that making them public can cause any problem.


DLu:
LGTM

luator:
@DLu Thanks for merging. Is there any chance to get this backported to kinetic as well? If yes, what is the procedure for this, can I do anything to help?

DLu:
I don't think this should break anyone's binary build so I'd merge an equivalent PR for kinetic too.

luator:
@DLu Alright. See #660.

PR #659, opened 2018-02-14
fix 'enable' for static_layer with rolling window

jabbenseth:
the "enable" option was useless for rolling window configurations

### DIFF
--- a/costmap_2d/plugins/static_layer.cpp
+++ b/costmap_2d/plugins/static_layer.cpp
@@ -293,6 +293,9 @@ void StaticLayer::updateCosts(costmap_2d::Costmap2D& master_grid, int min_i, int
   if (!map_received_)
     return;
 
+  if (!enabled_)
+    return;
+
   if (!layered_costmap_->isRolling())
   {
     // if not rolling, the layered costmap (master_grid) has same coordinates as this layer


DLu:
Hi, thanks for this contribution.

What behavior were you seeing without this fix?

jabbenseth:
Without the fix I wasn't able to disable the static_layer of our rolling window costmap. Our layers are: static_layer then obstacle_layer and finally inflation_layer. When unchecking the "enable" flag of the static_layer in rqt_reconfigure the static layer was still considered. 

This observation was only made if the costmap was set to "rolling_window". I expected the behavior like with the "static_map: true" setting

mikeferguson:
Yep -- that makes sense -- both of the updateWithX functions in the non-rolling block already have the !enabled part. We must have missed that when writing the rolling window support.

mikeferguson:
@DLu this should probably be back-ported to Indigo as well -- I imagine that the bug exists there too.

jabbenseth:
Yes, we discovered on indigo and checked again for kinetic. Should I open another PR for indigo?

DLu:
Yes, please open PRs for Indigo and Lunar.

PR #663, opened 2018-02-15
Fixes #206 for kinetic

### LINKED ISSUE #206: When using robot_radius, footprint is incorrectly set by costmap2d_ros

DLu:
Version of #343 for Kinetic. Lunar PR pending approval of this one. 

(@commaster90  I can't officially assign you as a reviewer, but please take a quick look)

### DIFF
--- a/costmap_2d/src/footprint.cpp
+++ b/costmap_2d/src/footprint.cpp
@@ -225,15 +225,18 @@ std::vector<geometry_msgs::Point> makeFootprintFromParams(ros::NodeHandle& nh)
       if (makeFootprintFromString(std::string(footprint_xmlrpc), points))
       {
         writeFootprintToParam(nh, points);
+        return points;
       }
     }
     else if (footprint_xmlrpc.getType() == XmlRpc::XmlRpcValue::TypeArray)
     {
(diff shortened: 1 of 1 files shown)


patrickcjh (commented):
You still need the rest of the changes from #343. Otherwise it will enter the first "if" since the `footprint` parameter was there, and therefore skipping the `radius` parameter.

DLu:
Oh, good point. That look better?

mikeferguson:
LGTM

patrickcjh:
:+1:

vpradeep07:
Thanks for the fix!  I can confirm that upgrading the navigation stack from version `1.14.2` to `1.14.3` resolved the bug mentioned here.

Issue #634, opened 2017-11-14
AMCL: Assertion 'cidx >= 0' failed


mfe7:
I'm using AMCL on Ubuntu 16.04 with Kinetic. The error that often comes up is:
`amcl: .../amcl/src/amcl/pf/pf.c:525: pf_cluster_stats: Assertion `cidx >= 0' failed.`

This sometimes occurs after I send a new initialpose estimate after driving for a few seconds, but I haven't yet found a consistent procedure that causes it to happen.

I have tried building from source and downloading the package with apt, and see the same behavior both ways. I didn't see this error on my robot with Ubuntu 14.04 with Indigo, but I also have changed the type of laserscanner.

Looking at the source code, this error seems to mean there are no particles - any suggestions of where to look for the cause?

mfe7:
I built from source with the indigo-devel branch and haven't seen the error (still on 16.04)

mikeferguson:
Given that new info, there are only two possibly relevant commits in Kinetic that were not in indigo:

 * https://github.com/ros-planning/navigation/commit/65261a83e7269906f726f14d7e8cc30e5f9901f0#diff-f9c8b2ff23a212af56d0b3de46fcf48b
 * https://github.com/ros-planning/navigation/commit/232cfd232068e48ea22ea7df4afbb07bd0f2d8e5#diff-f9c8b2ff23a212af56d0b3de46fcf48b

If I had to guess, I'd think it was the latter? Perhaps try reverting that commit on the kinetic-devel branch and see if your problem goes away? If so, we can look deeper into why / how to fix it.

tutorgaming:
I also have this issue. (ros-kinetic)  when trying to set the initial pose from rviz, amcl always crashed.

SteveMacenski:
I have also experienced this issue. I see it happen when using `robot_localization` frequently but for some reason doesn't rear its head using `robot_pose_ekf` or not that I'm able to notice in terms of a crash.

DLu:
@Tutorgaming @SteveMacenski If you can provide replicable steps, it will be easier to debug.

SteveMacenski:
I'm generally unable to trigger it on demand. We don't recall seeing this issue in Indigo either.

I agree the second commit is probably the issue. I would test to see if reverting that resolves the issue but the behavior is sporadic for me. I'll have to wait until next time it comes up to test. If @Tutorgaming could test if it happens more consistently in your application, that would be awesome.

JinglinZhang:
[...]

PR #666, opened 2018-02-16
removing recomputation of cluster stats causing assertion error (#634)

SteveMacenski:
Related to ticket #634


mikeferguson:
There is significant interest in getting this merged and released. Unfortunately, it will revert #599 (which is a forward port of #403). Obviously, bad stats is a better issue to have than amcl asserting to death, so I'm inclined to merge this.

If anyone has a better solution, please speak up before Monday -- when I will merge this and cut a new release for Kinetic.

boilerbots:
Looks like they are making a release for Kinetic starting yesterday, can this make it in? You need to tag the latest for Kinetic it seems?

Issue #662, opened 2018-02-15
Problem with Voxel plugin: clearing logic seems to be wrong


facontidavide:
Hi,
 something pretty weird happened to me when I try to use the Voxel Layer.
To give you an idea, please take a look to this video:

https://vimeo.com/255893590

As you can see, a moving object (in this case, two cars) are moving in front of the robot and, as expected, they are detected and the local costmap is updated accordingly.

What is "wrong" is that the voxel grid __is NOT cleaned up__ when the cars disappear. Needless to say that this prevents the local/global planner to work correctly.

After some debugging and testing I am pretty sure that I found the problem. Let me illustrate it using the following scientific representation:

![image](https://user-images.githubusercontent.com/2822888/36257952-3d77f15e-1259-11e8-9cc0-ddce70b5fa97.png)

The max_obstacle_height is set to 2.0 meters. All the points above this threshold are discarded.
A moving __obstacle is detected below the threshold__, but once the obstacle disappears, the ray will hit a wall at a Z coordinate that is __above the threshold__. The ray is discarded and is not used for cleaning the VoxelGrid.  

Note that the same is true for min_obstacle_height. We don't want to discard clearing rays that hit the ground!

In short: the min/max obstacle layers should be applied only to marking, but we should use ALL the rays for clearing, no matter their height.

facontidavide:
For your information. using a value of __z_voxels__ smaller than __max_obstacle_height / z_resolution__ makes this problem even worse.
I am trying to find the reason in the code, may be you can figure it out faster than me ;)

UPDATE:
In other words, if **(z_voxels < max_obstacle_height / z_resolution)**, it is more probable that some voxels are marked but NOT cleared when they could.

The problems seems to be related to these two checks that fails:

https://github.com/ros-planning/navigation/blob/kinetic-devel/costmap_2d/plugins/voxel_layer.cpp#L353
https://github.com/ros-planning/navigation/blob/kinetic-devel/voxel_grid/src/voxel_grid.cpp#L133

facontidavide:
FYI. I have a fork where this problem is considerably mitigated and the API was NOT modified.

It is not ready for a PR, but if you want to discuss the potential solution to this problem, it might me a starting point.

Cheers

mikeferguson:
The extensive discussion in https://github.com/ros-planning/navigation/issues/267 might also be of interest to you -- it's a slightly different issue, but given your type of sensor, might be something you'll see.

facontidavide:
[...]

PR #667, opened 2018-02-16
Fixing issue in Voxel clearing (issue #662)

facontidavide:
Hi,

this commit https://github.com/ros-planning/navigation/commit/435a5a762036958e099bf5873b4bbccf82e39660 should solve the problem reported.

The idea is simple: use ALL the valid rays to clear.

The second commit is purely a pedantic "code simplification" that you can just ignore if you want.

Clearing points were discarded because either their height Z are outside the marking limits or because the Z coordinate of the point exceeded the size_z_ of the grid.

You can compare the results before:

https://vimeo.com/255893590

and after the changes

https://vimeo.com/256061162

You will noticed that there are still few voxels which are not cleared. The reason is that some rays of the velodyne are not hitting any object (they are pointing at the sky), therefore they are discarded by the velodyne_pointcloud driver.
In other words, PointCloud and VoxelLayer do not provide the equivalent of "infinite is valid", for this reason there are still marked voxel that can not be cleared.

### DIFF
--- a/costmap_2d/include/costmap_2d/observation.h
+++ b/costmap_2d/include/costmap_2d/observation.h
@@ -80,7 +80,8 @@ public:
    */
   Observation(const Observation& obs) :
       origin_(obs.origin_), cloud_(new pcl::PointCloud<pcl::PointXYZ>(*(obs.cloud_))),
-      obstacle_range_(obs.obstacle_range_), raytrace_range_(obs.raytrace_range_)
+      obstacle_range_(obs.obstacle_range_), raytrace_range_(obs.raytrace_range_),
+      within_range_(obs.within_range_)
   {
   }
 
@@ -97,6 +98,7 @@ public:
   geometry_msgs::Point origin_;
   pcl::PointCloud<pcl::PointXYZ>* cloud_;
   double obstacle_range_, raytrace_range_;
+  std::vector<bool> within_range_;
 };
 
 }  // namespace costmap_2d
--- a/costmap_2d/plugins/voxel_layer.cpp
+++ b/costmap_2d/plugins/voxel_layer.cpp
@@ -146,13 +146,14 @@ void VoxelLayer::updateBounds(double robot_x, double robot_y, double robot_yaw, 
     const Observation& obs = *it;
 
     const pcl::PointCloud<pcl::PointXYZ>& cloud = *(obs.cloud_);
+    const std::vector<bool>& in_range = obs.within_range_;
 
     double sq_obstacle_range = obs.obstacle_range_ * obs.obstacle_range_;
 
     for (unsigned int i = 0; i < cloud.points.size(); ++i)
     {
       // if the obstacle is too high or too far away from the robot we won't add it
-      if (cloud.points[i].z > max_obstacle_height_)
+      if (in_range[i] == false || cloud.points[i].z > max_obstacle_height_)
         continue;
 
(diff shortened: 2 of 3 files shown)


mikeferguson:
Davide, I've [commented in the issue](https://github.com/ros-planning/navigation/issues/662#issuecomment-376565069) that I think this can be accomplished through the configuration parameters available today. If that is the case, that is preferable, since it has less chance of possibly breaking other user's configurations. I'm going to add a few notes here and close this pull request to let dialogue continue in the issue. We can reopen if it turns out we really do need to modify things. Thanks!

costmap_2d/include/costmap_2d/observation.h
```diff
@@ -97,6 +98,7 @@ class Observation
   geometry_msgs::Point origin_;
   pcl::PointCloud<pcl::PointXYZ>* cloud_;
   double obstacle_range_, raytrace_range_;
+  std::vector<bool> within_range_;
```
mikeferguson, line 101: rather than creating a possibly very large bool array, I would suggest that we should instead pass the min_obstacle_height_ and max_obstacle_height_ into the Observation from the ObservationBuffer (as we already do for raytrace_range_ and obstacle_range_)

costmap_2d/plugins/voxel_layer.cpp
```diff
@@ -146,32 +146,34 @@ void VoxelLayer::updateBounds(double robot_x, double robot_y, double robot_yaw,
...
+    const std::vector<bool>& in_range = obs.within_range_;
 
     double sq_obstacle_range = obs.obstacle_range_ * obs.obstacle_range_;
 
     for (unsigned int i = 0; i < cloud.points.size(); ++i)
     {
+      const pcl::PointXYZ& point = cloud.points[i];
       // if the obstacle is too high or too far away from the robot we won't add it
-      if (cloud.points[i].z > max_obstacle_height_)
+      if (in_range[i] == false || point.z > max_obstacle_height_ )
```
mikeferguson, line 157: this could then be updated to compare against the obs.max_obstacle_height_

costmap_2d/plugins/voxel_layer.cpp
```diff
@@ -313,10 +315,11 @@ void VoxelLayer::raytraceFreespace(const Observation& clearing_observation, doub
     double t = 1.0;
 
     // we can only raytrace to a maximum z height
-    if (wpz > max_obstacle_height_)
+    const double max_z = std::min(max_obstacle_height_, size_z_ * z_resolution_ );
```
mikeferguson, line 318: I think we need to look closer at this specific item, as mentioned in the issue.

costmap_2d/include/costmap_2d/observation.h
facontidavide: I agree, we can do that

costmap_2d/plugins/voxel_layer.cpp
facontidavide: From my point of view, either we apply this change or we simply remove the parameter __z_voxels__, since this value is supposed to be (max_obstacle_height / z_resolution)

facontidavide:
Let's agree that we disagree :)

I guess that the best way to "convince you" that these changes are necessary might be to build a unit test that show clearly the behavior of the VoxelGrid before and after the PR.

Does it make sense?

PR #668, opened 2018-02-23
Update laser_model_type enum on AMCL.cfg

B0gdar:
Adding likelihood_field_prob laser model option on AMCL.cfg to be able to control dynamic parameters with this laser sensor model.

Likely to affect all branches that implement this laser_sensor_model.


mikeferguson:
Normally, changes to dynamic reconfigure would bust ABI -- but this particular one appears not to, since we're only adding another named option.

PR #674, opened 2018-03-05
[NOT FOR MERGE] omg-local-planner implementation

lounick:
Initial commit for review and giving some reference for the development. It can't be finished until the services are defined as messages in the omg ros bridge.


lounick:
That was a mistake. Has to be finished internally first.

PR #680, opened 2018-03-15
Fixed map import & particle filter bugs and augmented with some debug info & utility for AMCL

alicegong:
- Added to check orientation included in map_msg.info to rotate map upon import (originally assume to discard this info!)
- Dissected its source code to locate procedures _not_ implemented as proposed in _"Probabilistic Robotics"_,
- Rewrote its adaptive sampling procedures to appropriately adjust sample size according to state uncertainty,
- Replaced its uniform re-sampling procedures with weighted re-sampling, enabling consistent convergence,
- Fixed its problematic bias in calculation of particle posterior probability against unknown areas on map, 
- Modified sensor measurement model - behaviour during in-place rotation,
- Implemented an augmentation on an another thread to listen to odometry & extrapolate between pose update using odometry (uses/requires MRPT, turn it on with cmake option BUILD_EXTRAPOLATION)
- Improved estimation accuracy and consistency by at least 5x, validated error margins to be within 10cm for over 10m travelled in real environments 🎆  .

Note: Commits only include fixing CI build ATM, b/c all core changes were made during time on another project.

CI build has passed; Currently fixing failed localization threshold tests (huh)


mikeferguson:
A 5x improvement in localization accuracy would of course be a welcome improvement, however, this obviously can't go very far until it passes build/test. Even then, this is such a huge change, I'd suggest it probably shouldn't go into Kinetic at this point (since it is an LTS), and should instead target Lunar or later.

A few comments on what would make this actually able to review/merge:
 * Many files have had ONLY their mode changed (100644 -> 100755), which should be reverted (no need to be executable). Similarly, changelogs should not be affected, and should be reverted.
 * Ideally, the commit history would break out the (seemingly trivial) build system updates into it's own commit (here, I'm talking about the changes in whitespace, renaming of how folders are included, etc)
 * Then, I would suggest having a commit that does any pure refactor which is not part of a specific change, and which is not intended to have any impact on actual function/performance.
 * Then, each specific change should be it's own commit. So that would be a commit for pretty much each of your bullet points above. This way, they can be reviewed individually, without trying to figure out which lines correspond which sets of changes.

I realize that such a refactor of your commit history is significant work -- but realistically, it's impossible for anyone to review, in a single pass, 4600 line changes on a code base that is over 20 years old and in active use on several thousand robots worldwide.

alicegong:
I agree it's significant work, but I'd like to share my changes if it can be beneficial to, as you said, several thousand robots worldwide. 

I am not sure why git included files with only mode changes... I'll revert.

Many lines of changes were just whitespace changes - I think many files don't follow a 4 space = 1 soft tab convention, and I struggle to find a common convention every file follow - so please advise me on what to do here.

So according to the feedback you provided 👍 , I propose the following commits:
- [ ] purely sync up the indentations...
- [ ] pure refactor (into functions/class/etc)
- [ ] more refactor (the ones that were refactor into different files)
- [ ] build system update (may be combined with above if not too much?)
- [ ] each specific change in bullet points I listed (in as much isolation as possible)

Please do let me know if you and/or the team don't currently have the bandwidth to do this though, so I'll plan this with my other priorities accordingly. 

If I get a go-ahead, we'll close this pull request, and I'll open a new one with another target branch? / I could really use the CI tests before we close this though. (CI build is all fixed now)
But Lunar is supported till 2019 (vs Kinectic, which is 2021, later than Lunar)? Is it because we should try to incorporate such significant changes to a newer but more experimental version?

mikeferguson:
That set of commits seems good - if you can take the time to split things up in that order, I will take the time to review it. I will note that such a PR will take at least two weeks of open comment -- hopefully some folks will step forward to test on their robots during that time as well.

Kinetic is an LTS, which is why it as a longer EOL -- but it's also been "allegedly stable" for over 2 years, which is why I'm not keen to put big changes into it. Sometime later this year (probably May/June?) we will fork off from Lunar for the new Melodic release, which is the next LTS (with support until 2023). I know of several organizations who skipped Kinetic for their real robots, and are still operating on Indigo (which is EOL in 2019), so I suspect that Melodic will get a lot of real robot use in it's first year.

alicegong:
OK sounds good! I'll gradually work on it - the ambitious goal would be to finish it before end of school year at the end of April (I'm graduating), but otherwise hopefully before end of May to have enough time to test with the community before forking off from Lunar for the new Melodic release.

>  at least two weeks of open comment -- hopefully some folks will step forward to test on their robots during that time as well

I'd love that!

alicegong:
Couldn't change branches here so this has been migrated to #716 and is currently WIP so this PR has been closed

PR #681, opened 2018-03-15
Make sure to call setFootprint() before an in-place rotation

### LINKED ISSUE #327: ObstacleCostFunction has an empty footprint when the first goal is an in-place rotation

gerkey:
This PR should fix #327, which we ran into on a project recently. I'm sure that there are a variety of ways to address the issue, but this set of changes seems to me the least intrusive way to ensure that `setFootprint()` is called before what might be an in-place rotation without any prior translation.

### DIFF
--- a/dwa_local_planner/include/dwa_local_planner/dwa_planner.h
+++ b/dwa_local_planner/include/dwa_local_planner/dwa_planner.h
@@ -118,7 +118,8 @@ namespace dwa_local_planner {
        * @param  new_plan The new global plan
        */
       void updatePlanAndLocalCosts(tf::Stamped<tf::Pose> global_pose,
-          const std::vector<geometry_msgs::PoseStamped>& new_plan);
+          const std::vector<geometry_msgs::PoseStamped>& new_plan,
+          std::vector<geometry_msgs::Point> footprint_spec);
 
       /**
        * @brief Get the period at which the local planner is expected to run
--- a/dwa_local_planner/src/dwa_planner.cpp
+++ b/dwa_local_planner/src/dwa_planner.cpp
@@ -242,11 +242,14 @@ namespace dwa_local_planner {
 
   void DWAPlanner::updatePlanAndLocalCosts(
       tf::Stamped<tf::Pose> global_pose,
-      const std::vector<geometry_msgs::PoseStamped>& new_plan) {
+      const std::vector<geometry_msgs::PoseStamped>& new_plan,
+      std::vector<geometry_msgs::Point> footprint_spec) {
     global_plan_.resize(new_plan.size());
     for (unsigned int i = 0; i < new_plan.size(); ++i) {
       global_plan_[i] = new_plan[i];
     }
+
+    obstacle_costs_.setFootprint(footprint_spec);
 
     // costs for going away from path
     path_costs_.setTargetPoses(global_plan_);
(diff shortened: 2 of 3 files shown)


mikeferguson:
Seems reasonable to me -- I'm going to leave it for a day or two so @DLu can possibly comment since I think he is more familiar with these local planners.

gerkey:
Hmm, now that I've looked at the code again, I should probably make `footprint_spec` a const reference when passed into `updatePlanAndLocalCosts()`. If you guys think that this overall approach makes sense, then I'll also make and test that change.

pavloblindnology:
Why not to remove footprint setting from `DWAPlanner::findBestPath` additionally? It's not needed anymore after your changes.
https://github.com/ros-planning/navigation/blob/956c2c9e6fddce5484343cfc97dbc4e7217edcc2/dwa_local_planner/src/dwa_planner.cpp#L295

DLu (changes requested):
I concur with the two points made thus far.
 * Should be a const ref
 * The footprint is not needed in findBestPath

I've made these changes in [a branch here](https://github.com/gerkey/navigation/compare/set_footprint_before_in_place_rotation...DLu:set_footprint_before_rotation?expand=1) that you can merge in if you'd like.

enwpoa332:
Not sure if this solves the problem. 

In `DWAPlannerROS::computeVelocityCommands()`, if the first nav goal is on top of `current_pose_`, then `latchedStopRotateController_.isPositionReached()` returns `true`, and `DWAPlanner::checkTrajectory()` is called. This leads to `ObstacleCostFunction::scoreTrajectory()`, which depends on `ObstacleCostFunction::setFootprint()` having set the `ObstacleCostFunction::footprint_spec_` member. With a const reference instead of the `setFootprint()`, the `footprint_spec_.size()` will still return zero.

DLu:
Closed in favor of #829

PR #682, opened 2018-03-16
Fix minor typo

mikeferguson:
This typo caused some confusion because we were searching for a semicolon in our configuration.


mikeferguson:
Forward port of #678

Issue #679, opened 2018-03-15
Control loop missing its desired rate (waitForTransform related).


pavloblindnology:
The PR #270 described in #196 severely slows down the control loop speed (similar problem was described in #468).
The reason is that waitForTransform is called multiple times during a control loop. In `dwa_local_planner` it's called 3 times:
1. https://github.com/ros-planning/navigation/blob/51dfabcb4d6deee5a63329e1e6dfab6ae563d47d/dwa_local_planner/src/dwa_planner_ros.cpp#L153
2. https://github.com/ros-planning/navigation/blob/51dfabcb4d6deee5a63329e1e6dfab6ae563d47d/dwa_local_planner/src/dwa_planner_ros.cpp#L259
3. https://github.com/ros-planning/navigation/blob/51dfabcb4d6deee5a63329e1e6dfab6ae563d47d/dwa_local_planner/src/dwa_planner_ros.cpp#L274

The calls leads to either
https://github.com/ros-planning/navigation/blob/51dfabcb4d6deee5a63329e1e6dfab6ae563d47d/base_local_planner/src/goal_functions.cpp#L88
or 
https://github.com/ros-planning/navigation/blob/51dfabcb4d6deee5a63329e1e6dfab6ae563d47d/base_local_planner/src/goal_functions.cpp#L176
both of which use 'waitForTransform' with `ros::Time::now()`.

Calling `waitForTransform` with `ros::Time::now()` several times in a row in a single thread results in their "serialization" - each of them locks for the next TF update time step `dt`, till the corresponding TF transform is available again - i.e. the final lock time is `~N*dt` for `N` runs in a row.
One possible solution to this is to use single time for all `waitForTransform` calls, e.g. by keeping it in `LocalPlannerUtil` and updating it in `isGoalReached` method of local planner.

P.S. This issue arises mostly when using global localization package which doesn't "predict" into the future (like gmapping) - e.g. hector_slam.

rhklite:
i saw a PR #688  for kinetic. I'm also seeing this issue described in #196  and #468 in indigo. Is this issue present in indigo aswell?

pavloblindnology:
@rhklite Well, if you say you're seeing it in Indigo than probably it's present in Indigo.

HappySamuel:
Hi All

Is this issue still exist in ROS Noetic?

Best,
Samuel

pavloblindnology:
Hi,
[...]

PR #688, opened 2018-03-26
Slow control loop with dwa_local_planner (fix for #679)

pavloblindnology:
Fix for #679

### DIFF
--- a/base_local_planner/include/base_local_planner/goal_functions.h
+++ b/base_local_planner/include/base_local_planner/goal_functions.h
@@ -116,10 +116,10 @@ namespace base_local_planner {
      * @return True if achieved, false otherwise
      */
   bool getGoalPose(const tf::TransformListener& tf,
-  		  const std::vector<geometry_msgs::PoseStamped>& global_plan,
-  		  const std::string& global_frame,
-        const ros::Time& time,
-  		  tf::Stamped<tf::Pose> &goal_pose);
+      const std::vector<geometry_msgs::PoseStamped>& global_plan,
+      const std::string& global_frame,
+      const ros::Time& time,
+      tf::Stamped<tf::Pose> &goal_pose);
 
   /**
    * @brief  Check if the goal pose has been achieved
--- a/base_local_planner/include/base_local_planner/goal_functions.h
+++ b/base_local_planner/include/base_local_planner/goal_functions.h
@@ -104,6 +104,7 @@ namespace base_local_planner {
       const tf::Stamped<tf::Pose>& global_robot_pose,
       const costmap_2d::Costmap2D& costmap,
       const std::string& global_frame,
+      const ros::Time& time,
       std::vector<geometry_msgs::PoseStamped>& transformed_plan);
 
   /**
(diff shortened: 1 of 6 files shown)


mikeferguson:
CI didn't trigger initially -- now running

base_local_planner/src/goal_functions.cpp
```diff
@@ -103,12 +104,8 @@ namespace base_local_planner {
     try {
       // get plan_to_global_transform from plan frame to global_frame
       tf::StampedTransform plan_to_global_transform;
-      tf.waitForTransform(global_frame, ros::Time::now(),
-                          plan_pose.header.frame_id, plan_pose.header.stamp,
-                          plan_pose.header.frame_id, ros::Duration(0.5));
-      tf.lookupTransform(global_frame, ros::Time(),
```
mikeferguson, line 109: I wonder if would be desirable to keep this as ros::Time() which will get the latest available transform (rather than a possibly outdated one). Although, it might also be that having all transforms done at the same time is better for consistency.

base_local_planner/include/base_local_planner/goal_functions.h
```diff
@@ -117,6 +118,7 @@ namespace base_local_planner {
   bool getGoalPose(const tf::TransformListener& tf,
   		  const std::vector<geometry_msgs::PoseStamped>& global_plan,
   		  const std::string& global_frame,
+        const ros::Time& time,
```
mikeferguson: The indentation here is off -- but it appears that is because you properly used spaces, but the lines above/below use some tabs.

dwa_local_planner/src/dwa_planner_ros.cpp
```diff
@@ -251,10 +260,6 @@ namespace dwa_local_planner {
 
   bool DWAPlannerROS::computeVelocityCommands(geometry_msgs::Twist& cmd_vel) {
     // dispatches to either dwa sampling control or stop and rotate control, depending on whether we have been close enough to goal
-    if ( ! costmap_ros_->getRobotPose(current_pose_)) {
-      ROS_ERROR("Could not get robot pose");
-      return false;
-    }
```
mikeferguson, line 259: why is this removed?

base_local_planner/include/base_local_planner/goal_functions.h
pavloblindnology: Fixed.

base_local_planner/src/goal_functions.cpp
pavloblindnology: We cannot just replace `ros::Time::now()` with `ros::Time()` in line
https://github.com/ros-planning/navigation/blob/43e9297d588276efa9868efe2ed56ac6bf08e180/base_local_planner/src/goal_functions.cpp#L106
as transform in line
https://github.com/ros-planning/navigation/blob/43e9297d588276efa9868efe2ed56ac6bf08e180/base_local_planner/src/goal_functions.cpp#L115
will use possibly later timestamp of current robot pose, issuing TF error 'extrapolation into the future', mentioned here #196

dwa_local_planner/src/dwa_planner_ros.cpp
pavloblindnology: It's explained in line 157 of my dwa_planner_ros.cpp.
If we update robot pose after setting time in planner_util we can get a more recent time stamp of robot pose, which will cause TF 'extrapolation into the future` error in line
https://github.com/ros-planning/navigation/blob/43e9297d588276efa9868efe2ed56ac6bf08e180/base_local_planner/src/goal_functions.cpp#L115

pavloblindnology:
Any update on this?

mikeferguson:

PR #698, opened 2018-03-28
Make trajectory scoring scales consistent (fixes #689))

### LINKED ISSUE #689: Trajectory scoring scales inconsistency
According to [this](http://wiki.ros.org/base_local_planner?distro=lunar#Trajectory_Scoring_Parameters) and [this](http://wiki.ros.org/dwa_local_planner?distro=lunar#Trajectory_Scoring_Parameters), the trajectory obstacle scoring is computed as
`occdist_scale * (maximum obstacle cost along the trajectory in obstacle cost (0-254))`
While in the code, `occdist_scale` is multiplied by grid resolution:

- https://github.com/ros-planning/navigation/blob/df82021fabc6ebdf0d01b86f3082e33dd186f6b9/base_local_planner/src/trajectory_planner.cpp#L89
- https://github.com/ros-planning/navigation/blob/df82021fabc6ebdf0d01b86f3082e33dd186f6b9/dwa_local_planner/src/dwa_planner.cpp#L75

While meter scoring is perfectly appropriate for goal/path distance scoring - imho, it looks very strange and inappropriate for obstacle scoring to depend on map resolution. Bcz, I want to match, say, obstacle cost value 100 to 1 meter distance to path. For this I need `0.01(occdist_scale)*100(cost)` and `1(path_distance_bias)*1(dist_to_path_in_meters)`. This looks naturally and easy to understand. Also, default path and goal biases are set unrealistically high (32.0 and 24.0, respectively). With such values you get final scores like:
`obstacle score = 0.05(resolution) * 0.01(occdist_scale) * 100(cost) = 0.05`
`path score = 32.0(path_distance_bias) * 1(dist_to_path_in_meters) * 0.5 = 16.0`
`goal score = 24.0(goal_distance_bias) * 1(dist_to_goal_in_meters) * 0.5 = 12.0`
which makes obstacle score simply negligible!

In `dwa_local_planner`, additionally, there is inconsistency between next lines
`total_cost = pdist_scale_ * resolution * path_cost + gdist_scale_ * resolution * goal_cost + occdist_scale_ * occ_cost;`
- https://github.com/ros-planning/navigation/blob/df82021fabc6ebdf0d01b86f3082e33dd186f6b9/dwa_local_planner/src/dwa_planner.cpp#L200
and 
- https://github.com/ros-planning/navigation/blob/df82021fabc6ebdf0d01b86f3082e33dd186f6b9/dwa_local_planner/src/dwa_planner.cpp#L67
- https://github.com/ros-planning/navigation/blob/df82021fabc6ebdf0d01b86f3082e33dd186f6b9/dwa_local_planner/src/dwa_planner.cpp#L71
- https://github.com/ros-planning/navigation/blob/df82021fabc6ebdf0d01b86f3082e33dd186f6b9/dwa_local_planner/src/dwa_planner.cpp#L75

As we can see, when computing point cloud scores, path and goal scores aren't additionally multiplied by 0.5 and obstacle score isn't multiplied by resolution at all.

pavloblindnology:
Fixes #689

### DIFF
--- a/base_local_planner/src/trajectory_planner.cpp
+++ b/base_local_planner/src/trajectory_planner.cpp
@@ -86,7 +86,6 @@ namespace base_local_planner{
         double resolution = costmap_.getResolution();
         gdist_scale_ *= resolution;
         pdist_scale_ *= resolution;
-        occdist_scale_ *= resolution;
       }
 
       oscillation_reset_dist_ = config.oscillation_reset_dist;
--- a/dwa_local_planner/cfg/DWAPlanner.cfg
+++ b/dwa_local_planner/cfg/DWAPlanner.cfg
@@ -16,8 +16,8 @@ gen.add("sim_time", double_t, 0, "The amount of time to roll trajectories out fo
 gen.add("sim_granularity", double_t, 0, "The granularity with which to check for collisions along each trajectory in meters", 0.025, 0)
 gen.add("angular_sim_granularity", double_t, 0, "The granularity with which to check for collisions for rotations in radians", 0.1, 0)
 
-gen.add("path_distance_bias", double_t, 0, "The weight for the path distance part of the cost function", 32.0, 0.0)
-gen.add("goal_distance_bias", double_t, 0, "The weight for the goal distance part of the cost function", 24.0, 0.0)
+gen.add("path_distance_bias", double_t, 0, "The weight for the path distance part of the cost function", 0.6, 0.0)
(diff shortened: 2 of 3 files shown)


mikeferguson:
@pavloblindnology I'm really hesitant to change defaults within a release -- we just don't know how many robots might be using that default value. Would you object to targeting this at Melodic instead?

pavloblindnology:
@mikeferguson No, I don't mind.

mikeferguson:
https://github.com/ros-planning/navigation/pull/763

PR #699, opened 2018-03-28
Fix trajectory obstacle scoring in dwa_local_planner (fixes #697)

### LINKED ISSUE #697: Invalid trajectory's obstacle scoring in dwa_local_planner
According to next line, trajectory's obstacle score equals to score of the last point in `dwa_local_planner`.
https://github.com/ros-planning/navigation/blob/43e9297d588276efa9868efe2ed56ac6bf08e180/base_local_planner/src/obstacle_cost_function.cpp#L97
While it should be the maximum of scores for all points, as it's done in `base_local_planner`
https://github.com/ros-planning/navigation/blob/43e9297d588276efa9868efe2ed56ac6bf08e180/base_local_planner/src/trajectory_planner.cpp#L306
Otherwise, the planner just doesn't care about obstacles located far enough from the trajectory endpoint (of course if not in collision with them).

pavloblindnology:
Fixes #697

### DIFF
--- a/base_local_planner/src/obstacle_cost_function.cpp
+++ b/base_local_planner/src/obstacle_cost_function.cpp
@@ -94,7 +94,7 @@ double ObstacleCostFunction::scoreTrajectory(Trajectory &traj) {
     if(sum_scores_)
         cost +=  f_cost;
     else
-        cost = f_cost;
+        cost = std::max(cost, f_cost);
   }
   return cost;
 }


mikeferguson:
@pavloblindnology apologies for delay in review. This seems reasonable -- can you think of any tuning someone might have done that would rely on the original value? (Based on your tickets, I can surmise you've spent a lot of time reading this part of the codebase)

pavloblindnology:
@mikeferguson Personally I tried to play with `occdist_scale`, `cost_scaling_factor`, `inflation_radius`, `footprint` and `robot_radius` to make robot pass farther from the obstacles. And it helped in certain situations.

DLu (approved):
You know, I think was initially wary of this changing people's existing tuning, but I think this will solve more problems than it will cause.

mikeferguson:
I'm not so sure about merging into Kinetic -- if I have a tuned robot and things suddenly change -- eesh. For now, I am going to cherry pick this into Melodic for our first release -- and we can make a decision on backport later.

mikeferguson:
https://github.com/ros-planning/navigation/pull/763

PR #700, opened 2018-03-28
Map server wait for a valid time, fix #573

### LINKED ISSUE #573: The map_server keeps the init stamp for the map so 0 in simulation.
The `map_server` uses its launch time for the stamp of the map message ([here](https://github.com/ros-planning/navigation/blob/43e9297d588276efa9868efe2ed56ac6bf08e180/map_server/src/main.cpp#L179), which makes sense).

But in case we use it with Gazebo, there is no `ros::Time::waitForValid()` statement so if they are launched at the same time, the stamp is always 0.

I do not know if it is the expected behavior but I had an issue where I was using some tf transform helpers that use the message stamp for the `lookupTransform`. So in simulation it was 0 which means, "take the latest available transform", while on the real robot it was using the loading time of the `map_server`, so it always get a timeout.

So I was wondering if it would be a good idea to add the `ros::Time::waitForValid()` in the `map_server` constructor?

Edit: fix the link to source code

romainreignier:
When launching the map_server with Gazebo, the current time is picked
before the simulation is started and then has a value of 0.
Later when querying the stamp of the map, a value of has a special
signification on tf transform for example.

### DIFF
--- a/map_server/src/main.cpp
+++ b/map_server/src/main.cpp
@@ -174,6 +174,8 @@ class MapServer
           ROS_ERROR("%s", e.what());
           exit(-1);
       }
+      // To make sure get a consistent time in simulation
+      ros::Time::waitForValid();
       map_resp_.map.info.map_load_time = ros::Time::now();
       map_resp_.map.header.frame_id = frame_id;
       map_resp_.map.header.stamp = ros::Time::now();


mikeferguson:
Retriggered CI (never ran...)

mikeferguson:
Thanks for the contribution! -- LGTM -- once CI completes I will merge.

PR #703, opened 2018-03-29
global_planner: enable default_tolerance with potential field, fix #262

### LINKED ISSUE #262: default_tolerance parameter in global_planner does not working
It seems like default_tolerance parameter is not used in [global planner](https://github.com/ros-planning/navigation/tree/hydro-devel/global_planner) package.

[makePlan function in global planner](https://github.com/ros-planning/navigation/blob/hydro-devel/global_planner/src/planner_core.cpp#L208-L209) receives tolerance but not use anywhere.

nlimpert:
Previously the default_tolerance parameter was not used leading to
failure when the desired goal was inside an obstacle.
In case the goal is inside an obstacle the global_planner now tries to
find a cell with pot_field_max_cost as its maximum value as close as
possible around the desired goal pose. This is achieved by calculating a
potential field in the area of the desired goal pose.
This area is defined by the default_tolerance parameter.

@mikeferguson this should fix #262.
See a demonstration video at https://fh-aachen.sciebo.de/s/SReCNmR5Larlxt4


DLu:
There's quite a lot going on here, but shouldn't the logic be based on `tolerance` instead of `default_tolerance_`?

nlimpert:
Of course you are right - Thanks for the notice!

mikeferguson:
@DLu looks like you started to review this, it's another global_planner one if you have a chance.

global_planner/include/global_planner/planner_core.h
```diff
@@ -170,6 +172,7 @@ class GlobalPlanner : public nav_core::BaseGlobalPlanner {
          * @brief Store a copy of the current costmap in \a costmap.  Called by makePlan.
          */
         costmap_2d::Costmap2D* costmap_;
+        costmap_2d::Costmap2DROS* costmap_ros_;
```
DLu: It doesn't look like you use costmap_ros_ anywhere in this PR.

global_planner/include/global_planner/planner_core.h
```diff
@@ -205,6 +209,15 @@ class GlobalPlanner : public nav_core::BaseGlobalPlanner {
         dynamic_reconfigure::Server<global_planner::GlobalPlannerConfig> *dsrv_;
         void reconfigureCB(global_planner::GlobalPlannerConfig &config, uint32_t level);
 
+        unsigned int pot_field_max_cost_;
+
+        std::vector<float> proj_modes;
```
DLu: Likewise, proj_modes is not used here either.

global_planner/include/global_planner/planner_core.h
```diff
@@ -205,6 +209,15 @@ class GlobalPlanner : public nav_core::BaseGlobalPlanner {
...
+
+        double new_goal_x;
+        double new_goal_y;
+        double last_official_goal_x;
+        double last_official_goal_y;
```
DLu: The standard style in this code is for member variables to end with an underscore.

global_planner/include/global_planner/planner_core.h
```diff
@@ -205,6 +209,15 @@ class GlobalPlanner : public nav_core::BaseGlobalPlanner {
         dynamic_reconfigure::Server<global_planner::GlobalPlannerConfig> *dsrv_;
         void reconfigureCB(global_planner::GlobalPlannerConfig &config, uint32_t level);
 
+        unsigned int pot_field_max_cost_;
+
+        std::vector<float> proj_modes;
+
+        double new_goal_x;
+        double new_goal_y;
```
DLu: Why did you make `new_goal_x` and `y` member variables and not just local variables?

global_planner/src/planner_core.cpp
```diff
@@ -143,6 +144,7 @@ void GlobalPlanner::initialize(std::string name, costmap_2d::Costmap2D* costmap,
         private_nh.param("planner_window_y", planner_window_y_, 0.0);
         private_nh.param("default_tolerance", default_tolerance_, 0.0);
         private_nh.param("publish_scale", publish_scale_, 100);
+        private_nh.param("pot_field_min_", publish_scale_, 100);
```
DLu: A) parameter names should not end with an underscore.
B) It probably shouldn't be setting publish_scale, but I'm not sure what its supposed to set either.

global_planner/src/planner_core.cpp
```diff
@@ -222,6 +227,15 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
...
     boost::mutex::scoped_lock lock(mutex_);
+
+    // reset last_official_goal if it diverges
+    if (fabs(last_official_goal_x - goal.pose.position.x) > 0.1 || fabs(last_official_goal_y - goal.pose.position.y) > 0.1) {
+      last_official_goal_x = goal.pose.position.x;
+      last_official_goal_y = goal.pose.position.y;
+      new_goal_x = goal.pose.position.x;
+      new_goal_y = goal.pose.position.y;
+      ROS_DEBUG("NEW GOAL, setting last goal to: %f %f", new_goal_x, new_goal_y);
+    }
```
DLu: If `initialize` is not called, then `last_official_goal_x` will not be initialized, so either, you should
 * Move the initialization to the constructor's initializers (preferred)
 * or Move this if statement to after the initialized check on the next line.

global_planner/src/planner_core.cpp
```diff
@@ -233,11 +247,12 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
 
     ros::NodeHandle n;
     std::string global_frame = frame_id_;
+    geometry_msgs::PoseStamped goal_ = goal;
```
DLu: Likewise, local variables like `goal_` should NOT end in underscores.

global_planner/include/global_planner/planner_core.h
```diff
@@ -38,6 +38,7 @@
  *         David V. Lu!!
  *********************************************************************/
 #define POT_HIGH 1.0e10        // unassigned cell potential
+#define POT_FIELD_TIMEOUT 1.0  // Timeout of potential field method
```
DLu: This should probably be a ROS parameter instead of a `#define`

global_planner/src/planner_core.cpp
```diff
@@ -247,6 +262,107 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
...
+        if (ros::Time::now() - begin > ros::Duration(POT_FIELD_TIMEOUT)) {
+            ROS_ERROR_NAMED("global planner", "Failed to find a point via potential field "
+                                              "in the map within %f seconds", POT_FIELD_TIMEOUT);
+            break;
+        }
+
+        pot_field_target_phi = angles::normalize_angle(atan2(pot_field_target_y, pot_field_target_x));
+
+        float cur_x = std::cos(pot_field_target_phi) * (costmap_->getResolution() * 0.5);
+        float cur_y = std::sin(pot_field_target_phi) * (costmap_->getResolution() * 0.5);
```
DLu: Can you explain what this math is doing, please?

DLu (changes requested):
Thanks for this pull request. The fact that the tolerance was not used to begin with was an oversight on my part. 

There's a couple key stylistic things and extraneous bits of code that make this PR a little tricky. However, my main question is about how exactly you're picking the new goal location. At the very least, it could use some more comments about how that calculation is done. There's also [a nice-ish implementation in `move_base`](https://github.com/ros-planning/navigation/blob/melodic-devel/move_base/src/move_base.cpp#L376) that just searches with increasing distances, which is a little more straightforward.

nlimpert:
I'm sorry for the super-late response!
I agree that there are major flaws in this PR and I also agree that the solution implemented in `move_base` is much more elegant.

So I currently tend to re-implement this based on the [solution](https://github.com/ros-planning/navigation/blob/b52e0ea9362131dc81326da471792ab8698c9564/move_base/src/move_base.cpp#L376) in `move_base` that @DLu pointed at.

However seeing this makes me wonder why this would need to be implemented in two different places.
But I guess it's better to have a working `default_tolerance` parameter even though the feature is already available in `move_base` (while it doesn't seem to be usable without `move_base`'s `planService`).

I think this would look completely different than the current PR. Would you like me to create a new PR or simply update this one?

nlimpert:
Closing this as it will be targeting melodic and it needs a proper rework of the proposed feature.

I'll open a new PR once I'm ready.

PR #708, opened 2018-04-17
Allow specification of free/occupied thresholds for map_saver

ipa-fez:
Adds optional `-to` and `-tf` arguments to map_saver to specify free and occupied thresholds rather than hardcoding free to exactly 0 and occupied to exactly 100.

This helps f.e. with saving maps generated by cartographer.

### DIFF
--- a/map_server/src/map_saver.cpp
+++ b/map_server/src/map_saver.cpp
@@ -131,7 +131,7 @@ free_thresh: 0.196
 
 #define USAGE "Usage: \n" \
               "  map_saver -h\n"\
-              "  map_saver [-to <threshold_occupied>] [-tf <threshold_free>] [-f <mapname>] [ROS remapping args]"
+              "  map_saver [--occ <threshold_occupied>] [--free <threshold_free>] [-f <mapname>] [ROS remapping args]"
 
 int main(int argc, char** argv)
 {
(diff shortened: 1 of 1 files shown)


rsmallegoor:
Is there an open issue requesting/supporting this feature? I would like to know why this helps with saving maps generated by Cartographer.

Somewhat related: saving the full range instead of hard-coded thresholds was briefly discussed [here](https://github.com/googlecartographer/cartographer_ros/issues/475).

ipa-fez:
Cartographer doesn't generate a binary gridmap where 0 is free and 100 is occupied, the map is continuous. This means that with the current map_saver that uses `== 100` as occupied, many of the detected walls with slightly lower certainty will be saved as "unknown" in the .pgm.

Saving the raw map and setting the thresholds in the .yaml would probably be the nicer solution, but this is enough to produce binary maps for planning.

mikeferguson:
I like this feature -- I'm just not keen on the command line names -- "tf" throughout most of ROS will be thought of as relating to "transform library". What about using "-occ" and "-free" (I know they're longer, but they are more descriptive)?

ipa-fez:
I renamed them to `--occ` and `--free`.

mikeferguson:
@ipa-fez thanks. I'm going to forward port to Lunar/Melodic -- if you could please add documentation on wiki.ros.org for these new parameters (and note they are Kinetic+), that'd be great. No big rush, since our next release is probably a few weeks away.

PR #709, opened 2018-04-20
Add const to getLocalPlan

Rayman:
Hey, I noticed the function getLocalPlan that does not modify it's first argument. I added const to make this more clear.


mikeferguson:
Seems reasonable -- but I think it's a little late to change API in Kinetic -- I'm going to suggest we retarget this to Melodic once that branch is created.

Rayman:
:+1:

mikeferguson:
superceded by #720

PR #713, opened 2018-04-28
Added handling for NaN linear.z components of GPS in `robot_pose_ekf`

garethellis0:
Added a check in `odom_estimation_node.cpp` that checks if `linear.z` values in the GPS `odom` message are `NaN` (something common in such messages), and if so, sets it to 0, and sets the corresponding variance to a very high value so we ignore it.


mikeferguson:
I can't figure out what the build failure was -- closing and reopening to retrigger since it appears it might have been a buildfarm issue

Issue #706, opened 2018-04-13
RotateRecovery parameter names differ from TrajectoryPlannerROS parameter names


MarcoStb1993:
The parameters specified in TrajectoryPlannerROS that according to the documentation will be accepted by RotateRecovery have different than those of RotateRecovery which leads to RotateRecovery using the default values instead of the ones specified by the user in base_local_planners.yaml.
http://wiki.ros.org/rotate_recovery
(acc_lim_th, max_rotational_vel, min_in_place_rotational_vel )
http://wiki.ros.org/base_local_planner?distro=kinetic
(acc_lim_theta, max_vel_theta, min_in_place_vel_theta)

mikeferguson:
Adding melodic milestone to this -- since I'd like to try to address all of these random parameter issues we've found over the years in this new LTS.

mikeferguson:
Is in PR -- closing so I can keep track of what we have to actually still do before release!

PR #721, opened 2018-05-15
Fix for #706

DLu:


mikeferguson:
If you rebase this on top of the latest merge, should pass now

mikeferguson:
@DLu -- I'm cherry picking this into my branch of ABI/API fixes. I'm going to move it into it's own header so I can re-use for all the params we're fixing

PR #728, opened 2018-05-22
switch AMCL to use TF2

mikeferguson:


mikeferguson:
This is tested/working. I was going to wait until I had planners/costmaps working for this to go in -- but I'm still having a hell of a time (several tests won't pass, and, as before, the robot just kinda sits there). I am continuing to work on TF2 conversion, but I would recommend we review this and merge it now, and come back with the other fixes in a later PR.

aaronhoy (approved):
LGTM

PR #738, opened 2018-06-01
Costmap_2d plugin universal parameters and pre-hydro warnings

marting87:
Several fixes for **costmap_2d**:

- Clarified some comments and descriptions (minor)
- Renamed _resetOldParameters_ to _loadOldParameters_. I could not find the reset part in the function :-)
- We might want to encourage the user to configure their plugins in the new way, therefore I upscaled the pre-Hydro info message to a warning.
- _move_base_ will spawn two _costmap_2d_ instances which each use their own set of parameters. From rosout it is not clear which message belongs to which costmap. I've added the costmap name in some log messages:
```
[ INFO] [1527860648.255545032, 90.085000000]: global_costmap: Using plugin "inflation_layer"
[ INFO] [1527860648.362837438, 90.192000000]: local_costmap: Using plugin "static_layer"
```
- The parameters **static_map** and **map_type** are unused in combination with plugins. I've added a warning to inform the user if they are still on the parameter server. Users may think static_map is enabled from their configuration if `static_map = true` and they use for example only an obstacle_filter in combination with _plugins_. 

**-  Now comes the major part of the pull request.**
When filters are enabled the filter parameters need to be explicitly set in the namespace of that specific filter. For example, to use an _inflation_radius_ in the inflation_layer of the global_costmap:
```
global_costmap:
  plugins: 
   - {name: static_layer, type: "costmap_2d::StaticLayer"}
   - {name: inflation_layer, type: "costmap_2d::InflationLayer"}
  inflation_layer:
    inflation_radius: 1.5
```
Setting the inflation_radius in the global_costmap or local_costmap works only for a pre-Hydro configuration, since the move_parameter function is taking care to move the parameter to the appropriate namespace of the filter.

I've added a function to copy parameters that are set outside the scope of the layer (in global_costmap or local_costmap) into their filter, taking into consideration that one might use more that one filter of the same type. 

This make the configuration files backwards-compatible and enables the possibility to set for example a universal inflation_radius for both the local and global costmap. _*As in: [RobotSetup Tutorial](http://wiki.ros.org/navigation/Tutorials/RobotSetup)_

    <rosparam file="$(find bot)/cfg/costmap_common_params.yaml" command="load" ns="global_costmap" />
    <rosparam file="$(find bot)/cfg/costmap_common_params.yaml" command="load" ns="local_costmap" />

**with costmap_common_params.yaml:**
```
inflation_radius: 1.5
```


mikeferguson:
@marting87 I need to review this one still -- I will shortly

costmap_2d/include/costmap_2d/costmap_2d_ros.h
```diff
@@ -241,7 +241,10 @@ class Costmap2DROS
   void readFootprintFromConfig(const costmap_2d::Costmap2DConfig &new_config,
                                const costmap_2d::Costmap2DConfig &old_config);
 
-  void resetOldParameters(ros::NodeHandle& nh);
+  void loadOldParameters(ros::NodeHandle& nh);
+  void warnForOldParameters(ros::NodeHandle& nh);
+  void checkOldParam(ros::NodeHandle& nh, const std::string &param_name);
+  void copyParentParameters(const std::string& plugin_name, const std::string& plugin_type, ros::NodeHandle& nh);
```
DLu, line 248: Do you think any of the functionality could be aided by the new functions [here](https://github.com/ros-planning/navigation/blob/melodic-devel/nav_core/include/nav_core/parameter_magic.h)

costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -135,7 +137,9 @@ Costmap2DROS::Costmap2DROS(std::string name, tf::TransformListener& tf) :
     {
       std::string pname = static_cast<std::string>(my_list[i]["name"]);
       std::string type = static_cast<std::string>(my_list[i]["type"]);
-      ROS_INFO("Using plugin \"%s\"", pname.c_str());
+      ROS_INFO("%s: Using plugin \"%s\"", name_.c_str(), pname.c_str());
```
DLu, line 133: Would making this `ROS_INFO_NAMED` work here?

DLu:
Generally, this looks good. I made an assumption that if you updated your configuration to have a plugins parameter, you would also move the parameters into the proper namespace.

costmap_2d/src/costmap_2d_ros.cpp
marting87: Yes, good one

marting87: Perhaps use ROS_..._NAMED("name.c_str(),..") throughout the whole file?

costmap_2d/include/costmap_2d/costmap_2d_ros.h
marting87: Yes! We could use if for map_type and static_map in checkOldParam. I would suggest to keep copying from the parent namespace to ensure backward compatibility with the old way of defining properties of filters and to give the user the possibility to set for example a default inflation radius for all inflation filters.

DLu:
@marting87 Do you have the bandwidth to implement the remaining suggestions?

marting87:
Unfortunately my time is limited these days (Busy weeks at work).

PR #739, opened 2018-06-04
feat(orientation_filter): Added additional orientation filter options

reinzor:
Enables plan references with different orientations for omni-base
controllers. The following options are added:

- Backward (backward path traversal, pose points to previous point)
- Leftward (lateral path traversal in the positive y direction)
- Rightward (lateral path traversal in the negative y direction)

Also altered the orientation calculation and added additional parameter `window_size` that can be used to smoothen the orientation calculation based on the position derivative.

### DIFF
--- a/global_planner/cfg/GlobalPlanner.cfg
+++ b/global_planner/cfg/GlobalPlanner.cfg
@@ -16,9 +16,12 @@ orientation_enum = gen.enum([
     gen.const("Interpolate", int_t, 2, "Orientations are a linear blend of start and goal pose"),
     gen.const("ForwardThenInterpolate", 
                              int_t, 3, "Forward orientation until last straightaway, then a linear blend until the goal pose"),
+    gen.const("Backward",    int_t, 4, "Orientations point to the previous point on the path"),
+    gen.const("Leftward",    int_t, 5, "Orientations point to the next point on the path (in lateral direction)"),
+    gen.const("Rightward",   int_t, 6, "Orientations point to the previous point on the path + pi / 2 (in lateral direction)"),
                             ], "How to set the orientation of each point")
 
-gen.add("orientation_mode",  int_t, 0, "How to set the orientation of each point", 1, 0, 3,
+gen.add("orientation_mode",  int_t, 0, "How to set the orientation of each point", 1, 0, 6,
         edit_method=orientation_enum)
 
 exit(gen.generate(PACKAGE, "global_planner", "GlobalPlanner"))
--- a/global_planner/cfg/GlobalPlanner.cfg
+++ b/global_planner/cfg/GlobalPlanner.cfg
@@ -5,23 +5,27 @@ from dynamic_reconfigure.parameter_generator_catkin import ParameterGenerator, i
 
 gen = ParameterGenerator()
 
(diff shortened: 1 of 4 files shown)


mikeferguson:
@DLu -- do you have a chance to look over this? I think you're more versed in the global_planner than I am

global_planner/cfg/GlobalPlanner.cfg
```diff
@@ -16,9 +16,12 @@ orientation_enum = gen.enum([
     gen.const("Interpolate", int_t, 2, "Orientations are a linear blend of start and goal pose"),
     gen.const("ForwardThenInterpolate", 
                              int_t, 3, "Forward orientation until last straightaway, then a linear blend until the goal pose"),
+    gen.const("Backward",    int_t, 4, "Orientations point to the previous point on the path"),
+    gen.const("Leftward",    int_t, 5, "Orientations point to the next point on the path (in lateral direction)"),
+    gen.const("Rightward",   int_t, 6, "Orientations point to the previous point on the path + pi / 2 (in lateral direction)"),
```
DLu: Would you mind tweaking the descriptions here such that they specify which axis is leading? i.e. Backward is the negative x axis points at the next point, and Leftward is the positive y axis (I think).

global_planner/src/orientation_filter.cpp
```diff
@@ -60,6 +60,24 @@ void OrientationFilter::processPath(const geometry_msgs::PoseStamped& start,
...
+                pointToNext(path, i);
+                set_angle(&path[i], angles::normalize_angle(getYaw(path[i]) - M_PI_2));
+            }
+            break;
+        case RIGHTWARD:
+            for(int i=0;i<n-1;i++){
+                pointToNext(path, i);
+                set_angle(&path[i], angles::normalize_angle(getYaw(path[i]) + M_PI_2));
+            }
+            break;
```
DLu: Nonblocking suggestion: If you wanted to get really fancy, you could have one additional `ARBITRARY_OFFSET` mode that had a dynamically reconfigurable variable that would specify what angle to offset each angle with, and then use that variable in [`pointToNext`](https://github.com/ruvu/navigation/blob/1f4cb257dc299cd51fd2f3ea866ef0e2cd382c20/global_planner/src/orientation_filter.cpp#L115). Then, when you selected LEFTWARD in the dynamic reconfigure callback, it could set the variable to -M_PI_2. 

This would also have the benefit of reducing the number of calls to quaternion conversion methods.

DLu (commented):
Thanks for the PR. Let me know what you think of my suggestions, and then we can move forward.

It would be cool (though not necessary) if you had a video showing the new behaviors.

global_planner/src/orientation_filter.cpp
reinzor: Thanks for the suggestion. However, I currently do not see a use-case for this.

reinzor:
Video showing the behaviors:

https://www.youtube.com/watch?v=NQ2z90i5V2I&feature=youtu.be

DLu (approved):

reinzor:
Thanks, I also updated the ROS wiki of melodic

moriarty:
FYI: @reinzor: @rayman opened issue #760 because my PR #752 appears to have broken this 👎

PR #740, opened 2018-06-06
Allow costmap_2d's voxel_layer to remove obstacles after a timeout.

vpradeep07:
This adds a new feature to the costmap_2d's voxel_layer, allowing voxels to be removed after a certain specified timeout.  This is done by maintaining a `std::map` of all current voxels and the time at which they were inserted into the map.  This "obstacle_timeout" can also be changed via dynamics reconfigure.

By default, the `obstacle_timeout` parameter is set to 0.0, which results in no change from the current behavior, and no additional computation.  In this default case, obstacles remain indefinitely in the voxel_layer until they are raytraced away.

This feature also required a small API addition to `clearVoxelInMap` which allows the unknown threshold and clear thresholds to be passed in.  (In my opinion, the incantation without thresholds `clearVoxelInMap(unsigned int x, unsigned int y, unsigned int z)` should be removed from the API entirely, but that's probably for a separate discussion)

Here's a related question on ROS Answers:
https://answers.ros.org/question/215445/why-doesnt-costmap-clear-the-old-obstacles-outside-the-scanner-scope/

_This is my first PR into `ros-planning/navigation` so happy to hear if there's some additional process/workflow I need to be following._

### DIFF
--- a/costmap_2d/cfg/VoxelPlugin.cfg
+++ b/costmap_2d/cfg/VoxelPlugin.cfg
@@ -16,6 +16,7 @@ gen.add("mark_threshold", int_t, 0, 'The maximum number of marked cells allowed 
 combo_enum = gen.enum([ gen.const("Overwrite", int_t, 0, "b"),
                         gen.const("Maximum",   int_t, 1, "a") ],
                       "Method for combining layers enum")
+gen.add("obstacle_timeout", double_t, 0, "Seconds after which a voxel should be cleared.  0.0 implies living forever", 0.0, 0, 60)
 
 gen.add("combination_method", int_t, 0, "Method for combining two layers", 1, 0, 2, edit_method=combo_enum)
 
--- a/costmap_2d/include/costmap_2d/voxel_layer.h
+++ b/costmap_2d/include/costmap_2d/voxel_layer.h
@@ -38,6 +38,9 @@
 #ifndef COSTMAP_2D_VOXEL_LAYER_H_
 #define COSTMAP_2D_VOXEL_LAYER_H_
 
+#include <map>
+#include <boost/tuple/tuple.hpp>
+#include <boost/tuple/tuple_comparison.hpp>
 #include <ros/ros.h>
 #include <costmap_2d/layer.h>
 #include <costmap_2d/layered_costmap.h>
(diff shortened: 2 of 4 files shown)


mikeferguson:
@vpradeep07 thanks for your contribution. There isn't any additional workflow to follow -- it just takes us a little while to review things.

One thing I do notice from a quick pass is that this changes API/ABI -- do you need this in Kinetic, or would Melodic be OK (we haven't made a first release in Melodic yet, so we can still bust the API/ABI all we want)

vpradeep07:
Sorry for the delay (I think my github notifications are getting routed to the wrong place for this repo).

Regd API/ABI compatibility: Good point, I wasn't thinking about that at all when I submitted the PR.  I think I have to make header file changes to support this change.  At what level does API/ABI compatibility need to be maintained (e.g. all header files, a subset of important headers, etc)?

This feature is a definitely requirement for one of our projects running on Kinetic.  If you need to reject this PR due to API/ABI breakages, I understand.  In this case, we'll continue using our fork of the navigation stack until we port forward to a future ROS release (probably not for 6-12 months).  That's not a great situation, but it's manageable for us and not a blocker.

If you have any suggestions for how to re-implement this feature in Kinetic without breaking ABI/API compatibility (or with breaking a less important ABI/API), I'm definitely open to suggestions.

vpradeep07:
From an API/ABI standpoint, would it be ok to create a new layer called `VoxelTimeoutLayer`, which contained the additional voxel timeout functionality?  I'd probably remove the dynamic reconfigure cfg modifications, since those would break API/ABI compatibility in the existing `voxel_layer.h`.

Since I would be purely adding functionality and not changing the existing API/ABI, hopefully it would be an acceptable change :-)

DLu:
I would definitely recommend making this its own Layer, which doesn't even need to be merged into this repository. You could just subclass the existing VoxelLayer and add your functionality. 

Thanks for tackling this!

vpradeep07:
Thanks @DLu and @mikeferguson for the feedback.  Closing this PR for now.  This feature probably would be better as it's own layer, since any sort of patch will for this feature will break ABI compatibility.

With that said, I just became aware of the following package, which seems to do exactly what we're looking for: http://wiki.ros.org/spatio_temporal_voxel_layer
We'll give `spatio_temporal_voxel_layer` a shot instead of building yet another voxel grid.  Thanks @SteveMacenski for pointing this out!

PR #735, opened 2018-05-29
Allow specification of free/occupied thresholds for map_saver (#708)

mikeferguson:
* add occupied threshold command line parameter to map_saver (--occ)
* add free threshold command line parameter to map_saver (--free)

PR #742, opened 2018-06-11
Allow specification of free/occupied thresholds for map_saver (#708)

mikeferguson:
* add occupied threshold command line parameter to map_saver (--occ)
* add free threshold command line parameter to map_saver (--free)


mikeferguson:
Lunar port of #735

PR #745, opened 2018-06-21
Fix bearing for extremely close poses

timethy:
I think that delta_rot1 should be M_PI if the translation is backwards.
This fixes in-place rotations that slightly go backwards.

### DIFF
--- a/amcl/src/amcl/sensors/amcl_odom.cpp
+++ b/amcl/src/amcl/sensors/amcl_odom.cpp
@@ -263,11 +263,15 @@ bool AMCLOdom::UpdateAction(pf_t *pf, AMCLSensorData *data)
     // Avoid computing a bearing from two poses that are extremely near each
     // other (happens on in-place rotation).
     if(sqrt(ndata->delta.v[1]*ndata->delta.v[1] + 
-            ndata->delta.v[0]*ndata->delta.v[0]) < 0.01)
-      delta_rot1 = 0.0;
-    else
+            ndata->delta.v[0]*ndata->delta.v[0]) < 0.01) {
+      // Make sure we still update translation in right direction!
+      double projected_movement = cos(old_pose.v[2])*ndata->delta.v[0] +
+                                  sin(old_pose.v[2])*ndata->delta.v[1];
+      delta_rot1 = projected_movement < 0.0 ? M_PI : 0.0; // or M_PI
+    } else {
       delta_rot1 = angle_diff(atan2(ndata->delta.v[1], ndata->delta.v[0]),
                               old_pose.v[2]);
+    }
     delta_trans = sqrt(ndata->delta.v[0]*ndata->delta.v[0] +
                        ndata->delta.v[1]*ndata->delta.v[1]);
     delta_rot2 = angle_diff(ndata->delta.v[2], delta_rot1);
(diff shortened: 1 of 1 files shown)


mikeferguson:
Can you provide a little more information -- I'm not sure I'm following the math here.

wienans:
@timethy Thanks for this fix, it worked just fine :) had the same issue while driving very slowly backwars and having a rotational component. It looked in rviz like amcl thinks the robot is moving forward but actually the robot is moving backward. 
@mikeferguson i think projected_movement is a complicated way to check if the robot is driving forward or backward, but i don't understand it fully too, because i am not into all the variables. Maybe it is achievable with other simpler math.

timethy:
Hi, sorry, I didn't follow this thread as it fixed the problem for me. Is there still interest in merging this upstream, if yes I can look into it.

wienans:
Hi @timethy as i had this issue with the current releases this is a present issue. And maybe also present for other dirstro's. 
Maybe you can give a bit inside to the projected_movement math you did. Would be interesting for me too.

timethy:
Hi, so I went through the math again. It is based on the Algorithm sample_motion_model_odometry from the book Probabilistic Robotics [1] (you can find a pdf online; it's also where the two figures come from):

sample_motion_model_odometry

In the code, `delta_rot1`, `delta_trans` and `delta_rot2` correspond to these three relative motion parameters and are calculated from the difference in poses, which is in `ndata->delta.v[0]` (delta x = x' - x), `ndata->delta.v[1]` (delta y = y' - y) and `ndata->delta.v[2]` (delta theta = theta' - theta).

Essentially the robots motion is approximated to happen in a first turn (by `delta_rot1`), then a straight translation (by `delta_trans`) and finally a second (by `delta_rot2`).
After sampling and applying (independent) noise to `delta_rot1`, `delta_trans` and `delta_rot2` the robot's position estimate is updated. (See figure)
This is done by adding the traveled distance `delta_trans` along the heading after the first rotation (i.e. `omega` + `delta_rot1`).

motion_model_odometry

Now the problem
 is, if the robot is (almost) turning on the spot it becomes (almost) impossible to split the turn into `delta_rot1` and `delta_rot2` simply because `arctan(dy/dx)` is undefined if dx is (close to) 0.
So instead, the motion is just assumed to be a straight line motion first along the current heading by `delta_trans` and then followed by a turn by `delta_rot2` by setting `delta_rot1` to zero.
In the case of a perfect in-turn motion, `delta_trans` would be zero. However, in real life the robot is still going to move slightly in `x` and `y`, so in practice `delta_trans` is only close to +0.

Now assume (in terms of the robot's old heading theta) the new robot's position lies **behind** the robot's current position. Then delta_trans (an unsigned) value doesn't capture this and the new robot's estimate would lie **ahead** of the robot's current position. Sure, if `delta_trans` is really small, it's not a big problem. However over the course of a turn (and typically it's either a left or a right turn but never both) this can seriously accumulate.

What my change does is to see if the new position lies ahead or behind and set `delta_rot1` to 0 or π, respectively. In the latter case, this means we assume the robot does a half turn, moves in that direction and then does another (almost) half turn. While this might sound counterintuitive, it works well.

Happy to make the code nicer, or explore other ideas. One option would be to simply switch to a different noise model using only `delta_trans` and `delta_rot2` in this case. (Which I would find cleaner.)
Or one could even switch to a different noise model with noise on delta x, delta y and delta omega, in the case that `delta_trans` are small.

[1] Thrun, Sebastian, Wolfram Burgard, and Dieter Fox. "Probabilistic robotics. 2005." Massachusetts Institute of Technology, USA (2005).

mikeferguson:

Issue #750, opened 2018-06-26
Already moving robot does not stop if planner can not make a plan immediately


sunilsulania:
I have a robot running in a physical loop path with move_base running only in one of its straight segments. So i switch (Give goal to actionserver) to move_base while robot is moving. 
In a condition when **MoveBase::makePlan** return false immediately after switching(giving goal to actionserver) velocities are not updated to zero causing robot to not stop. 
makePlan returns false because there is an obstacle covering goal. And as an effect of this, robot goes crashing into obstacle.

I think I got the reason for it :

```
bool gotPlan = n.ok() && makePlan(temp_goal, *planner_plan_);

if(gotPlan){
.
.
.}
else if(state_==PLANNING)
{
.
.
    if(runPlanner_ && (ros::Time::now() > attempt_end || planning_retries_ > uint32_t(max_planning_retries_)))
    {
        .
        publishZeroVelocity();
        .
    }
.
.}
```
`(ros::Time::now() > attempt_end || planning_retries_ > uint32_t(max_planning_retries_))` This term is false and robot does not stop

mikeferguson:
I think I might have seen this before and was blaming the simulation. Could you open a PR?

mikeferguson:
Ok -- I think this is different from what I've seen -- but I think you're right -- I don't see anywhere in the executeCb() where we set velocities to zero. That should probably be done right before we change state to planning.

DLu:
Closed via #751

PR #751, opened 2018-07-05
Fix for #750: Added publishZeroVelocity() before starting planner

sunilsulania:
Edit for Issue #750


sunilsulania:
I have tested this on my stage simulation environment and this edit seems to have fixed the bug. I am not very sure about future planned changes in move_base and if I have added fix at correct location, please confirm before merging.

PR #757, opened 2018-07-15
Kinetic devel

leankyr:
in /navigation/amcl/CMakeLists.txt I suggest adding the line 
add_compile_options(-std=c++11)

Otherwise /navigation/amcl/src/amcl_node.cpp will not build with catkin_
make in ROS-kinetic

[CMakeLists.txt](https://github.com/ros-planning/navigation/files/2195639/CMakeLists.txt)


mikeferguson:
If you're making a PR to Melodic, it has to start from melodic-devel branch. Also, I'm skeptical we need to add flags since CI is passing on both Kinetic and Melodic branches.

Issue #760, opened 2018-07-16
PR #752 broke orientation filter


Rayman:
Hey @moriarty, after your PR, the orientation filter doesn't work anymore. I don't see anyting obvious wrong in the code, but it behaves differently. See the screenshots below.

melodic-devel (96c5f9fce6bdc0128623b3e4b184bcbe2909403d)
![screenshot from 2018-07-16 09-47-17](https://user-images.githubusercontent.com/91969/42748151-5aeb1f16-88df-11e8-81b4-c06e87a44567.png)

8ea5462f5395acf9741253ff38302d29175dd870
![screenshot from 2018-07-16 09-50-31](https://user-images.githubusercontent.com/91969/42748153-5f2667e8-88df-11e8-969f-771dfaa298a4.png)

moriarty:
Damn. I didn’t catch that when I was testing #752, but that might explain something I saw with #755.

I will take a look

moriarty:
I think this is a difference in ```tf2::getYaw``` @Rayman can you try replacing that with ```tf2::getEulerYPR``` ?

The old under the hood looks like it called getRPY and threw away the pitch and roll.

http://docs.ros.org/kinetic/api/tf/html/c++/transform__datatypes_8h_source.html#l00138
```
  131 static inline double getYaw(const Quaternion& bt_q){
  132   tfScalar useless_pitch, useless_roll, yaw;
  133   tf::Matrix3x3(bt_q).getRPY( useless_roll, useless_pitch,yaw);
  134   return yaw;
  135 }
  136 
  138 static inline double getYaw(const geometry_msgs::Quaternion& msg_q){
  139   Quaternion bt_q;
  140   quaternionMsgToTF(msg_q, bt_q);
  141   return getYaw(bt_q);
  142 }
```

http://docs.ros.org/kinetic/api/tf/html/c++/Matrix3x3_8h_source.html#l00365
```
  365         void getRPY(tfScalar& roll, tfScalar& pitch, tfScalar& yaw, unsigned int solution_number = 1) const
  366         {
  367         getEulerYPR(yaw, pitch, roll, solution_number);
  368         }
```




http://docs.ros.org/jade/api/tf2/html/impl_2utils_8h_source.html
```
00086 inline
00087 void getEulerYPR(const tf2::Quaternion& q, double &yaw, double &pitch, double &roll)
00088 {
00089   double sqw;
00090   double sqx;
00091   double sqy;
00092   double sqz;
00093 
[...]

PR #752, opened 2018-07-11
[tf2] cherry pick: Move some packages to tf2

moriarty:
This cherry-picks commit 144cae29
https://github.com/ros-planning/navigation/pull/531 from @vrabaud which targeted Kinetic.

I fixed the conflicts.

Conflicts:
	base_local_planner/package.xml
	base_local_planner/src/trajectory_planner_ros.cpp
	carrot_planner/package.xml
	global_planner/src/orientation_filter.cpp

 Author:    Vincent Rabaud
 Date:      Mon Oct 24 22:08:55 2016 +0200

PR #755, opened 2018-07-15
tf2 for Melodic

moriarty:
This cherry-picks the two commits from @vrabaud for #458.

Then I fixed the compile errors and rebased onto melodic after #752 was merged today.
There were lots of conflicts because some things were already ported to tf2.

The recent changes for #601 
https://github.com/ros-planning/navigation/commit/0bd8f30254678652ec144571f83234b21e0ab2c4
caused compile errors because the code wasn't there at the time of #458. But the code is nearly duplicate of the getRobotPose function in CostMaps

I will test this more on a Robot today/tomorrow

PR #762, opened 2018-07-17
Fix: Replaced tf2 setEuler calls with setRPY calls

reinzor:
Introduced by https://github.com/ros-planning/navigation/pull/752. Closes https://github.com/ros-planning/navigation/issues/760

### DIFF
--- a/base_local_planner/src/trajectory_planner_ros.cpp
+++ b/base_local_planner/src/trajectory_planner_ros.cpp
@@ -518,7 +518,7 @@ namespace base_local_planner {
       pose.pose.position.y = p_y;
       pose.pose.position.z = 0.0;
       tf2::Quaternion q;
-      q.setEuler(p_th, 0, 0);
+      q.setRPY(0, 0, p_th);
       tf2::convert(q, pose.pose.orientation);
       local_plan.push_back(pose);
     }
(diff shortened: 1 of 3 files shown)


mikeferguson:
Retriggering CI -- not sure why inflation test failed (I've seen it occasionally happen a while ago, but not recently -- so it is still likely unrelated -- we'll see if this passes on second try).

mikeferguson:
heads up to @moriarty

DLu (approved):

moriarty:
Thanks!

mikeferguson:
This seems to work fine for me -- but my setup is using navfn, so it isn't testing global planner (and thus orientation filter) -- is there someone who has tested this on a robot config that does use orientation filter? (Also, if someone could share which robot config they test with, if it's freely available on Gazebo, I can look at adding it to my manual test suite)

moriarty:
@rayman & @reinzor are you testing this with @ruvu or @tue-robotics stuff? @LoyVanBeek TU/e stuff is available in simulation isn't it?

LoyVanBeek:
I think this is @reinzor's @ruvu stuff cuz' it's vacation time @tue-robotics apparently. And yes, you can test our stuff in simulation, but with a custom [simulator](https://github.com/tue-robotics/fast_simulator).

moriarty:
@mikeferguson, Rein added a comment here:
https://github.com/ros-planning/navigation/pull/752/files/88ada10d9129d529a373162f42fb74cd0821d649#r202910243
(updated link)

mikeferguson:
This is clearly better than the mainline -- and I think the Euler vs. RPY is what is also screwing up #755 as well. After changing those 4 other instances, I've got a working nav stack (in sim, with navfn + base_local_planner)

Issue #264, opened 2014-10-26
Remove dependency on PCL


mikeferguson:
Sensor_msgs now includes a C++ iterator for pointcloud2 messages, we should use this in costmap updater to avoid a dependency on PCL. 

As mentioned in https://github.com/ros/common_msgs/pull/16 this probably needs a tf/tf2 sensor_msgs::PointCloud2 transform to be implemented.

vrabaud:
Hi, this has actually been implemented:
https://github.com/ros/geometry_experimental/blob/indigo-devel/tf2_sensor_msgs/include/tf2_sensor_msgs/tf2_sensor_msgs.h
Now, navigation actually needs to switch to tf2 before that happens. #298 is here for that.

mikeferguson:
So, Jade has come and gone years ago. @DLu -- I'm planning to work on this in the next few weeks, and I'm targeting Melodic.

mikeferguson:
Remaining PCL items:

 * navfn_ros.h/cpp uses pcl_publisher -- needs to be converted to sensor_msgs iterator
 * base_local_planner/map_grid_visualizer.h -- uses pcl_publisher
 * dwa_local_planner still depends on PCL, but I can't find any usage.

moriarty:
> dwa_local_planner still depends on PCL, but I can't find any usage.

```
________________________________________________
Errors     << dwa_local_planner:make /home/fetch/melodic-stable/logs/dwa_local_planner/build.make.000.log                                                                                                         
In file included from /opt/ros/melodic-stable/include/base_local_planner/map_grid_visualizer.h:40:0,
                 from /home/fetch/melodic-stable/src/navigation/dwa_local_planner/include/dwa_local_planner/dwa_planner.h:47,
                 from /home/fetch/melodic-stable/src/navigation/dwa_local_planner/src/dwa_planner.cpp:37:
/opt/ros/melodic-stable/include/base_local_planner/map_grid_cost_point.h:37:10: fatal error: pcl/register_point_struct.h: No such file or directory
 #include <pcl/register_point_struct.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Your branch is probably still a work in progress but I tried to compile it and found these usages

```
fetch at fetch3 in ~/melodic-stable/src/navigation/dwa_local_planner on remove_pcl
$ grep -r pcl . -n
./src/dwa_planner.cpp:49:#include <pcl_conversions/pcl_conversions.h>
[...]

PR #765, opened 2018-07-21
remove left over PCL depends in dwa_local_planner

mikeferguson:


DLu:
Part of #264

moriarty:
while this branch is open, navfn/CMakeLists.txt missed a line

moriarty:
It also looks like it's still depended on in costmap_2d but only used in two tests.

DLu:
There's PCL stuff in the two costmap test files, but the static tests PCL line is commented out, and the module_tests isn't currently used, so for the sake of expediency, I'm not converting those.

mikeferguson:
LGTM -- passing CI.

PR #770, opened 2018-07-27
Fix debian builds (closes #769)

### LINKED ISSUE #769: Melodic debians are missing depends
I’ll be working on this later tonight. Filing an issue so we don’t duplicate effort.

mikeferguson:


mikeferguson:
First commit definitely belongs. Second is a bit of a wild guess since I can't figure out why the costmap_2d package isn't building. I compared with robot_calibration (something I released a few weeks ago and is building, and has similar depends), and it seems either it is the addition of Boost, or the removal of TF that is causing the issue....

mikeferguson:
Forget that first comment -- robot_calibration has same Boost line, so my second commit was definitely garbage...

I'm taking a closer look at what TF depends on...

mikeferguson:
Ok -- figured it out -- had to scroll back up further -- we have a find_package on tf2_geometry_msgs but it is not listed in package.xml

PR #771, opened 2018-07-30
[melodic] Fix deps

DLu:
For AMCL, [previous commit](https://github.com/ros-planning/navigation/commit/960ec291e02b7b80282762f4d8cd77e698a0e728) did not add `sensor_msgs` in all the right places. I went a little further and put all the packages in the `CATKIN_DEPENDS` and added two packages used by `amcl_node`.

I'm not certain what's causing the `costmap_2d` problems, but I added some dependencies just in case.

### DIFF
--- a/costmap_2d/CMakeLists.txt
+++ b/costmap_2d/CMakeLists.txt
@@ -15,6 +15,7 @@ find_package(catkin REQUIRED
             roscpp
             sensor_msgs
             std_msgs
+            tf2
             tf2_geometry_msgs
             tf2_ros
             tf2_sensor_msgs
@@ -74,7 +75,10 @@ catkin_package(
         roscpp
         sensor_msgs
         std_msgs
+        tf2
+        tf2_geometry_msgs
         tf2_ros
+        tf2_sensor_msgs
         visualization_msgs
         voxel_grid
     DEPENDS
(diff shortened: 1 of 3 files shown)


mikeferguson:
Just a note - the PR for my release hasn't been merged yet -- so build failures you're seeing are still related to 1.16.0, which doesn't include my fixes. I agree there may be other issues that turn up after that actually gets on the build farm, but I'm hesitant to wildly add dependencies without actually looking at what headers get included.

TLDR: Please hold on merging this until I can review it fully.

mikeferguson:
There are several other issues I'm working through right now -- the AMCL portions of this are included (and slightly improved) in #773.

costmap_2d/CMakeLists.txt
```diff
@@ -74,7 +75,10 @@ catkin_package(
         roscpp
         sensor_msgs
         std_msgs
+        tf2
+        tf2_geometry_msgs
         tf2_ros
+        tf2_sensor_msgs
```
mikeferguson, line 81: this and tf2_geometry_msgs can't be CATKIN_DEPENDS, since they are declared as build dependencies only in the package.xml. This is because they are header-only packages, and those header files are only used in .cpp files (if they were leaked into our headers, then they become part of the public API, and we need to make them run depends as well as making them CATKIN_DEPENDS).

PR #772, opened 2018-07-30
Dwa planner mbf

vsteinb:
Added a second inheritance to DWAPlannerROS from the mbf_costmap_core controller interface to be able to use move_base_flex while backwards compatible to move_base.


spuetz:
Same as #774 : circular dependencies. @corot How could we handle that?

mikeferguson:
Closing due to unresolved circular dependency

PR #774, opened 2018-07-31
Implement the new move base flex interface for the global planner

vsteinb:
Added a second inheritance to GlobalPlanner from the mbf_costmap_core::CostmapPlanner interface to be able to use move_base_flex while backwards compatible to move_base.

### DIFF
--- a/global_planner/CMakeLists.txt
+++ b/global_planner/CMakeLists.txt
@@ -8,6 +8,7 @@ find_package(catkin REQUIRED
     dynamic_reconfigure
     geometry_msgs
     nav_core
+    mbf_costmap_core
     navfn
     nav_msgs
     pluginlib
--- a/global_planner/package.xml
+++ b/global_planner/package.xml
@@ -24,6 +24,7 @@
   <build_depend>pluginlib</build_depend>
   <build_depend>roscpp</build_depend>
   <build_depend>tf</build_depend>
+  <build_depend>mbf_costmap_core</build_depend>
 
   <run_depend>costmap_2d</run_depend>
   <run_depend>dynamic_reconfigure</run_depend>
@@ -34,8 +35,10 @@
   <run_depend>pluginlib</run_depend>
   <run_depend>roscpp</run_depend>
   <run_depend>tf</run_depend>
+  <run_depend>mbf_costmap_core</run_depend>
   <export>
       <nav_core plugin="${prefix}/bgp_plugin.xml" />
+      <mbf_costmap_core plugin="${prefix}/bgp_mbf_plugin.xml" />
   </export>
 
 </package>
--- a/global_planner/src/planner_core.cpp
+++ b/global_planner/src/planner_core.cpp
@@ -40,6 +40,7 @@
 #include <tf/transform_listener.h>
 #include <costmap_2d/cost_values.h>
 #include <costmap_2d/costmap_2d.h>
+#include <mbf_msgs/GetPathResult.h>
 
 #include <global_planner/dijkstra.h>
 #include <global_planner/astar.h>
@@ -49,6 +50,7 @@
 
 //register this planner as a BaseGlobalPlanner plugin
 PLUGINLIB_EXPORT_CLASS(global_planner::GlobalPlanner, nav_core::BaseGlobalPlanner)
+PLUGINLIB_EXPORT_CLASS(global_planner::GlobalPlanner, mbf_costmap_core::CostmapPlanner)
 
 namespace global_planner {
 
@@ -68,11 +70,11 @@ void GlobalPlanner::outlineMap(unsigned char* costarr, int nx, int ny, unsigned 
 }
 
 GlobalPlanner::GlobalPlanner() :
-        costmap_(NULL), initialized_(false), allow_unknown_(true) {
+        costmap_(NULL), initialized_(false), allow_unknown_(true), canceled_(false) {
 }
 
 GlobalPlanner::GlobalPlanner(std::string name, costmap_2d::Costmap2D* costmap, std::string frame_id) :
-        costmap_(NULL), initialized_(false), allow_unknown_(true) {
+        costmap_(NULL), initialized_(false), allow_unknown_(true), canceled_(false) {
     //initialize the planner
     initialize(name, costmap, frame_id);
 }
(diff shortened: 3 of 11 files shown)


mikeferguson:
So, this really isn't going to work -- we can't depend on things from the MBF repo, since the MBF repo has packages that depend on at least costmap_2d, base_local_planner, and nav_core.

spuetz:
Aaa circular dependencies. :/ @corot How could we handle that?

corot:
I think build just fails because mbf_msgs dependency is missing; same for the local planner.

I don't see any circular dependency; true mbf_costmap_core depends on costmap_2d, but also does nav_core

global_planner/src/planner_core.cpp
```diff
@@ -271,7 +286,7 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
     if (!costmap_->worldToMap(wx, wy, goal_x_i, goal_y_i)) {
         ROS_WARN_THROTTLE(1.0,
                 "The goal sent to the global planner is off the global costmap. Planning will always fail to this goal.");
-        return false;
+        return mbf_msgs::GetPathResult::INTERNAL_ERROR;
```
spuetz, line 289: `INVALID_GOAL` would be the right outcome number code here.

global_planner/src/planner_core.cpp
```diff
@@ -256,7 +271,7 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
     if (!costmap_->worldToMap(wx, wy, start_x_i, start_y_i)) {
         ROS_WARN(
                 "The robot's start position is off the global costmap. Planning will always fail, are you sure the robot has been properly localized?");
-        return false;
+        return mbf_msgs::GetPathResult::INTERNAL_ERROR;
```
spuetz, line 274: `INVALID_START` would be the right number code here.

global_planner/src/planner_core.cpp
```diff
@@ -221,11 +223,24 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
...
     boost::mutex::scoped_lock lock(mutex_);
     if (!initialized_) {
         ROS_ERROR(
                 "This planner has not been initialized yet, but it is being used, please call initialize() before use");
-        return false;
+        return mbf_msgs::GetPathResult::NOT_INITIALIZED;
+    }
+
+    if (canceled_) {
+        ROS_ERROR("This planner was canceled!");
```
spuetz, line 242: The error msg is a bit unspecific.

global_planner/src/planner_core.cpp
```diff
@@ -221,11 +223,24 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
 
 bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geometry_msgs::PoseStamped& goal,
                            double tolerance, std::vector<geometry_msgs::PoseStamped>& plan) {
+    double cost; // Dummy
+    std::string message; //Dummy
+    return makePlan(start, goal, tolerance, plan, cost, message) == mbf_msgs::GetPathResult::SUCCESS;
+}
+
+uint32_t GlobalPlanner::makePlan(const geometry_msgs::PoseStamped &start, const geometry_msgs::PoseStamped &goal,
+                  double tolerance, std::vector<geometry_msgs::PoseStamped> &plan, double &cost, std::string &/*message*/) {
```
spuetz, line 232: - /*message*/ ? 
- where is the cost computed
- where is the message set? -> The message should be defined following to the error meaning.

global_planner/package.xml
```diff
@@ -24,6 +24,7 @@
   <build_depend>pluginlib</build_depend>
   <build_depend>roscpp</build_depend>
   <build_depend>tf</build_depend>
+  <build_depend>mbf_costmap_core</build_depend>
```
spuetz, line 28: mbf_msgs is missing here

global_planner/package.xml
```diff
@@ -34,8 +35,10 @@
   <run_depend>pluginlib</run_depend>
   <run_depend>roscpp</run_depend>
   <run_depend>tf</run_depend>
+  <run_depend>mbf_costmap_core</run_depend>
   <export>
```
spuetz, line 39: mbf_msgs

global_planner/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ find_package(catkin REQUIRED
     dynamic_reconfigure
     geometry_msgs
     nav_core
+    mbf_costmap_core
     navfn
```
spuetz, line 12: mbf_msgs is missing here

mikaelarguedas:
I believe what @mikeferguson means is that the ROS release infrastructure works on a per repository basis.
So circular dependencies should not only be avoided at the package level but also at the repository level. And this PR will introduce a repository level circular dependency between `navigation` and `move_base_flex` that will prevent them from being released

spuetz (changes requested):
There are some things to change. And some things also missing. mbf_msgs dependency, and cost is not computed or set and message is not set. The rest looks fine to me.

mikeferguson:
Yes, what @mikaelarguedas mentioned is the problem -- since we certainly want to release debians.

I think the proper way to get this out would be to include a minimal set of headers in nav_core (Ala this PR: https://github.com/ros-planning/navigation/pull/605). Since we haven't sync'd the first round of debians yet -- it is definitely still early enough to get this into Melodic.

Or split the core stuff out of move_base_flex into a new repo (but before I would support navigation depending on said code, I'd suggest that move_base_flex_core be part of ors-planning, so that if in the future you guys decide to stop maintaining it, we can still release it and the navigation stack).

jspricke:
I would propose to not change navigation and handle the extended plugins at runtime in mbf.

corot:
That's what we currently do; when dealing with "old-stile" API (nav_core) plugins, MBF wraps them so they can be used indistinctly from "new-stile" API (mbf_core) plugins. Problem is that we miss the richer interface, as nav_core plugins don't provide error code or message, just success/failure, nor can be canceled (plus other new fancy stuff).

corot:
So as I see, there are several possibilities (all for melodic and later, kinetic will just remain as it is):

- add mbf_core and mbf_msgs to ROS navigation repo, and upgrade the plugins to the new interface (backward compatible!). Main problem is the increased complexity (possibly confusing for newcomers) of ROS navigation stack
- move the plugins to a separated repo, so the can freely depend on both ROS nav (core, msgs, costmaps...) and MBF (mbf_core, mbf_msgs). Probably too much work for the maintainers.
- duplicate the plugins and re-release with a mbf_ prefix. Duplicate == :imp: 
- let things as they are and make a common interface (both C++ plugin and ROS API) for ROS 2.0 navigation.

mikeferguson:
Going to close this since we never resolved the circular dependency - and we certainly aren't going to make this change in Kinetic

PR #785, opened 2018-09-04
amcl: Add compile option C++11

mintar:
This is required to build the melodic-devel branch of the navigation stack on kinetic. That is quite useful for porting other packages to the new navigation API without having to install a full-blown melodic system.


mikeferguson:
Might it make sense to add this in a more generic way? (for instance, see discussion in https://github.com/ros-perception/sparse_bundle_adjustment/pull/6)

mintar:
Yup, that works and is much nicer! I've updated my PR.

Issue #786, opened 2018-09-05
Undefined reference in navfn library


ipa-fez:
https://github.com/ros-planning/navigation/blob/c7bcd005fde62d3eaa2069b96f33ba4c472eadb8/navfn/include/navfn/navfn.h#L96

The function in the cpp file doesn't use a const COSTMAP* which means that the declaration and definition don't match up.

This produces a linker error if you try to use this function:
```
navfn_test.cpp:(.text+0x44): undefined reference to `navfn::create_nav_plan_astar(unsigned char const*, int, int, int*, int*, float*, int)'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The const from the header is probably wrong, since the array will get modified.

DLu:
For reference: 
https://github.com/ros-planning/navigation/blob/c7bcd005fde62d3eaa2069b96f33ba4c472eadb8/navfn/src/navfn.cpp#L58

I'm not sure if the A* code actually works, so I would generally be wary of running it, but if you had a PR to fix the const issue I'd merge it.

ipa-fez:
https://github.com/ros-planning/navigation/pull/788 and #789 

I ended up not using this function and am now using the NavFn class directly so I have no idea if the code works, but this should fix the signature mismatch.

mikeferguson:
Fixed in at least Melodic

PR #788, opened 2018-09-06
Remove const from create_nav_plan_astar

ipa-fez:
Fixes https://github.com/ros-planning/navigation/issues/786

Fixes function signature so it matches the implementation

### DIFF
--- a/navfn/include/navfn/navfn.h
+++ b/navfn/include/navfn/navfn.h
@@ -93,7 +93,7 @@ positions at about 1/2 cell resolution; else returns 0.
 
 */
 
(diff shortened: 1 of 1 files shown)


mikeferguson:
Hmm, so this bug has existed for 9 years (based on git-blame). Does the underlying code even work? I'm hesitant to merge what I presume is an ABI breaker if the underlying code it lets you access doesn't even work (same concern does not apply to melodic, where I have merged)

ipa-fez:
As mentioned in the issue, I ended up not using the function after all so I don't know if it works either.

DLu:
Fixed in melodic/noetic (see #789). Won't fix in kinetic because of ABI break.

Issue #768, opened 2018-07-25
Global costmap resized by static layer even if layered_costmap_->isSizeLocked() is set to true


ghost:
In the function staticLayer::incomingMap(navigation/costmap_2d/plugins/static_layer.cpp line 166), the first if statement results in master map(layered_costmap) being resized even if the setting for layered_costmap_->isSizeLocked is set to true. This seems unintended to me and I was wondering the changes below should be made. 

There are few other changes to be made for this to fully work, but for now I just wanted to ask whether I was correct in reading this as a bug.

The lines in question: 
https://github.com/ros-planning/navigation/blob/7fa21a2759806ffea7fee9be436ab27b6b3c102f/costmap_2d/plugins/static_layer.cpp#L174-L185

And the changes I'm suggesting:
```
-  if (!layered_costmap_->isRolling() && (master->getSizeInCellsX() != size_x ||
+  if (!layered_costmap_->isRolling() && !layered_costmap_->isSizeLocked()
+   && (master->getSizeInCellsX() != size_x ||
    master->getSizeInCellsY() != size_y ||
    master->getResolution() != new_map->info.resolution ||
    master->getOriginX() != new_map->info.origin.position.x ||
-   master->getOriginY() != new_map->info.origin.position.y ||
-   !layered_costmap_->isSizeLocked()))
+   master->getOriginY() != new_map->info.origin.position.y)) 
```

moriarty:
Sorry this comment won't answer your question. I just saw a wall of unformatted text and had to say something.

I guess there is no "how to file a bug" for this repository, but some more strict repos would require a little bit more basic information for [example](https://github.com/ev3dev/ev3dev/blame/ev3dev-jessie/ISSUE_TEMPLATE.md)

FYI, if you find the lines in GitHub, you can click on them, and then select "copy permalink" which is much easier to read... Then you can also use three ` marks, to type code blocks and use diff format to show clearer what you mean.

https://github.com/ros-planning/navigation/blob/7fa21a2759806ffea7fee9be436ab27b6b3c102f/costmap_2d/plugins/static_layer.cpp#L174-L185

I _might_ have made a typo, but from reading your text and typing it again (this is easier if you try it on your machine first and just copy some of the output of git diff), I think you want to replace the lines above, with:

```diff
-  if (!layered_costmap_->isRolling() && (master->getSizeInCellsX() != size_x ||
+  if (!layered_costmap_->isRolling() && !layered_costmap_->isSizeLocked()
[...]

PR #795, opened 2018-09-20
Dev map parameters

dorezyuk:
Hi all, 

- I would like to suggest to unify the map parameters across the library. For this I wrote a small struct called map parameters (first commit)

- I think in the static layer plugin the logic should be AND NOT layered_costmap_->isSizeLocked() instead of an OR with the parameters. Maybe I am missing something here?


dorezyuk:
Actually I just found out that someone already had the same suspicion about the static layer:  #768

mikeferguson:

PR #800, opened 2018-10-12
[kinetic] Add additional linked libraries

DLu:
Based on [this answers.ros.org post](https://answers.ros.org/question/303209/slow-build-time/) I did some poking at the compile time of `costmap_2d` with `catkin_make`. It does in fact take forever when these dependencies are missing. 

Will port to indigo, lunar and melodic later.


mikeferguson:
LGTM - note that when porting to Melodic, you should remove the PCL part, since we have removed PCL entirely from Melodic branch

aaronhoy (approved):
LGTM

PR #807, opened 2018-11-01
feat: AMCL Diagnostics

reinzor:
Diagnostic task that monitors the estimated standard deviation of the
filter.

Users can tweak the `max_std_x`, `max_std_y` and `max_std_yaw` parameters according to their needs. A warning diagnostic message will be published if the standard deviation exceeds one of these limits.

### DIFF
--- a/amcl/src/amcl_node.cpp
+++ b/amcl/src/amcl_node.cpp
@@ -1599,7 +1599,7 @@ AmclNode::standardDeviationDiagnostics(diagnostic_updater::DiagnosticStatusWrapp
 
   if (std_x > max_std_x_ || std_y > max_std_y_ || std_yaw > max_std_yaw_)
   {
-    diagnostic_status.summaryf(diagnostic_msgs::DiagnosticStatus::WARN, "Too large");
+    diagnostic_status.summary(diagnostic_msgs::DiagnosticStatus::WARN, "Too large");
   }
   else
   {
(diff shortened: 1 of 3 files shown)


amcl/src/amcl_node.cpp
```diff
@@ -1563,3 +1581,28 @@ AmclNode::applyInitialPose()
...
+  diagnostic_status.add("std_y", std_y);
+  diagnostic_status.add("std_yaw", std_yaw);
+  diagnostic_status.add("max_std_x", max_std_x_);
+  diagnostic_status.add("max_std_y", max_std_y_);
+  diagnostic_status.add("max_std_yaw", max_std_yaw_);
+
+
+  if (std_x > max_std_x_ || std_y > max_std_y_ || std_yaw > max_std_yaw_)
+  {
+    diagnostic_status.summaryf(diagnostic_msgs::DiagnosticStatus::WARN, "Too large");
```
DLu:
```suggestion
    diagnostic_status.summary(diagnostic_msgs::DiagnosticStatus::WARN, "Too large");
```

DLu (commented):
This seems like a great change, but I would like to change the names of the variables. Rather than `max_std_x`, how would you feel about `cov_warn_level_x`. It's not great, and I'm open to suggestions. However, I feel like the current name doesn't reflect that the only thing it effects is the diagnostic status. Personally, I would have done `covar` instead of `cov` but we already have `initial_cov_xx`

reinzor:
For me the reason for specifying the standard deviation is that it is easier to understand the value you have to specify. What about `std_warn_level_x`, `std_warn_level_y`, `std_warn_level_yaw`?

DLu:
Apologies for the delay. Those names work for me.

reinzor:
Renamed to warn level

Issue #817, opened 2018-12-01
A small bug in map_grid.cpp


MichaelPan0905:
In base_local_planner/src/map_grid.cpp, line 150, here is a sentence to find "steps" which is used to smooth(I think) the global_plan we input, which is like this:
`int steps = ((sqrt(sqdist) - sqrt(min_sq_resolution)) / resolution) - 1;`
I think this sentence is wrong, it should be like this:
`int steps = ((sqrt(sqdist) - sqrt(min_sq_resolution)) / resolution) + 1;`
cause the "steps" tells us how many nodes we should add between two origin nodes in global_plan_input, and it should be positive. The old sentence may lead to less nodes that we add than we should, and in some extreme cases, like the sqrt(sqdist) equals 3 times of the resolution, it might carsh in the next two lines:
```
double deltax = (loop_x - last_x) / steps;
double deltay = (loop_y - last_y) / steps;
```
because the steps equals 0.

DLu:
Closed via #819

PR #819, opened 2018-12-06
[kinetic] Fix for adjusting plan resolution

DLu:
Possible fix for #817 

This does change the behavior such that one of the existing tests no longer works. However, I disagree with the results of that test, and it had some suspect parts anyway (i.e. checking the same x coordinate twice each, as opposed to x and y)

### DIFF
--- a/base_local_planner/src/map_grid.cpp
+++ b/base_local_planner/src/map_grid.cpp
@@ -139,15 +139,14 @@ namespace base_local_planner{
     double last_y = global_plan_in[0].pose.position.y;
     global_plan_out.push_back(global_plan_in[0]);
 
-    // we can take "holes" in the plan smaller than 2 grid cells (squared = 4)
-    double min_sq_resolution = resolution * resolution * 4;
+    double min_sq_resolution = resolution * resolution;
 
     for (unsigned int i = 1; i < global_plan_in.size(); ++i) {
       double loop_x = global_plan_in[i].pose.position.x;
       double loop_y = global_plan_in[i].pose.position.y;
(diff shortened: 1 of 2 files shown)


base_local_planner/src/map_grid.cpp
```diff
@@ -139,15 +139,14 @@ namespace base_local_planner{
...
-    double min_sq_resolution = resolution * resolution * 4;
+    double min_sq_resolution = resolution * resolution;
 
     for (unsigned int i = 1; i < global_plan_in.size(); ++i) {
       double loop_x = global_plan_in[i].pose.position.x;
       double loop_y = global_plan_in[i].pose.position.y;
       double sqdist = (loop_x - last_x) * (loop_x - last_x) + (loop_y - last_y) * (loop_y - last_y);
       if (sqdist > min_sq_resolution) {
-        int steps = ((sqrt(sqdist) - sqrt(min_sq_resolution)) / resolution) - 1;
+        int steps = ceil((sqrt(sqdist) - sqrt(min_sq_resolution)) / resolution) + 1;
```
aaronhoy: This took me a bit to parse.  Since sqrt(min_sq_resolution) == resolution, can't this expression be reduced to just:

`ceil(sqrt(sqdist) / resolution)`

??

DLu: You're totally right. Good call.

aaronhoy (approved):
LGTM.

DLu:
Build failure seems unrelated, but I can't figure out what is causing it. Any guesses @mikeferguson ?

mikeferguson:
Should this also go into lunar/melodic ?

DLu:
Also done.

Issue #794, opened 2018-09-17
move_base Warning: Invalid argument "/map" passed to canTransform argument target_frame


sirawats:
### **ros melodic , ubuntu 18.04**
I do roslaunch move_base.launch then it doesn't work with repeat these warning
```
Warning: Invalid argument "/map" passed to canTransform argument target_frame in tf2 frame_ids cannot start with a '/' like: 
         at line 134 in /tmp/binarydeb/ros-melodic-tf2-0.6.3/src/buffer_core.cpp
Warning: Invalid argument "/map" passed to canTransform argument target_frame in tf2 frame_ids cannot start with a '/' like: 
         at line 134 in /tmp/binarydeb/ros-melodic-tf2-0.6.3/src/buffer_core.cpp
Warning: Invalid argument "/map" passed to canTransform argument target_frame in tf2 frame_ids cannot start with a '/' like: 
         at line 134 in /tmp/binarydeb/ros-melodic-tf2-0.6.3/src/buffer_core.cpp
```
I have no idea what is cause this issue because I can run it on another PC and work well.
I didn't edit anything in Navigation package. Please help.
PS. amcl and slam gmapping is working fine.

so this is my .launch file
```
<?xml version="1.0" ?>
<launch>
  <!-- Arguments -->
  <arg name="model" default="$(env MECANION_MODEL)" doc="model type [mecanum]"/>
  <arg name="cmd_vel_topic" default="/cmd_vel" />
  <arg name="odom_topic" default="odom" />

  <!-- move_base -->
  <node pkg="move_base" type="move_base" respawn="false" name="move_base" output="screen">
    <param name="base_local_planner" value="dwa_local_planner/DWAPlannerROS" />
    <rosparam file="$(find mecanion_navigation)/param/costmap_common_params_$(arg model).yaml" command="load" ns="global_costmap" />
    <rosparam file="$(find mecanion_navigation)/param/costmap_common_params_$(arg model).yaml" command="load" ns="local_costmap" />
    <rosparam file="$(find mecanion_navigation)/param/local_costmap_params.yaml" command="load" />
    <rosparam file="$(find mecanion_navigation)/param/global_costmap_params.yaml" command="load" />
    <rosparam file="$(find mecanion_navigation)/param/move_base_params.yaml" command="load" />
    <rosparam file="$(find mecanion_navigation)/param/dwa_local_planner_params_$(arg model).yaml" command="load" />
[...]

PR #831, opened 2019-01-11
[melodic] Remove leading slashes from default frame_id parameters

DLu:
Possible solution for #794 

[TF2 Migration Guide](http://wiki.ros.org/tf2/Migration)


mikeferguson:
Huh, no idea why travis suddenly triggered. I'll take a closer look at that later, but yeah, this looks good!

PR #832, opened 2019-01-14
Support two modes in map_saver.

jingedawang:
The default mode of map_saver currently is "trinary". These commits support two modes in map_saver, "trinary" and "scale", the same in map_server.

The "scale" mode scales the value between "threshold_free_" and "threshold_occupied_", while transforming -1 to transparent pixel.

#### Features：
+ Support two modes - `trinary` and `scale`. Default is "trinary", with backward compatibility.
+ Unified namespace "map_server" in `map_server` and `map_saver` nodes.
+ Extract `MapMode` enum into `map_mode.h`, which can be reused in both `map_server` and `map_saver`.

---
Update in 2019.1.17:
+ Use libpng instead of OpenCV.
+ Fix a bug in `image_loader.cpp` when choose scale mode. `alpha` should be `double`, 0.0 for transparent, 1.0 for solid. So the value in alpha channel should be divided by 255.0 in advance.
+ Fix a bug in `image_loader.cpp`. `alpha` should be 1.0 when channels is less than 4.
+ Add `map_saver_test`.


map_server/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ find_package(catkin REQUIRED
 find_package(Bullet REQUIRED)
 find_package(SDL REQUIRED)
 find_package(SDL_image REQUIRED)
+find_package(OpenCV REQUIRED COMPONENTS core imgcodecs)
```
mikeferguson: is there some way to do this without bringing OpenCV -- that's a pretty heavy dependency for something as simple as the map saver?

jingedawang: Since `SDL_image` is unable to write images. `libpng` may be ok, but a little troublesome. So I choose the `OpenCV` library. OpenCV is included by ROS, so I think it is not a big problem. But I will listen to your suggestion.

jingedawang:
@mikeferguson I think the code works, but one of the checks still failed. Looks like some process out of time. I was confused. Can you help me check the reason ?

jingedawang:
Hi, can anybody help me finish it ? It should only need a little time to review.

jingedawang:
Excuse me, why is it hasn't been dealt with entil now ? @mikeferguson

DLu:
@jingedawang Please be patient. This is one of 37 PRs made against this repository and the maintainers have limited time.

In the meantime, please merge in the changes from the upstream melodic-devel. This includes dropping the travis-ci configuration, which should help the CI pass.

jingedawang:
@DLu Thanks for your advise. I have merged the new commits from upstream.

mikeferguson:

PR #835, opened 2019-01-15
[kinetic] Add close_polygon flag to polygonOutlineCells method to make it optional

corot:
The default behavior is the previous one: close polygons, so change will be transparent to existing code

### DIFF
--- a/costmap_2d/include/costmap_2d/costmap_2d.h
+++ b/costmap_2d/include/costmap_2d/costmap_2d.h
@@ -255,8 +255,10 @@ public:
    * @brief  Get the map cells that make up the outline of a polygon
    * @param polygon The polygon in map coordinates to rasterize
    * @param polygon_cells Will be set to the cells contained in the outline of the polygon
+   * @param close_polygon Close the polygon by going from the last point to the first (default behavior)
    */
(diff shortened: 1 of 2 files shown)


DLu:
@corot Is the use case where you want to use this when the "polygon" is not actually a polygon, but instead a path?

corot:
> @corot Is the use case where you want to use this when the "polygon" is not actually a polygon, but instead a path?

No, it's when the polygon you want to outline is partially outside of the costmap; without this flag, the method creates a fictitious line along the costmap border.

DLu:
I see what's going on...the method header does not include a standard polygon message, but instead requires you to pass in the grid coordinates of the inbounds portions of the polygon, because (I believe) the rayTrace is not smart enough to disinclude out of bounds cells. Could you instead filter the out of bounds MapLocations after this call? 

It's a weird method that doesn't seem to be used anywhere in this repo, so I'm trying to envision what use cases it might have. It has existed since before the git migration, so I couldn't find any good historical clues.

mikeferguson:

corot:
There's a [PR for melodic](https://github.com/ros-planning/navigation/pull/868). I'll open one for noetic, if the melodic one gets merged

PR #838, opened 2019-01-16
[melodic] Add parameters to tailor clear costmaps recovery behavior

corot:
- force_updating: force costmap update after clearing, so we don't need to wait for update thread (defaults to false)
- affected_maps: clear only local, global or both costmaps (defaults to both). Clearing local costmap is normally way faster to update, so this allows you to save a lot of time to recover from controller failures.

These changes are particularly useful when we use [move_base_flex](https://github.com/magazino/move_base_flex), as we can minimize the time used by the recovery.
Default behavior is the same as current one, so you won't notice anything until you set the parameters to a non-default value

### DIFF
--- a/clear_costmap_recovery/include/clear_costmap_recovery/clear_costmap_recovery.h
+++ b/clear_costmap_recovery/include/clear_costmap_recovery/clear_costmap_recovery.h
@@ -50,16 +50,16 @@ namespace clear_costmap_recovery{
     public:
       /**
        * @brief  Constructor, make sure to call initialize in addition to actually initialize the object
-       * @param  
-       * @return 
+       * @param
+       * @return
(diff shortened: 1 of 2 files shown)


doisyg:
Thanks for this PR, exactly what I needed with MBF. 
On Melodic, working with navigation from source, but not with the deb. Is there a way to know when it will be release/sync?

corot:
> On Melodic, working with navigation from source, but not with the deb. Is there a way to know when it will be release/sync?

First, the change must be released; easy to notice cause it will be a release commit (probably named 1.16.3) after this merge.
Then, u can change the current progress of the new release [here](http://repositories.ros.org/status_page/ros_melodic_default.html)

PR #839, opened 2019-01-17
[melodic] Speedup (~25%) inflation layer update

corot:
This is an follow up of PR #525 implemented by @MirkoFerrati at @Magazino.
As stated in that merged PR, the next biggest CPU eater is the map lookup for distances on enqueue method. This PR speeds up another ~25% by moving from map<double> to vector<>, using pre-cached integer distances.

It has been running on @Magazino robots for several months, so it's definitely well tested.

### DIFF
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -192,7 +192,6 @@ private:
   std::vector<std::vector<CellData>> inflation_cells_;
 
   bool* seen_;
-  bool* queued_;
   int seen_size_;
 
   unsigned char** cached_costs_;
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -168,6 +168,7 @@ private:
 
   void computeCaches();
   void deleteKernels();
+  int generateIntegerDistances();
   void inflate_area(int min_i, int min_j, int max_i, int max_j, unsigned char* master_grid);
 
   unsigned int cellDistance(double world_dist)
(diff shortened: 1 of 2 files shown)


corot:
Can we retake this PR?
It provides a nice performance boost and has been widely used for years, and also merged on ROS2 navigation. So is a safe bet.

matijazigic:
Hello @corot, first of all, thanks for sharing your code.
I tested your code and it worked very well but I noticed one problem. Here are pics:

Using your code:
inflation_with_changes

Current inflation code:
inflation_without_changes

Notice the bulge on rounded objects. 
I haven't had time yet to investigate where is the problem. Do you have any hints?

corot:
wow!  never noticed that...  Thanks for reporting! 
No, no clue. A first guess is that the are not inflating the cells in the right order (from closer to farther to the nearest obstacle). 
I'll try to investigate.

corot:
Found it. The guilty us the enqueued check. With:
image

without:
image

Price is a small decrease on performance, partially compensated by using emplace on adding new cells (I think for melodic we must still adhere to C++03, but there are already auto variables, so...).

Please pull the new version and give it a try!

corot:
:see_no_evil: I just noticed that [ROS2 navigation](https://github.com/ros-planning/navigation2/blob/main/nav2_costmap_2d/plugins/inflation_layer.cpp#L298) already has this PR with the bulges fixed...
I wonder why nobody worried to mention it here!

matijazigic:
Hello @corot, I'll give it a try later today. Tnx for your time. :clap:

SteveMacenski:
> I just noticed that ROS2 navigation already has this PR with the bulges fixed...

If I recall correctly, I was hounding for unit tests and they uncovered the issue Shrijit fixed. 

But in Nav2 when you’re ready to move over 😉

matijazigic:
Tested! Working as expected! :+1:

corot:
> But in Nav2 when you’re ready to move over

Looking forward to start using it... asa we move to ROS2!

mikeferguson:

PR #841, opened 2019-01-17
[kinetic] Fix install

DLu:
Working on getting tests to work...


mikeferguson:
@DLu -- I'd suggest we just remove the travis.yaml file -- travis hasn't been running for like 2 years, since ROS build farm added PR builders. I think something changed in the overall ors-planning org, and now it is firing again.

mikeferguson:
superceded by #843

DLu:
I'm reopening this, because the recovery test requires the files, but they are not installed, so it seems like they should be available.

mikeferguson:
It's not ideal that the tests use resources from other packages, but I guess that's OK since costmap_2d is mostly a library.

mikeferguson:
@DLu please forward port to melodic/lunar. Thx

Issue #842, opened 2019-01-18
Travis CI builds failing


DLu:
All of the Travis CI is failing to build because of two failed tests:

```
build/amcl/test_results/amcl/rostest-test_texas_greenroom_loop.xml: 1 tests, 1 errors, 0 failures, 0 skipped

build/clear_costmap_recovery/test_results/clear_costmap_recovery/rostest-test_clear_tests.xml: 1 tests, 1 errors, 0 failures, 0 skipped
```

I think I have a fix for the `clear_costmap_recovery` one in #841 but the other one is eluding me at the moment. I'm going to use this ticket to track the progress.

mikeferguson:
We actually don't want to use Travis -- We have the ROS build farm. I think someone changed something on the ros-planning org, and now travis is running. We should actually remove the travis.yaml -- since it hasn't been maintained since ROS farm actually supports PR builders.

PR #843, opened 2019-01-18
[indigo] Remove Travis

DLu:
Fix for #842


mikeferguson (approved):

mikeferguson:
I think this needs to be done in all active branches (especially the default one)

mikeferguson:
manually applied to melodic/kinetic

PR #846, opened 2019-01-24
[kinetic] Fix #845

### LINKED ISSUE #845: global_planner crashes when orientation mode is any of interpolation types
Hi,

ROS Kinetic, Ubuntu 16.04:

For the global_planner planning in the navigation stack for move_base, I am experiencing the following issue:

When an invalid (or inaccessible) goal is given to the planner with the orientation mode set to either None or Forward, the planner simply reports an error "Failed to get a plan." and continues its retries/resets, regardless of rest of the configuration.

However, with the same set of configuration, but with the orientation mode switched to either Interpolate or Forward then Interpolate, the planner reports the "Failed to get a plan." error once and then crashes the move_base node.

Thanks!

DLu:
Quick & Easy fix for #845

### DIFF
--- a/global_planner/src/orientation_filter.cpp
+++ b/global_planner/src/orientation_filter.cpp
@@ -54,6 +54,7 @@ void OrientationFilter::processPath(const geometry_msgs::PoseStamped& start,
                                     std::vector<geometry_msgs::PoseStamped>& path)
 {
     int n = path.size();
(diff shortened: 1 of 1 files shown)


mikeferguson:
Should this also go into lunar/melodic?

DLu:
Already done.

PR #849, opened 2019-01-31
[Windows][melodic] AMCL Windows build bring up.

seanyen:
* Add HAVE_UNISTD and HAVE_DRAND48 and portable_utils.hpp for better cross compiling.
* Variable length array is not supported in MSVC, conditionally disable it.
* Fix install location for shared lib and executables on Windows.
* Use isfinite for better cross compiling.

This change is verified against the https://aka.ms/ros project.


seanyen:
@DLu Hi! Hope I am pinging the correct maintainer. Can you take a look #849 #850 #851? Hope I get some feedbacks if it requires iterations.

PR #850, opened 2019-01-31
[Windows][melodic] map_server Windows build bring up

seanyen:
* Fix install location for Windows build. (On Windows build, shared library uses RUNTIME location, but not LIBRARY)
* Use boost::filesystem::path to handle path logic and remove the libgen.h dependency for better cross platform.
* Fix gtest hard-coded and add YAML library dir in CMakeList.txt.

This change is verified by https://aka.ms/ros project.


seanyen:
Thank for the merge!

PR #851, opened 2019-01-31
[Windows][melodic] Navigation (except for map_server and amcl) Windows build bring up

seanyen:
* Add HAVE_SYS_TIME_H for conditionally disable the code for the system which doesn't have this header and related APIs. (The part to be conditionally disabled are all debug code, so I think it's fine to disable them for Windows build.)
* Fix executable and shared lib install path for Windows.
* **__attribute__ ((unused))** is GCC specific attribute, MSVC doesn't have it so let's conditionally define it.
* Add #include <algorithm> to the header which uses std::min and std::max. Otherwise, MSVC will complain undefined functions.


seanyen:
I separate out map_server and AMCL changes here: (because it touches more code logic.)
#850 #849

seanyen:
@DLu @mikeferguson Hope I am pinging the correct owner. Let me know if anything I should iterate on. Thanks!

seanyen:
@DLu @mikeferguson Please excuse me to ping you again. Any chance I can move this PR forward? Thanks! Btw, this PR is also re-based to the latest. Hope that helps the review.

mikeferguson:
Retriggering CI (didn't run for some reason)

PR #858, opened 2019-02-09
[melodic] remove unused costmap_pub_frequency

IronClimber:
Remove unused param costmap_pub_frequency from global planner


DLu:
Thanks!

PR #867, opened 2019-02-16
moving clearing area method to costmap_layer so other applications ca…

SteveMacenski:
…n clear other types.

Essentially: imagine you have a costmap layer that has other data structures storing information than the `costmap_` char pointer. Moving the implementation of clearing an area of the costmap to the costmap itself lets someone override that function and implement the additional requirements to clear the other parts of the state. This is relevant for users of STVL utilizing the default clear costmaps recovery. The bug was reported and trying to get that all squared away before my next release

The API change was made such that it don't effect existing users or costmap_2d or clear costmaps recovery - the default behavior should work the same as before.

### DIFF
--- a/clear_costmap_recovery/src/clear_costmap_recovery.cpp
+++ b/clear_costmap_recovery/src/clear_costmap_recovery.cpp
@@ -135,19 +135,7 @@ void ClearCostmapRecovery::clearMap(boost::shared_ptr<costmap_2d::CostmapLayer> 
   costmap->worldToMapNoBounds(start_point_x, start_point_y, start_x, start_y);
   costmap->worldToMapNoBounds(end_point_x, end_point_y, end_x, end_y);
 
-  unsigned char* grid = costmap->getCharMap();
-  for(int x=0; x<(int)costmap->getSizeInCellsX(); x++){
-    bool xrange = x>start_x && x<end_x;
-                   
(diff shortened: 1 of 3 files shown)


mikeferguson:
@SteveMacenski this changes ABI. I'd welcome in Melodic-devel, but I think it's a bit late to be screwing with ABI in Kinetic. Would that work for your application?

SteveMacenski:
@mikeferguson it does change the API, but since its a virtual base and none of the plugins have that function implemented, it shouldn't change anything for users from a function point of view. I would plan to cherry pick to Melodic as well. 

A couple of folks in the STVL channels have been trying to use it with the base navigation stack utilizing the clear costmaps recovery behavior. Since it only clears out the `costmap_` and not the underlying data structure, the data will be repopulated in the next update cycle. The other option is to create a custom recovery behavior for just STVL but that doesn't seem very clean and it make sense to me on a high level that the costmap layer itself should probably be handling cleaning of data than an outside object comes in and scrubs it without notification.

Let me know if you think we could get this in for Kinetic or if you have suggestions

SteveMacenski:
@mikeferguson any suggestions or further thoughts? I understand the ABI concern but since its not a functional change and not effecting anything outside the package, it seems like a reasonable move to me

PR #868, opened 2019-02-20
[melodic] Add close_polygon flag to polygonOutlineCells method to make it optional

corot:
The default behavior is the previous one: close polygons, so change will be transparent to existing code


DLu:
Melodic version of #835 

@mikeferguson I always forget the exact rules...does this count as ABI breaking?

mikeferguson:

PR #871, opened 2019-02-26
[kinetic] Fix Bounds Bug (plus test)

DLu:
This is the commit from #632 with an added test. Without that commit, at least one of the tests fail.


DLu:
@ros-pull-request-builder

aaronhoy (approved):
LGTM

PR #881, opened 2019-03-27
suggesting adding back in robot_pose_ekf as iRobot Create 2 (which is…

OrBeProgrammed:
… currently being marketed and sold, and is used on some versions of Turtlebot 2 which is being released for melodic) relies on it.

### DIFF
--- a/robot_pose_ekf/CHANGELOG.rst
+++ b/robot_pose_ekf/CHANGELOG.rst
@@ -1,0 +1,103 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package robot_pose_ekf
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1.14.4 (2018-06-19)
+-------------------
+* Merge pull request `#713 <https://github.com/ros-planning/navigation/issues/713>`_ from garethellis0/kinetic-devel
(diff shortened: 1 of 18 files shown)


SteveMacenski:
You should look at [robot_localization](https://github.com/cra-ros-pkg/robot_localization). It will fit your needs and is more extensible. AFAIK there's been no reason to use robot_pose_ekf for awhile now. The defaults should be enough to give you relatively similar results as robot_pose_ekf. 

I also highly doubt any of the maintainers still use this package so maintaining it is challenging and redundant.

OrBeProgrammed:
Thanks I will look at the other package.

mikeferguson:
Yeah, we made a conscious effort to drop this in favor of robot_localization -- if you really want to release it, I suggest we create a separate repo for robot_pose_ekf on melodic+ and that you release from there.

I can certainly create a repo and add a team so you have commit rights on it -- let me know if you'd like to proceed.

OrBeProgrammed:
Actually that sounds great. There are some people using it for PR2 I saw on another thread somewhere so that would help there too.  That will let this robot function in melodic while I work on moving over to the new package you mentioned as time permits.

mikeferguson:
Repos have been created -- @k-okada has access to them, please reach out to him if you're looking to contribute

k-okada:
FYI released into melodic -> https://github.com/ros/rosdistro/pull/20794

PR #888, opened 2019-05-16
fix expected_update_rate_--->minimum_update_interval_

cf-zhang:
expected update rate is HZ , and the meaning in  isCurrent() is Time Duration.

### DIFF
--- a/costmap_2d/include/costmap_2d/inflation_layer.h
+++ b/costmap_2d/include/costmap_2d/inflation_layer.h
@@ -82,6 +82,8 @@ public:
     deleteKernels();
     if (dsrv_)
         delete dsrv_;
+    if (seen_)
+      delete[] seen_;
   }
 
   virtual void onInitialize();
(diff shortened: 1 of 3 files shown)


cf-zhang:
> expected update rate is HZ , and the meaning in isCurrent() is Time Duration.

```
~<name>/<source_name>/expected_update_rate (double, default: 0.0)

    How often to expect a reading from a sensor in seconds. A value of 0.0 will allow infinite time between 
readings. This parameter is used as a failsafe to keep the navigation stack from commanding the robot 
when a sensor has failed. It should be set to a value that is slightly more permissive than the actual rate of 
the sensor. For example, if we expect a scan from a laser every 0.05 seconds we might set this parameter 
to be 0.1 seconds to give a generous buffer and account for some amount of system latency. 

```
I went to wiki check it again, the description for `expected_update_rate` is namely the time Duration. I'm sorry abort this pull request, but I still think the name of this parameter is ambiguous.

mikeferguson:

PR #894, opened 2019-06-24
[melodic] On global_planner, add skip_poses param to optionally skip a number of intermediate poses

corot:
The idea is to make the resulting paths less dense (and so lighter and faster to process)


SteveMacenski:
If you are looking for less dense plans, is there a reason you dont prune them for your next use in an intermediate step or use a coarser costmap representation for planning?

corot:
> If you are looking for less dense plans, is there a reason you dont prune them for your next use in an intermediate step or use a coarser costmap representation for planning?

Prune the path before using is a very reasonable option, for sure. But sounds like a waste of time to compose a very dense path to then remove most of the poses. As for using a coarser costmap, the risk is to close narrow passages. And in any case, halving the resolution (in our cases, 0.05 instead of 0.025) still creates unnecessary dense paths.

All that said, I agree this skip poses trick sounds like a bit "artificial", so if not interesting for more people, I will drop and do a post-processing, as you suggest.

mikeferguson:

PR #896, opened 2019-06-24
[melodic] On global_planner, do not clear the starting cell in the costmap

corot:
I wonder what's the reason to do so. Planner seems to work just fine without it.

The reason I need to remove this is because sometimes I call SBPL lattice planner right after global_planner, and it creates paths that go into collision. The reason is that, because of this clearRobotCell method, the robot starts in a zero-cost cell, and SBPL assumes that no action can bring the robot into collision from here (as it assumes the closest obstacle is at at least at inflation_radius m away)

### DIFF
--- a/global_planner/include/global_planner/planner_core.h
+++ b/global_planner/include/global_planner/planner_core.h
@@ -176,7 +176,6 @@ class GlobalPlanner : public nav_core::BaseGlobalPlanner {
     private:
         void mapToWorld(double mx, double my, double& wx, double& wy);
         bool worldToMap(double wx, double wy, double& mx, double& my);
-        void clearRobotCell(const geometry_msgs::PoseStamped& global_pose, unsigned int mx, unsigned int my);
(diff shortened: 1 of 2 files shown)


DLu (changes requested):
This is my answer for a lot of things, but it is that way because [its always been that way](https://github.com/ros-planning/navigation/blob/melodic-devel/navfn/src/navfn_ros.cpp#L234). My impression is that certain robots would get sensor noise for the cell they were in, and this would enable them to succeed more often. 

Since this is disrupting your other workflow, I agree that this users should at least have the option to disable this. If you hide the behavior behind a param (defaulting to the current behavior) I'll merge it.

corot:
Or alternatively, I can just restore the costmap to its initial status, whatever you like the most.

DLu:
I think having a parameter to control whether you clear it or never touch it should sufficient.

corot:
Closed in favor of #897

PR #897, opened 2019-06-27
[melodic] Add 'do_not_alter_costmap' parameter that forces planner to not change costmap

corot:
More general version of PR #896, with the requested change of making the change not applicable by default. Instead of just removing the call to _clearRobotCell_, I offer now the option of not altering the costmap at all by copying its underlying char array and working only on it.

Another good thing of this approach is that we can choose (TODO in another PR) to lock the costmap only while copying it, so we allow updates during planning. This is useless on current move_base, but on [move_base_flex](https://github.com/magazino/move_base_flex) it's optional to lock the costmaps, delegating on the plugins to do so. This is a good thing for slow planners like sbpl that only invest a minimal time to copy the costmap at the beginning and then take very long while planning over his copy; on current move_base, all this time the costmap is frozen without any benefit.

### DIFF
--- a/global_planner/src/planner_core.cpp
+++ b/global_planner/src/planner_core.cpp
@@ -99,7 +99,7 @@ void GlobalPlanner::initialize(std::string name, costmap_2d::Costmap2D* costmap,
 
         unsigned int cx = costmap->getSizeInCellsX(), cy = costmap->getSizeInCellsY();
 
-        private_nh.param("do_not_alter_costmap", do_not_alter_costmap_, false);
+        private_nh.param("plan_on_costmap_copy", plan_on_costmap_copy_, false);
         private_nh.param("old_navfn_behavior", old_navfn_behavior_, false);
         if(!old_navfn_behavior_)
             convert_offset_ = 0.5;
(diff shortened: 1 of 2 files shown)


global_planner/src/planner_core.cpp
```diff
@@ -99,6 +99,7 @@ void GlobalPlanner::initialize(std::string name, costmap_2d::Costmap2D* costmap,
 
         unsigned int cx = costmap->getSizeInCellsX(), cy = costmap->getSizeInCellsY();
 
+        private_nh.param("do_not_alter_costmap", do_not_alter_costmap_, false);
```
DLu: I would prefer `plan_on_costmap_copy` to avoid the negation.

global_planner/src/planner_core.cpp
```diff
@@ -273,24 +273,34 @@ bool GlobalPlanner::makePlan(const geometry_msgs::PoseStamped& start, const geom
...
 
-    //clear the starting cell within the costmap because we know it can't be an obstacle
-    clearRobotCell(start, start_x_i, start_y_i);
-
     int nx = costmap_->getSizeInCellsX(), ny = costmap_->getSizeInCellsY();
 
+    unsigned char* costmap_char_array;
+    if(do_not_alter_costmap_){
+        //copy costmap's underlying char array so we can freely modify it (we clear robot cell and outline map borders)
+        costmap_char_array = new unsigned char[nx * ny];
```
DLu: I would move this up to the parent class so you don't realloc every iteration.

DLu (changes requested):

dorezyuk:
Hey guys, what the state of this PR? I think the idea not to alter the costmap is very right, however, I'm not sure if copying of the costmap is really necessary. Could we just write

```
    if(clear_robot_pose_)
        clearRobotCell(start, start_x_i, start_y_i);
    else if(costmap_->getCost(start_x_i, start_y_i) >= costmap_2d::INSCRIBED_INFLATED_OBSTACLE) {
        ROS_WARN_THROTTLE(1.0, "The robot's start position is in collision");
        return false;
    }
```
with a `clear_robot_pose_` parameter defaulting to true?

mikeferguson:

Issue #899, opened 2019-07-08
costmap_2d::StaticLayer-plugin of layered-costmap does not respect the maps coordinate frame in updateBounds() calculation


bitmeal:
This issue is filed for documentation purposes and helping other users find out about the problem and classifying their observations. A pull request solving the issue has been made: #900 

### scenario
The image below illustrates the following scenario:
* a map is served from a map-server and used in the `global_costmap` and `local_costmap`
* the `global_costmap`  and the `local_costmap` both use one single `costmap_2d::StaticLayer` exclusively
* `global_costmap` uses the `/map`-frame (bottom left)
* the `local_costmap` uses `/odom` as its `global_frame` (`global_frame: /odom`) and is a rolling window round the robot with frame `/base_link`

*To reproduce the following error, it is important that no other layer request another portion of the local costmap to be updated!*

![old_behav](https://user-images.githubusercontent.com/1504749/60840625-3bcbeb00-a1d0-11e9-8cc0-927ec33d33ef.png)

*The* real-world use-case *uses an external tracking- and observation-system using cameras, which generates robot poses (as tf), as well as obstacle information as a map-server compatible node. Therefore only `costmap_2d::StaticLayer`s with different resolutions and update rates are used.*

**The images here are taken from a simulation, using gazebo and a static map served by the *map_server*-node.**

### behavior
As can be observed, the local costmap (axis aligned to `/odom`-frame) is only filled with data in the *positive sector of the `/odom`-frame*. The areas of interest are highlighted with orange circles and extensions of the axes are show corresponding to their color (red, green) in the image below.

*Expected behavior would be: all areas visible as marked in the `global_costmap` to be marked in the `local_costmap`.*

Possible observations include:
* updates of the positive costmap-frames sector only
* updates only in specific regions

![old_behav_note_c](https://user-images.githubusercontent.com/1504749/60840654-48e8da00-a1d0-11e9-9697-e9ae36e078a0.png)

### the error
When updating a layered costmap, `costmap_2d::LayeredCostmap` asks all layers for the region they want to update in the next cycle. The region is fetched using the `updateBounds()`-method. `updateBounds` operates in *meters*, and specifies an *axis-aligned bounding-box* (`minx_, maxx_, miny_, maxy_`) to be updated, in the costmaps frame.

[...]

PR #900, opened 2019-07-08
fix bounds calculation of StaticLayer-plugin for maps served in different frame than the costmap

bitmeal:
This pull request addresses issue #899 .

The `costmap_2d::StaticLayer`-plugin does not respect the maps coordinate frame, and assumes it to be the same as the costmaps `global_frame`, when calculating the portion of the map to update.

This pull request fixes the issue and transforms the maps bounds in the frame of the costmap and calculates the corresponding axis-aligned bounding-box to server as the portion/bounds to be updated.

For further details, see #899

### DIFF
--- a/costmap_2d/plugins/static_layer.cpp
+++ b/costmap_2d/plugins/static_layer.cpp
@@ -278,15 +278,64 @@ void StaticLayer::updateBounds(double robot_x, double robot_y, double robot_yaw,
 
   useExtraBounds(min_x, min_y, max_x, max_y);
 
+  // static layer mapToWorld() uses map-server map's origin/frame, not layered map's origin/frame
+  // if rolling window potentially global_frame_ and map_frame_ will not match -> transform accordingly
   double wx, wy;
+  tf2::Transform tf2_transform_map_global;
+  std::vector<tf2::Vector3> map_corners;
 
-  mapToWorld(x_, y_, wx, wy);
-  *min_x = std::min(wx, *min_x);
-  *min_y = std::min(wy, *min_y);
+  if( map_frame_ != global_frame_ )
+  {
+    geometry_msgs::TransformStamped transform_map_global;
+    try
+    {
+      transform_map_global = tf_->lookupTransform(global_frame_, map_frame_, ros::Time(0));
+    }
+    catch (tf2::TransformException ex)
+    {
+      ROS_ERROR("%s", ex.what());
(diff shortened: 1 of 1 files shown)


bitmeal:
The test(s) (which I did not run before - sorry!) fail, because the submitted fix expects the `global_frame` of the costmap and the frame given as `frame_id` from the map_server to exist!
See following line from the tests ([link](http://build.ros.org/job/Mpr__navigation__ubuntu_bionic_amd64/2/consoleFull))
```23:56:38 [31m[ERROR] [1562622998.312606795]: "frame" passed to lookupTransform argument target_frame does not exist. [0m```
Furthermore, a transformation between these frames should exist (`map_frame_` and `global_frame_` in costmap_2d::StaticLayer).

The costmaps `global_frame` is set to `"frame"` in each *rostest-test_obstacle_tests* costmap constructors, whereas the map_server is serving a map with `frame_id` set to `"/map"`.
**The tests pass, when the map_servers `frame_id` is set to be `"frame"`** (equal to the given costmaps frame) - it is failing from a missing transformation/frame.

It is possible to *"fix the tests"* by providing the `frame_id` parameter as `_frame_id:=frame` to the map_server. **The problem ist, that the tests should describe and test for expected behavior.** Thus - per the defined tests - the expected behavior would be, that the costmap does not interpret the map_servers `frame_id` and does not transform the map data to the costmaps `global_frame_`, but simply copies it. There is another problem arising now: handling or *not-handling* of the map_servers `frame_id`, the `globla_frame_` and their relations/transformations, is inconsistent in `costmap_d2::StaticLayer`, as it differs from using `rolling_window: true` to `rolling_window: false`; when `rolling_window: true` the **map data is transformed** to the costmaps `global_frame_`. From the current implementation, **no expected behavior can be deduced!**

I will file another issue regarding the above-mentioned problem and hope for a discussion, on what the expected behavior should be. This should allow to define the expected behavior in the tests and provide a more robust implementation of `costmap_2d::StaticLayer`.

*Linked Issue: #903*

bitmeal:
Any chance to get (at least) some discussion on this?

bitmeal:
Bump again, as this is not trivial in my opinion.
Could we get some discussion on this and the linked Issues #899 and #903 ?
For visibility: @mikeferguson and @DLu maybe?

mikeferguson:

PR #905, opened 2019-07-22
Restrict max velocity of the robot based on the robot position

708yamaguchi:
Overall
=======
This Pull Request enables to restrict robots' max velocity based on the given map.
This function is effective in the places where humans and robots may collide.

Demo
====
Demo is as follows: (The robot slows down in front of the crossroads.)
max_velocity_map_floor

In the above gif animation, we show max velocity map in the left side. The robot reduces max velocity in the place where map pixel is painted black.
Also, we show current max velocity of the robot (`max_vel_x`) in the pie chart. When the robot goes into the black area, the max velocity reduces.
The line graph means the current speed of the robot (`cmd_vel_x`). You can see the robot's speed is reduced in the black area.

Note
====
In order to use this Pull Request, the max velocity map must be published in the form of `nav_msgs/OccupancyGrid`(This topic can be published by `map_server`), and must be given to `max_velocity_map_node` by remapping the topic name. (Please see the launch file)

You can create your original max velocity map by drawing the black area in the existing map (by some kind of painting tool).
Since `max_velocity_map_node` accepts gray images, gradation is effective to reduce the robot's speed step by step.

### DIFF
--- a/max_velocity_map/CMakeLists.txt
+++ b/max_velocity_map/CMakeLists.txt
@@ -1,0 +1,200 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(max_velocity_map)
+
+## Compile as C++11, supported in ROS Kinetic and newer
+# add_compile_options(-std=c++11)
+
+## Find catkin macros and libraries
+## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
+## is used, also find other catkin packages
+find_package(catkin REQUIRED COMPONENTS
+  amcl
+  geometry_msgs
+  dynamic_reconfigure
+  roscpp
+)
+
+## System dependencies are found with CMake's conventions
+# find_package(Boost REQUIRED COMPONENTS system)
+
+
+## Uncomment this if the package has a setup.py. This macro ensures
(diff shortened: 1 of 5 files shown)


mikeferguson:
A couple comments/questions (not really reviewed though):

- How does this actually affect the max velocity of the robot? It uses dynamic reconfigure?
- Looks like it is an entirely standalone package/node - I'd suggest if that is the case, you have it in a separate repository and release separately (since it isn't a core/required part of the navigation stack, and that will make it much easier for you to release updates at your own speed). There are several cost map layers and planners that exist independently from this repository and are maintained/released by their authors independently from the navigation stack.
- Lastly, the buildfarm for indigo is long gone, so you would absolutely have to target a newer release (I'd suggest Melodic).

708yamaguchi:
Thank you for your comment.

This node affects the max velocity of the robot by changing dynamic reconfigure.

As you said, this is standalone node. Now, I am thinking about releasing this package in our lab's repository.

Thank you.

mikeferguson:
Awesome! - when you've released the package -- be sure to announce it on ROS Discourse so that people know about it and can try it out.

PR #913, opened 2019-08-20
Exposing stopped_ variable in costmap2D_ros with the getter function isStopped

nirwester:
Knowing if the costmap is stopped can be of use for different components; for example, costmap converters could synchronously stop their execution, avoiding unnecessary computation.


mikeferguson:
Definitely changes ABI, suggest holding for noetic

PR #914, opened 2019-08-22
Rotate Recovery Refactor

DLu:
While looking at #826, I got upset at how convoluted the logic was for the rotate recovery. Too many negative angles and odd calculations. I think this version is much clearer. 

I also took the opportunity to clean things up. 
 * The [first](https://github.com/ros-planning/navigation/commit/af850c4e4711c2457c4522eac2a6b46130f3e095) two [commits](https://github.com/ros-planning/navigation/commit/e405a0cf8264d46cc2eb1659786f189b1d9bdc49) are just white space cleanup and linting. 
 * The [third commit](https://github.com/ros-planning/navigation/commit/b59ce8d1caaa629deafc8174ad836ab57e9402bb) removes unused components.
 * The [fourth commit](https://github.com/ros-planning/navigation/commit/4d28dab896871d2cc6df8c1ce45cee2b3954ad90) is the big one that changes the calculations to be clearer. 

@ebimor Can you check that this fixes your use case?

### DIFF
--- a/rotate_recovery/include/rotate_recovery/rotate_recovery.h
+++ b/rotate_recovery/include/rotate_recovery/rotate_recovery.h
@@ -39,11 +39,7 @@
 #include <nav_core/recovery_behavior.h>
 #include <costmap_2d/costmap_2d_ros.h>
 #include <tf/transform_listener.h>
-#include <ros/ros.h>
 #include <base_local_planner/costmap_model.h>
-#include <geometry_msgs/Twist.h>
-#include <geometry_msgs/Point.h>
-#include <angles/angles.h>
 #include <string>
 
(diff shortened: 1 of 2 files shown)


aaronhoy (approved):
Seems reasonable.  Interesting, so we were never using the global costmap?

SteveMacenski:
This is computing a velocity command for the base, I wouldn't think it would use the global costmap, what would be the benefit?

aaronhoy:
@SteveMacenski I agree.  I was just noting that prior to this PR, the global costmap was brought in but not used.  The state of things with this PR makes sense.

ebimor:
@DLu  looks good to me! Thanks

PR #921, opened 2019-09-03
[kinetic] [AMCL] Feature: Selective Resampling

adi-vardi:
- Implement selective resampling as described in:

> Grisetti, Giorgio, Cyrill Stachniss, and Wolfram Burgard. "Improved techniques for grid mapping with rao-blackwellized particle filters." IEEE transactions on Robotics 23.1 (2007): 34.

- The main advantage is reducing the resampling rate when not needed and help avoid particle deprivation.

- Compute `n_effective` for current set as `N_eff = 1/(sum(k_i^2))`.
When `n_effective` falls below `N/2` (`N` being the current number of samples), continue to resampling. Otherwise, copy current set to the other set and skip the resampling (switching the set is needed, otherwise the cluster KD tree is not defined properly and it is not possible to compute the clusters' stats)

- When resampling is skipped, recompute cluster stats using `pf_cluster_stats()` in order to still have a result. The result is consistent, as all the samples are propagated using the motion model.

- Additional parameter to enable/disable selective resampling: `selective_resampling` (default: false)

### DIFF
--- a/amcl/include/amcl/pf/pf.h
+++ b/amcl/include/amcl/pf/pf.h
@@ -104,6 +104,7 @@ typedef struct _pf_sample_set_t
   pf_vector_t mean;
   pf_matrix_t cov;
   int converged; 
+  double n_effective;
 } pf_sample_set_t;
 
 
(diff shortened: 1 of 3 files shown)


mikeferguson:
Seems completely reasonable, although I did not fully check the math. It is completely hidden behind a feature flag, so no impact on current users.

Issue #924, opened 2019-09-24
static_layer cause costmap wrong when map size is changed


tongsky723:
When I used gmapping to create map while move base is running. Costmap became wrong when map size is changed.
As you can see from following pictures. Second one costmap became wrong when map is resized!
 
![2019-09-24 16-11-01 的屏幕截图](https://user-images.githubusercontent.com/28940801/65493812-2f5d5180-dee6-11e9-81cf-416563117c9f.png)
![2019-09-24 16-11-14 的屏幕截图](https://user-images.githubusercontent.com/28940801/65493818-32f0d880-dee6-11e9-8baf-863ff6f64102.png)

mikeferguson:
I believe this is a duplicate of #959

PR #792, opened 2018-09-07
Changed logic for when to resize layered costmap in static layer

ghost:
-Now the master layered costmap should no longer get resized by static layer when
isSizeLocked returns true

fixes #768

PR #925, opened 2019-09-24
[Kinetic] Fix costmap became wrong when map size is changed

tongsky723:
Fix costmap became wrong when map size is changed. As described in issue #924


mechwiz:
Thanks for creating the PR.
The problem is also described in detail here #792

tongsky723:
Thanks for let me know! I will close this PR because others already suggested the same solution.

mechwiz:
You can leave it open if you'd like. I just wanted to reference the other issue so that the reviewers can see more info regarding the problem. The other issue is closed so maybe this one has a higher chance of being seen if it remains open.

PR #929, opened 2019-10-03
[clear_costmap_recovery] add param invert_area_to_clear

doisyg:
I wanted the possibility to clear obstacles up to a specific distance from the robot, and not from a specific distance until the infinity. In other word to invert the area to clear in the costmap.
I also clarified what is going on through the warning message and the fact that the area cleared is squared based and not circle based as specified in the wiki : "by reverting to the static map outside of a given radius away from the robot"


mikeferguson:
Repo has disappeared, and there is a conflict so I can't merge.

Issue #784, opened 2018-08-30
exist numeric overflow problem in function "void AStarExpansion::add(....)"


chenxianbo:
```
void AStarExpansion::add(unsigned char* costs, float* potential, float prev_potential, int next_i, int end_x,  int end_y)
{
    .........
    if(costs[next_i]>=lethal_cost_ && !(unknown_ && costs[next_i]==costmap_2d::NO_INFORMATION))
        return;
    potential[next_i] = p_calc_->calculatePotential(potential, costs[next_i] + neutral_cost_, next_i, prev_potential);
     .......
}
```
when `neutral_cost_ == 50, costs[next_i] >= 206, costs[next_i] + neutral_cost_` will numeric overflow

PR #930, opened 2019-10-08
Capped the costs to 255 so that no UInt8 Overflow occurs

trickarcher:
This pull request fixes the Issue #784 . Added an additional variable of `uint8_t` which caps the cost to 255 to prevent any numeric overflow,


mikeferguson:

PR #936, opened 2019-10-10
Fixes gdist- and pdist_scale node paramter names

### LINKED ISSUE #935: Dynamic reconfigure pdist_scale and gdist_scale is not applied in local planner
This is due to lookup of wrong names in the parameter server.
I'm preparing a pull request to fix this.

DavidPL1:
Fixes #935. The parameter names are simply changed to fit those defined in the cfg files of BaseLocalPlanner.

### DIFF
--- a/base_local_planner/src/trajectory_planner_ros.cpp
+++ b/base_local_planner/src/trajectory_planner_ros.cpp
@@ -102,7 +102,7 @@ namespace base_local_planner {
       trans_stopped_velocity_ = 1e-2;
       double sim_time, sim_granularity, angular_sim_granularity;
       int vx_samples, vtheta_samples;
-      double pdist_scale, gdist_scale, occdist_scale, heading_lookahead, oscillation_reset_dist, escape_reset_dist, escape_reset_theta;
+      double path_distance_bias, goal_distance_bias, occdist_scale, heading_lookahead, oscillation_reset_dist, escape_reset_dist, escape_reset_theta;
       bool holonomic_robot, dwa, simple_attractor, heading_scoring;
       double heading_scoring_timestep;
       double max_vel_x, min_vel_x;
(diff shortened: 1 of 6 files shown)


mikeferguson:
This will break a lot of robot configurations out there. It would seem the proper fix is probably to update the .cfg file to use these names, since that was not working previously anyways.

DavidPL1:
Yes, for those who figured out which parameters to use it would break the configuration. Didn't think about that :D

I guess I could adapt this PR to change the names in the cfg file and try to spread awareness among the movebase users I know. I noticed for instance that the public TIAGo navigation config by PAL Robotics uses the wrong names...

mikeferguson:
This does not build - needs changes in the configureCB (see build errors)

DavidPL1:
Yes, I know. Sorry for the delay, had to finish my thesis first :sweat_smile: 
But I'm on it now and running a few tests first and checking for other parameters with the same problem.

DavidPL1:
Okay, so the latest commits should do the trick.
Since, if one still uses `pdist_scale` or `gdist_scale` as parameters in their config, they would be overridden by the dynamic reconfigure defaults, the values are copied over to the ones used by dynamic reconfigure (see [here](https://github.com/DavidPL1/navigation/blob/102140c3c76ea399f1bd364eaa4c92b0aa7b32d8/base_local_planner/src/trajectory_planner_ros.cpp#L177)).

In my tests both variants for setting the parameters worked and in both cases they could be changed using dynamic reconfigure.

I also noticed a problem with `max_vel_theta` and `min_vel_theta` when using `max_rotational_vel`: this value is parsed correctly but will instantly be overridden by the dynamic_reconfigure parameter defaults of `min_vel_theta` and `max_vel_theta`, as to dynamic reconfigure they seem to be undefined.
I have not fixed this yet, but this could be done similar to the above solution.

base_local_planner/src/trajectory_planner_ros.cpp
```diff
@@ -166,8 +166,25 @@ namespace base_local_planner {
...
+                                                                  "path_distance_bias",
+                                                                  "pdist_scale",
+                                                                  0.6);
+      goal_distance_bias = nav_core::loadParameterWithDeprecation(private_nh,
+                                                                  "goal_distance_bias",
+                                                                  "gdist_scale",
+                                                                  0.6);
+      // values of the deprecated params need to be applied to the current params, as defaults 
+      // of defined for dynamic reconfigure will override them otherwise.
+      if (private_nh.hasParam("pdist_scale"))
```
mikeferguson: should this logic actually have "& !private_nh.hasParam("path_distance_bias")" -- I presume we only want to do this once

DavidPL1: Isn't this executed just once? On node restart the path_distance_bias parameter is deleted if it is not set in the config file, so this needs to be done on every node start. Or do you mean something else?

~However, I should add your suggestion for consistency in case of both variables being defined in the config file, because the `loadParameterWithDeprecationWarning` call also favours path_distance_bias over pdist_scale.~
Forget this last part, the param is updated with the correct value regardless if both are defined in the config. So yes, if both are defined the `setParam` call is redundant.

Issue #939, opened 2019-10-27
Dijskstra not planning straight path in empty map


phcapde:
The dijkstra impementation in an empty map does not plan straight lines depending on the direction of the path. Vertical paths are ok, but close to horizontal have step-like shapes as illustrated by the potential and path below:
![image](https://user-images.githubusercontent.com/10817916/67632195-d136dc00-f8a0-11e9-9c82-2f624dbaad8f.png)

The bug is due to a bias in the horizontal/vertical cells update mechanism, in which horizontals are always pushed back on the queue before verticals. The provided fix simply alternates between pushing first horizontal or vertical to average out the bias. Long paths can still be slightly curved but no more "step-like" path are generated. A resulting path is the following:
![image](https://user-images.githubusercontent.com/10817916/67632233-56ba8c00-f8a1-11e9-839a-fdf866a64dc5.png)

dronefreak:
Hi,

Thanks for the fix.

PR #940, opened 2019-10-27
fixed dijkstra horizontal/vertical bias

phcapde:
as described in issue #939


VicenteMovai:
This would be a nice bugfix to release, I tested it and it works

mikeferguson:

PR #945, opened 2019-11-29
remove signals dep

acxz:
Boost > 1.7 has signals by default

### DIFF
--- a/fake_localization/CMakeLists.txt
+++ b/fake_localization/CMakeLists.txt
@@ -14,7 +14,7 @@ find_package(catkin REQUIRED
         )
 
 
-find_package(Boost REQUIRED COMPONENTS signals)
+find_package(Boost REQUIRED)
 
 catkin_package(
     CATKIN_DEPENDS


jwhendy:
Thought I'd look around at this in case it helps decide what to do.

I don't know that the boost 1.7+ justification can be used, as `melodic` caters to `bionic` and `bionic` is [on 1.65](https://packages.ubuntu.com/bionic/libboost-all-dev).

That said, there is precedence for this, and perhaps it's even "proper" to remove `signals`:
- this `geometry2/tf2` [PR](https://github.com/ros/geometry2/pull/354) removes it, stating that 1) it's `signals2`, which isn't the same `signals` pulled in by this `find_package()` call, and 2) that `signals2` doesn't need to be found via `find_pacakge()`
- there's an [identical PR](https://github.com/ros/geometry/pull/180) for `geometry/tf`
- `moveit` also [followed suit](https://github.com/ros-planning/moveit/pull/1353), as did [roscpp](https://github.com/ros/ros_comm/pull/1580)
- lastly, it's not clear if `signals` is actually used here? I [found it](https://github.com/ros/geometry/blob/9e73fb4b919b86b488ed8abf68a6b060d0b7f519/tf/include/tf/message_filter.h#L112) in `tf`, but I get no hits in this package. I only find it in `ros-planning/navigation` [here](https://github.com/ros-planning/navigation/blob/970eefe147c4f168281b3ca3e378a81baf350300/amcl/src/amcl_node.cpp#L33), yet `amcl` doesn't even have it in [CMakeLists.txt](https://github.com/ros-planning/navigation/blob/970eefe147c4f168281b3ca3e378a81baf350300/amcl/CMakeLists.txt#L8).

I don't really understand cpp, nor boost nor signals, but hoped to point in a few helpful directions for the maintainers here.

lopsided98:
Boost signals (v1) was [removed in 1.69](https://www.boost.org/users/history/version_1_69_0.html):
> Signals (v1) is now removed. Its removal was announced in 1.68 and its deprecation was announced in 1.54. Boost 1.68 is the last release that provides this library. Users are encouraged to use Signals2 instead. The Boost community thanks Douglas Gregor for his work on Signals which served its users well and which also inspired Signals2.

Because of this, the `fake_localization` package fails to build on Boost 1.69+. The fact that this package builds successfully with 1.69+ after this PR is applied shows that it does not actually depend on Boost signals.

lopsided98:
> I only find it in `ros-planning/navigation` [here](https://github.com/ros-planning/navigation/blob/970eefe147c4f168281b3ca3e378a81baf350300/amcl/src/amcl_node.cpp#L33), yet `amcl` doesn't even have it in [CMakeLists.txt](https://github.com/ros-planning/navigation/blob/970eefe147c4f168281b3ca3e378a81baf350300/amcl/CMakeLists.txt#L8).

That is not Boost signals, that is the [POSIX `signal.h` header](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html).

acxz:
@mikeferguson or anyone else that has the power to, can you look at this PR? It fixes the ROS package on Arch Linux, and this fix will be needed eventually for all distros.

Issue #885, opened 2019-04-17
footprint_pub and footprint_sub are different message types on same topic


mkhansenbot:
@DLu or @mikeferguson - While running in ROS2 we've noticed that the "footprint" topic has a different publisher message type (PolygonStamped) than subscriber type (Polygon). This doesn't work at all with the Opensplice RMW layer implementation (it causes the costmap nodes to crash). Looking at the code, I don't understand why the two are different. I tried changing them to the same (PolygonStamped) but that caused the costmap inflations to be wrong. 

Could you explain the intent of the publisher and subscriber being different message types on the same topic? We may need a different way to do this in ROS2. 
cc: @crdelsey

mikeferguson:
Can you give a pointer to where the subscriber is created? I'm not sure what that is all about.

The publisher I believe is mainly for visualization purposes -- so I don't think they are intended to be connected together (which sounds like what you are trying to do?)

mkhansenbot:
The subscriber is created here:
https://github.com/ros-planning/navigation/blob/b52e0ea9362131dc81326da471792ab8698c9564/costmap_2d/src/costmap_2d_ros.cpp#L138
```
 // subscribe to the footprint topic
  std::string topic_param, topic;
  if (!private_nh.searchParam("footprint_topic", topic_param))
  {
    topic_param = "footprint_topic";
  }

  private_nh.param(topic_param, topic, std::string("footprint"));
  footprint_sub_ = private_nh.subscribe(topic, 1, &Costmap2DROS::setUnpaddedRobotFootprintPolygon, this);

  if (!private_nh.searchParam("published_footprint_topic", topic_param))
  {
    topic_param = "published_footprint";
  }

  private_nh.param(topic_param, topic, std::string("oriented_footprint"));
  footprint_pub_ = private_nh.advertise<geometry_msgs::PolygonStamped>("footprint", 1);
```
Then later the callback is declared, note that it is expecting a Polygon msg, not PolygonStamped:
```
void Costmap2DROS::setUnpaddedRobotFootprintPolygon(const geometry_msgs::Polygon& footprint)
{
  setUnpaddedRobotFootprint(toPointVector(footprint));
}
```
So I'm not sure what the intent is here, are they supposed to be on 2 separate topics?

mikeferguson:
> So I'm not sure what the intent is here, are they supposed to be on 2 separate topics?

[...]

PR #947, opened 2019-12-17
fix published footprint topic name

furushchev:
It is assumed that the topic name used for publishing current footprint is customizable by getting the name from rosparam, but the value is not used and instead fixed name 'footprint' is used.
This PR fixes to be used the name set using rosparam.

The issue also occurs mixture of topic type between `geometry_msgs/Polygon` and `geometry_msgs/PolygonStamped`.  
The related issue: https://github.com/ros-planning/navigation/issues/885

### DIFF
--- a/costmap_2d/src/costmap_2d_ros.cpp
+++ b/costmap_2d/src/costmap_2d_ros.cpp
@@ -155,7 +155,7 @@ Costmap2DROS::Costmap2DROS(const std::string& name, tf2_ros::Buffer& tf) :
   }
 
   private_nh.param(topic_param, topic, std::string("oriented_footprint"));
-  footprint_pub_ = private_nh.advertise<geometry_msgs::PolygonStamped>("footprint", 1);
+  footprint_pub_ = private_nh.advertise<geometry_msgs::PolygonStamped>(topic, 1);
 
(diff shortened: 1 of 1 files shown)


SteveMacenski:
While I don't disagree with your proposal, you're also changing the default behavior since a non-defined `footprint_pub_` would publish on `footprint` and now it will on `oriented_footprint`. I think that deserves some kind of documentation / warning / notification (or change on a distribution update) just do you don't break users that didn't override the parameter

furushchev:
@SteveMacenski Totally agreed. Or should it be better to change the default value `oriented_footprint` to `footprint` and also change the default value of topic for `footprint_sub_` to, for instance, `current_footprint` not to use the same topic name? I thought it because users of `footprint_sub_` have to remap the default name `footprint` to other name in order to subscribe updated footprint polygon, on the other hand, subscribing messages from `footprint_pub_` for visualization works without any remapping. What do you think of it?

mikeferguson:
Here's my thought: let's change the default parameter value to "footprint" for melodic, yes, by defaut these will still be on the same topic (bad) but it won't change the behavior for anyone (good) and it will allow you to properly change the name via the param (good). For N-turtle, we can change the default parameter value back to "oriented_footprint" and note that in the release notes.

PR #948, opened 2019-12-19
Use const for getter methods

furushchev:
To safely manage costmap object in subclasses.


mikeferguson:
I can't find a definitive answer as to whether adding const to a function breaks ABI or not. If you happen to have something that shows that ABI does not change, we can merge this, otherwise I would suggest we hold this for noetic.

furushchev:
@mikaelarguedas Thank you for the feedback! Please hold for noetic since this is just a safety enhancement feature.

PR #956, opened 2020-01-10
Slow control loop with dwa_local_planner (melodic)

pavloblindnology:
Fix for #679. Replication of #688 for melodic.

### DIFF
--- a/base_local_planner/include/base_local_planner/goal_functions.h
+++ b/base_local_planner/include/base_local_planner/goal_functions.h
@@ -102,6 +102,7 @@ namespace base_local_planner {
       const geometry_msgs::PoseStamped& global_robot_pose,
       const costmap_2d::Costmap2D& costmap,
       const std::string& global_frame,
+      const ros::Time& time,
       std::vector<geometry_msgs::PoseStamped>& transformed_plan);
 
   /**
@@ -115,6 +116,7 @@ namespace base_local_planner {
   bool getGoalPose(const tf2_ros::Buffer& tf,
       const std::vector<geometry_msgs::PoseStamped>& global_plan,
       const std::string& global_frame,
+      const ros::Time& time,
       geometry_msgs::PoseStamped &goal_pose);
 
   /**
@@ -134,6 +136,7 @@ namespace base_local_planner {
       const std::vector<geometry_msgs::PoseStamped>& global_plan,
       const costmap_2d::Costmap2D& costmap,
       const std::string& global_frame,
+      const ros::Time& time,
       geometry_msgs::PoseStamped& global_pose,
       const nav_msgs::Odometry& base_odom,
       double rot_stopped_vel, double trans_stopped_vel,
--- a/base_local_planner/src/goal_functions.cpp
+++ b/base_local_planner/src/goal_functions.cpp
@@ -111,8 +111,7 @@ namespace base_local_planner {
     const geometry_msgs::PoseStamped& plan_pose = global_plan[0];
     try {
       // get plan_to_global_transform from plan frame to global_frame
-      geometry_msgs::TransformStamped plan_to_global_transform = tf.lookupTransform(global_frame, time,
-          plan_pose.header.frame_id, plan_pose.header.stamp, plan_pose.header.frame_id, ros::Duration(0.5));
+      geometry_msgs::TransformStamped plan_to_global_transform = tf.lookupTransform(global_frame, plan_pose.header.frame_id, time, ros::Duration(0.5));
 
       //let's get the pose of the robot in the frame of the plan
       geometry_msgs::PoseStamped robot_pose;
@@ -183,9 +182,7 @@ namespace base_local_planner {
 
     const geometry_msgs::PoseStamped& plan_goal_pose = global_plan.back();
     try{
-      geometry_msgs::TransformStamped transform = tf.lookupTransform(global_frame, time,
-                         plan_goal_pose.header.frame_id, plan_goal_pose.header.stamp,
-                         plan_goal_pose.header.frame_id, ros::Duration(0.5));
+      geometry_msgs::TransformStamped transform = tf.lookupTransform(global_frame, plan_goal_pose.header.frame_id, time, ros::Duration(0.5));
 
       tf2::doTransform(plan_goal_pose, goal_pose, transform);
     }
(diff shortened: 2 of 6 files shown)


pavloblindnology:
Any updates on this?

base_local_planner/include/base_local_planner/goal_functions.h
```diff
@@ -102,6 +102,7 @@ namespace base_local_planner {
       const geometry_msgs::PoseStamped& global_robot_pose,
       const costmap_2d::Costmap2D& costmap,
       const std::string& global_frame,
+      const ros::Time& time,
```
mikeferguson, line 105: I've spent quite a bit of time looking at this now - I really don't like littering extra params all over, and having all this book-keeping going on as to what time we are using for things (it seems like this will be very easy to use incorrectly).

base_local_planner/src/goal_functions.cpp
```diff
@@ -105,8 +106,7 @@ namespace base_local_planner {
     const geometry_msgs::PoseStamped& plan_pose = global_plan[0];
     try {
       // get plan_to_global_transform from plan frame to global_frame
-      geometry_msgs::TransformStamped plan_to_global_transform = tf.lookupTransform(global_frame, ros::Time(),
-          plan_pose.header.frame_id, plan_pose.header.stamp, plan_pose.header.frame_id, ros::Duration(0.5));
+      geometry_msgs::TransformStamped plan_to_global_transform = tf.lookupTransform(global_frame, plan_pose.header.frame_id, time, ros::Duration(0.5));
```
mikeferguson, line 114: here, it seems the correct thing to use is not a passed in time, but rather the timestamp of global_pose (which will usually have just been determined by a call to getRobotPose().

base_local_planner/src/goal_functions.cpp
```diff
@@ -177,9 +177,7 @@ namespace base_local_planner {
 
     const geometry_msgs::PoseStamped& plan_goal_pose = global_plan.back();
     try{
-      geometry_msgs::TransformStamped transform = tf.lookupTransform(global_frame, ros::Time(),
-                         plan_goal_pose.header.frame_id, plan_goal_pose.header.stamp,
-                         plan_goal_pose.header.frame_id, ros::Duration(0.5));
+      geometry_msgs::TransformStamped transform = tf.lookupTransform(global_frame, plan_goal_pose.header.frame_id, time, ros::Duration(0.5));
```
mikeferguson, line 185: why is ros::Time() a problem here? that should simply give us the latest transform - we don't even appear to be ever using the timestamp of the goal_pose returned, so?

base_local_planner/src/trajectory_planner_ros.cpp
```diff
@@ -387,7 +387,7 @@ namespace base_local_planner {
 
     std::vector<geometry_msgs::PoseStamped> transformed_plan;
     //get the global plan in our frame
-    if (!transformGlobalPlan(*tf_, global_plan_, global_pose, *costmap_, global_frame_, transformed_plan)) {
+    if (!transformGlobalPlan(*tf_, global_plan_, global_pose, *costmap_, global_frame_, ros::Time::now(), transformed_plan)) {
```
mikeferguson, line 409: and here, now() is later than the stamp we just got on the global_pose, so again, it would seem that approach would be reasonable.

dwa_local_planner/src/dwa_planner_ros.cpp
```diff
@@ -263,10 +272,6 @@ namespace dwa_local_planner {
 
   bool DWAPlannerROS::computeVelocityCommands(geometry_msgs::Twist& cmd_vel) {
     // dispatches to either dwa sampling control or stop and rotate control, depending on whether we have been close enough to goal
-    if ( ! costmap_ros_->getRobotPose(current_pose_)) {
-      ROS_ERROR("Could not get robot pose");
-      return false;
-    }
```
mikeferguson, line 269: while we currently call isGoalReached() right before computeVelocityCommands(), I really don't think we want to make this API harder to use -- using the timestamp of the current_pose_ rather than an extra timestamp allows this to stay and avoids extra bookkeeping.

mikeferguson (changes requested):
I've finally spent an appropriate amount of time looking this one (and the kinetic version) over -- I see the issues you're trying to fix, and very much appreciate your attempt to fix them - but I'm really not thrilled with the extra bookkeeping -- I'm convinced there has to be a better/simpler approach. Please let me know what you think of the proposed alternatives.

pavloblindnology:
@mikeferguson The whole idea was to keep all transformations done during one local planner cycle consistent. I agree it can be a little bit confusing. I see your point and agree that this consistent time is better to be the latest robot pose's time, though it can be quite in the past (tens of ms). This way we know that we're using the same time when making computations based on both robot's pose and transformed global plan. The only inconsistency left is goal computations, which, if using `ros::Time()` instead of book-kept time are not tied to robot's pose timestamp. Though goal computations seem less important than transformed plan computations, we still can fix it by passing robot's pose timestamp to `LocalPlannerUtil::getGoal` and  `goal_functions::getGoalPose`:
https://github.com/ros-planning/navigation/blob/9368db28b2d60e86ad1d55d4ae91b5f54a235a5c/base_local_planner/src/local_planner_util.cpp#L83
and
https://github.com/ros-planning/navigation/blob/9368db28b2d60e86ad1d55d4ae91b5f54a235a5c/base_local_planner/src/goal_functions.cpp#L174
BTW, this will be consistent with `TrajectoryPlannerROS`, which has its transformed_plan and goal timestamps consistent.

base_local_planner/src/goal_functions.cpp
pavloblindnology: Agree, it'll be more consistent, though global_pose's timestamp can be quite in the past (tens of ms).

base_local_planner/src/goal_functions.cpp
pavloblindnology: We don't use the timestamp of returned goal_pose, but the timestamp is used to define transformation itself. And it's better to be consistent with the rest of computations. That is why I still propose to pass timestamp here and pass robot's pose timestamp (in LatchedStopRotateController).
And we can define this timestamp argument's default value to be ros::Time() in order not to break other local planners.

dwa_local_planner/src/dwa_planner_ros.cpp
pavloblindnology: Agree.

mikeferguson:

PR #962, opened 2020-01-26
Global planning using A* algorithm with GradientPath method

MikeDegany:
There were some errors while using A* algorithm and the GradientPath method together. I made some minor changes to ignore them all. And also Manhattan distance heuristic changed to Euclidean distance. It could be useful for some future developers who want to plan a global path using both methods together. Thanks

### DIFF
--- a/global_planner/src/astar.cpp
+++ b/global_planner/src/astar.cpp
@@ -89,7 +89,9 @@ void AStarExpansion::add(unsigned char* costs, float* potential, float prev_pote
 
     potential[next_i] = p_calc_->calculatePotential(potential, costs[next_i] + neutral_cost_, next_i, prev_potential);
     int x = next_i % nx_, y = next_i / nx_;
-    float distance = abs(end_x - x) + abs(end_y - y);
+
+    //float distance = abs(end_x - x) + abs(end_y - y);
+    float distance = sqrt((end_x - x)*(end_x - x) + (end_y - y)*(end_y - y));	//Euclidean Distance
 
     queue_.push_back(Index(next_i, potential[next_i] + distance * neutral_cost_));
     std::push_heap(queue_.begin(), queue_.end(), greater1());
--- a/global_planner/src/gradient_path.cpp
+++ b/global_planner/src/gradient_path.cpp
@@ -81,7 +81,8 @@ bool GradientPath::getPath(float* potential, double start_x, double start_y, dou
         // check if near goal
         double nx = stc % xs_ + dx, ny = stc / xs_ + dy;
 
-        if (fabs(nx - start_x) < .5 && fabs(ny - start_y) < .5) {
+        //if (fabs(nx - start_x) < .5 && fabs(ny - start_y) < .5) {
+        if (fabs(nx - start_x) < 1 && fabs(ny - start_y) < 1) {		//To be able to work with A* algorithm and GradientPath method
             current.first = start_x;
             current.second = start_y;
             path.push_back(current);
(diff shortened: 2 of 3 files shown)


global_planner/src/planner_core.cpp
```diff
@@ -113,7 +113,7 @@ void GlobalPlanner::initialize(std::string name, costmap_2d::Costmap2D* costmap,
             p_calc_ = new PotentialCalculator(cx, cy);
 
         bool use_dijkstra;
-        private_nh.param("use_dijkstra", use_dijkstra, true);
+        private_nh.param("use_dijkstra", use_dijkstra, false);
```
DLu, line 116: Generally, we try to avoid changing any of the default parameter values so you don't create unexpected behavior on someone else's system. For your use case, this could just be reconfigured at launch, right?

global_planner/src/astar.cpp
```diff
@@ -89,7 +89,9 @@ void AStarExpansion::add(unsigned char* costs, float* potential, float prev_pote
 
     potential[next_i] = p_calc_->calculatePotential(potential, costs[next_i] + neutral_cost_, next_i, prev_potential);
     int x = next_i % nx_, y = next_i / nx_;
-    float distance = abs(end_x - x) + abs(end_y - y);
+
+    //float distance = abs(end_x - x) + abs(end_y - y);
```
DLu, line 93:
```suggestion
```
Please just remove code instead of commenting it out.

global_planner/src/astar.cpp
```diff
@@ -89,7 +89,9 @@ void AStarExpansion::add(unsigned char* costs, float* potential, float prev_pote
...
+    float distance = sqrt((end_x - x)*(end_x - x) + (end_y - y)*(end_y - y));	//Euclidean Distance
```
DLu, line 94: Can you add a parameter that changes which distance metric is used?

global_planner/src/gradient_path.cpp
```diff
@@ -81,7 +81,8 @@ bool GradientPath::getPath(float* potential, double start_x, double start_y, dou
         // check if near goal
         double nx = stc % xs_ + dx, ny = stc / xs_ + dy;
 
-        if (fabs(nx - start_x) < .5 && fabs(ny - start_y) < .5) {
+        //if (fabs(nx - start_x) < .5 && fabs(ny - start_y) < .5) {
+        if (fabs(nx - start_x) < 1 && fabs(ny - start_y) < 1) {		//To be able to work with A* algorithm and GradientPath method
```
DLu, line 85: Can you give an example of where the combination A*/Gradient was NOT finding a path that came within the distance before?

DLu (changes requested):

global_planner/src/planner_core.cpp
MikeDegany: Alright.

global_planner/src/astar.cpp
MikeDegany: Sorry I'm a little new to Git and things. I would add a parameter to choose which metric to use.

global_planner/src/gradient_path.cpp
MikeDegany: When I run "teb_local_planner_tutorials" demo launch file which uses a hierarchical approach to plan a trajectory using A*/Gradient for global planning, I get these errors:
[ERROR]: NO PATH!
[ERROR]: Failed to get a plan from potential when a legal potential was found. This shouldn't happen.

remco-r: I always had this error as well, this fixed it for me, thanks!

mikeferguson:

PR #964, opened 2020-01-26
Add frame ID to empty NavFn paths

### LINKED ISSUE #963: NavFn publishes empty plans for visualization
NavFn [will publish empty plans for visualization](https://github.com/ros-planning/navigation/blob/0f6b9f152cb9317da526f883fc154810fc4e9af0/navfn/src/navfn_ros.cpp#L364), but because these paths don't have a frame id, they'll just make tools like RViz yell because they aren't going visualize something that doesn't have a frame (`Invalid argument passed to canTransform argument source_frame in tf2 frame_ids cannot be empty`).

It was changed to be this way in 46db461a98a1832a2d but there was no rationale, so I don't understand why this behavior is important. It seems to me that it'll just cause confusion.

nickswalker:
RViz will complain about not being able to transform the path if it doesn't have a frame ID, so we'll just drop these

Closes #963

### DIFF
--- a/navfn/src/navfn_ros.cpp
+++ b/navfn/src/navfn_ros.cpp
@@ -349,13 +349,14 @@ namespace navfn {
     //create a message for the plan 
     nav_msgs::Path gui_path;
     gui_path.poses.resize(path.size());
-    gui_path.header.stamp = path[0].header.stamp;
     
     if(path.empty()) {
       //still set a valid frame so visualization won't hit transform issues
     	gui_path.header.frame_id = global_frame_;
+      gui_path.header.stamp = ros::Time::now();
     } else { 
       gui_path.header.frame_id = path[0].header.frame_id;
+      gui_path.header.stamp = path[0].header.stamp;
     }
 
     // Extract the plan in world co-ordinates, we assume the path is all in the same frame
--- a/navfn/src/navfn_ros.cpp
+++ b/navfn/src/navfn_ros.cpp
@@ -346,15 +346,17 @@ namespace navfn {
       return;
     }
 
+    //given an empty path we won't do anything
+    if(path.empty()) {
+    	return;
+    }
+
     //create a message for the plan 
     nav_msgs::Path gui_path;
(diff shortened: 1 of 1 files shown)


navfn/src/navfn_ros.cpp
```diff
@@ -346,15 +346,17 @@ namespace navfn {
       return;
     }
 
+    //given an empty path we won't do anything
+    if(path.empty()) {
```
SteveMacenski: This should not return, we still want to publish the empty path.

The ticket filed talks about not having the timestamp / frame. But you still have to have it published, which happens below, to get any result.

SteveMacenski (changes requested):

navfn/src/navfn_ros.cpp
nickswalker: Ah, okay. I thought the purpose of this topic was purely for visualization, in which case it doesn't make sense to visualize an empty path (therefore we would revert the code to the behavior in 46db461a98a1832a2d1b98b9d6f76f83dee12d63). But if it's the expectation that this topic returns the result of planning, failure included, that makes sense.

navfn/src/navfn_ros.cpp
```diff
@@ -346,17 +346,17 @@ namespace navfn {
...
-    //given an empty path we won't do anything
-    if(path.empty()) {
-    	return;
-    }
-
     //create a message for the plan 
     nav_msgs::Path gui_path;
     gui_path.poses.resize(path.size());
-    gui_path.header.frame_id = path[0].header.frame_id;
     gui_path.header.stamp = path[0].header.stamp;
```
SteveMacenski: `     gui_path.header.stamp = path[0].header.stamp; ` if path empty, this is invalid access

SteveMacenski (changes requested):

SteveMacenski (approved):
Seems reasonable to me

DLu:
Thanks!

PR #969, opened 2020-02-26
don't include a main() function in base_local_planner library

ipa-fez:
If you link against base_local_planner and forget to add a `main()` function (by for example defining it in some namespace by accident), your executable would use this main function, greeting you with very unexpected console output after executing your node.

I've therefore moved this into its own file and linked that against base local planner rather than including it in the library.

Since this seems to be some test code and the target isn't even installed, this could arguably also be removed completely.


DLu:
Good catch

PR #971, opened 2020-03-03
Implement selective resampling (forward port of #921)

mikeferguson:


SteveMacenski:
@mikeferguson is this a change we should make in ROS2 as well?

PR #984, opened 2020-04-03
add explicit call to ros::Timer::stop in the destructor

dorezyuk:
Hi all, 
there is a bug if the costmap_2d_ros class gets destructed while being in the timer-callback. 

Since the class members are destructed in the reverse order of their definition in the header file, we might end up accessing already destructed members inside the movementCB. 

One solution (which technically is better) is to define the timer_ member last - however, it's difficult to maintain (if someone decides to add more members). Therefore I decided to add an explicit `Timer::stop` call to the destructor

### DIFF
--- a/costmap_2d/src/costmap_2d_ros.cpp
+++ b/costmap_2d/src/costmap_2d_ros.cpp
@@ -184,6 +184,8 @@ void Costmap2DROS::setUnpaddedRobotFootprintPolygon(const geometry_msgs::Polygon
 
 Costmap2DROS::~Costmap2DROS()
 {
+  timer_.stop();
+
   map_update_thread_shutdown_ = true;
   if (map_update_thread_ != NULL)
   {


DLu:
* [Timer defined here](https://github.com/ros-planning/navigation/blob/3372226261378b31602723f866aba7fd71d9a7ac/costmap_2d/include/costmap_2d/costmap_2d_ros.h#L256) (line 256)
 * [`old_pose_` defined here](https://github.com/ros-planning/navigation/blob/3372226261378b31602723f866aba7fd71d9a7ac/costmap_2d/include/costmap_2d/costmap_2d_ros.h#L260) (line 260)
 * [`movementCB` using `old_pose_`](https://github.com/ros-planning/navigation/blob/3372226261378b31602723f866aba7fd71d9a7ac/costmap_2d/src/costmap_2d_ros.cpp#L421)

So it looks like you are correct. However, I don't think calling stop actually prevent in-progress timer callbacks from still executing, right?

dorezyuk:
> So it looks like you are correct. However, I don't think calling stop actually prevent in-progress timer callbacks from still executing, right?

Fortunately it does:
The Timer::Impl::stop will call the TimerManager::remove: [timer.cpp, line 80](https://github.com/ros/ros_comm/blob/6b27222b624b42ceb4721e60b88f84d6d24cb9f7/clients/roscpp/src/libros/timer.cpp#L80)

Which will try to remove the callback from the callback_queue: [timer_manager.h, line 399](https://github.com/ros/ros_comm/blob/6b27222b624b42ceb4721e60b88f84d6d24cb9f7/clients/roscpp/include/ros/timer_manager.h#L399)

Which will acquire the callback's own lock: [callback_queue.cpp, line 169](https://github.com/ros/ros_comm/blob/6b27222b624b42ceb4721e60b88f84d6d24cb9f7/clients/roscpp/src/libros/callback_queue.cpp#L169)

And this lock is held while the callback is active: [callback_queue.cpp, line 369](https://github.com/ros/ros_comm/blob/6b27222b624b42ceb4721e60b88f84d6d24cb9f7/clients/roscpp/src/libros/callback_queue.cpp#L369)

mikeferguson:
Cherry-picked to noetic - fd22136c5464a0641b2072b3de7a3fd439cdc639

PR #992, opened 2020-05-08
Increase scope of costmap mutex in publishCostmap to cover costmap calls

griswaldbrooks:


mikeferguson:
@griswaldbrooks have you tested this with the first path? Where the lock is not normally taken (easiest route would be by setting always_send_full_costmap). I'm pretty sure this is a recursive mutex, but just want to make sure we aren't introducing a deadlock in a different part of the code.

griswaldbrooks:
Sorry to have taken so long in getting back. Yes, it is a recursive mutex https://github.com/ros-planning/navigation/blob/melodic-devel/costmap_2d/include/costmap_2d/costmap_2d.h#L294
I've tested it with the `always_send_full_costmap` set to true.
We've deployed this (and a related fix to our fleet) and had gotten permission to back port this to open source.

mikeferguson:
Also cherry-picked to noetic

griswaldbrooks:
Thanks!

PR #998, opened 2020-05-29
[noetic] Adding CMake way to find yaml-cpp

seanyen:
On some package managers (e.g., `vcpkg` and `conda-fogre`), `pkg-config` file is not generated for `yaml-cpp` Windows build. Instead, the `CMake` config is generated for the package discovery.

This change is to make the `map_server` more portable to be built on those environments.


seanyen:
@mikeferguson This is ready for review and merge. This one is motivated by attempting to enable `ROS noetic` with [`conda-forge`](https://conda-forge.org/) binary packages on Winodws.

mikeferguson (approved):

mikeferguson:
LGTM and continues to pass build on ubuntu/debian

Issue #147, opened 2013-11-07
DWA Fails at the end of the path


DLu:
Discussed here: http://answers.ros.org/question/94536/testing-dwa-local-planner-approaches-goal-stops-keeps-swinging-left-right/#96786

Temporary fix is here:
https://github.com/ros-planning/navigation/blob/hydro-devel/dwa_local_planner/src/dwa_planner.cpp#L176
Set cheat_factor to 2 to avoid the problem, but doesn't address the fundamental problem. 

The fundamental problem stems from the weight on the alignment cost being set incorrectly (see https://github.com/ros-planning/navigation/commit/6a72c61 for one fix)

DLu:
Hey Hendrik,
Just to confirm, did you change the value of cheat_factor?
-DL!!

HWiese1980:
Ah, no, I didn't. That might be the cause. What would be a reasonable value and where do I have to set it?

DLu:
$MOVE_BASE_NAMESPACE/DWAPlannerROS/cheat_factor 2

HWiese1980:
Alright. Thanks, I'll give it a try and report back.

HWiese1980:
Okay, I have set the cheat_factor to 2. It still depends on the value of sim_time (1.5 works, 3.0 doesn't) and the xy_goal_tolerance (higher tolerance works, lower doesn't). Higher values for the cheat_factor don't appear to have bigger impact on this behavior.

DLu:
Ok. As implied by the name, cheat_factor isn't a real fix. Is your code
using this fix? https://github.com/ros-planning/navigation/pull/154

HWiese1980:
Yeah, that's what I thought. I'm on hydro-devel HEAD revision. If the fix hasn't yet been merged into it, I don't use it. I'll check it out and report back. 

//edit: alright, it obviously **is** merged, part of the head revision and thus part of my stack as well. So, yes, my code uses #154

DLu:
Hey all,
I spent the weekend looking at this. Can you try this branch please?
https://github.com/DLu/navigation/tree/hydro_end_path_fix
-DL!!

HWiese1980:
Oh dear, I hope it's not completely my fault in the end... having busted your whole weekend for nothing would be a pretty unpleasant thought. I'll try it and, again, report back.

[...]

PR #999, opened 2020-06-01
Rotational fixes

wjwagner:
This fixes a number of problems related to rotation within DWA and BLP.

- Fix issue #147 for both DWA and BLP by creating check for if the goal has changed or not
- Fixes rotatetogoal function for both DWA and BLP by correctly imposing velocity limits. Improved in-code documentation.
- Changed angular velocity limit parameter names and the variables that reference them in the code to unify naming conventions and reduce confusion about the meaning of parameters. See commit 7144ff6.
- Fixed BLP dynamic reconfigure


wjwagner:
I know this PR covers a lot, but all of the changes were rather interdependent. Basically I discovered some errors in the kinematics for the rotatetogoal functions, but when trying to implement these changes on both BLP and DWA I discovered that the naming convention between the parameters was not consistent. And not only were they inconsistent, they were misleading, hence the changes to the parameter names and the variables they set in the code.

I could try and break this up a bit more if that would be useful, but I'm not sure what the best way to do that would be.

mikeferguson:

PR #1000, opened 2020-06-01
Remove scaling of BLP occdist_scale by costmap resolution

wjwagner:
occdist_scale should not be scaled by the costmap resolution as it doesn't multiply a value that includes a distance.

### DIFF
--- a/base_local_planner/src/trajectory_planner_ros.cpp
+++ b/base_local_planner/src/trajectory_planner_ros.cpp
@@ -198,7 +198,6 @@ namespace base_local_planner {
           double resolution = costmap_->getResolution();
           goal_distance_bias *= resolution;
           path_distance_bias *= resolution;
-          occdist_scale *= resolution;
         } else {
           ROS_WARN("Trajectory Rollout planner initialized with param meter_scoring set to false. Set it to true to make your settings robust against changes of costmap resolution.");
         }


mikeferguson:
This seems totally reasonable - however, it will greatly change the tuning of any existing robot. Therefore it should probably be added behind a feature flag (which, honestly we have too many) or should be targeted to noetic-devel which is not yet considered stable.

wjwagner:
It appears that it doesn't actually change anything since occdist_scale is correctly not being scaled in the reconfigure callback in trajectory_planner.cpp. It appears that even though it is scaled in trajectory_planner_ros.cpp and passed to trajectory_planner the dynamic reconfigure callback resets it to the value on the parameter server. I put some ROS_INFOs in to test this, see my test_occdist_scale branch and [this commit](https://github.com/wjwagner/navigation/commit/5eca9c21bc6405706b28dc4e73bca6b0d354c678).
I set occdist_scale to a value of 0.1 on the parameter server and then ran move_base.

[ INFO] [1591126774.101520819, 11.807000000]: In trajectory_planner_ros.cpp occdist_scale is: 0.020000
[ INFO] [1591126774.125994857, 11.831000000]: In trajectory_planner.cpp reconfigure() occdist_scale is: 0.100000
[ INFO] [1591126774.287428830, 11.993000000]: Recovery behavior will clear layer 'obstacles'
[ INFO] [1591126774.291273322, 11.997000000]: Recovery behavior will clear layer 'obstacles'
[ INFO] [1591126774.312577880, 12.017000000]: odom received!
[ INFO] [1591126795.958841205, 33.584000000]: In trajectory_planner_ros.cpp generateTrajectory() occdist_scale is: 0.100000

mikeferguson:
@wjwagner - very interesting observation. I'll take a closer look at this tonight.

mikeferguson:
cherry-picked to noetic

PR #1005, opened 2020-07-03
Alternative to #641: Checking which layers to clear

DLu:
* Tries both `"obstacles"` and `"obstacle_layer"` as default parameters. 
 * Prints the actual layers that it will clear, as opposed to the value of the parameter. 
 * Prints an error if no layers will be cleared. 

CC: @LeroyR


LeroyR:
👍 

You need to update/verify test results.

SteveMacenski:
@DLu on this subject I filed a ticket in navigation2: https://github.com/ros-planning/navigation2/issues/1850 - do you agree with my idea of adding an API method for clearable for the layer itself so that we can remove the parameterization (which cough I've frequently forgotten existed cough)

DLu:
The problem with adding the `clearable()` method to the API is that sometimes you may want to specify which recoveries clear which layers. I've seen configurations where you clear different layers in different recoveries.

mikeferguson:

PR #1006, opened 2020-07-03
Initial map_server map_mode tests

DLu:
In reviewing #513, I made some tests to check for proper functioning of `MapMode`. 

While doing so, I found a bug in the `SCALE` mode. The intermediate values should have the range [`(0, 100)`](https://github.com/ros-planning/navigation/blob/b496f6891bfa8c6f471773bacb9b327e41bd36bb/map_server/include/map_server/image_loader.h#L47) (non-inclusive), but the value of 1 was being interpreted as 0, causing the tests to fail.


aaronhoy (approved):
LGTM

PR #502, opened 2016-07-20
Publish status info about recovery behaviors when they occur

PeterMitrano:
new message tells you where the behavior occured, the name, index,
and total number of behaviors.

Should I also do something for jade & kinetic? Seems like it would be appropriate.

Addresses #493

PR #1007, opened 2020-07-03
Publish recovery behavior kinetic

DLu:
Depends on https://github.com/ros-planning/navigation_msgs/pull/19

Replacement for #502


DLu:
@ros-pull-request-builder Rebuild this please

mikeferguson:
So we never actually released the updated messages - I'm doing that now (but it will take a week or two for the messages to make it to public and this PR to pass)

mikeferguson:
This will require the fixes added to melodic + noetic in https://github.com/ros-planning/navigation/pull/1071

mikeferguson (approved):
Yay tests are finally passing!

kyrofa:
Hey folks, this broke ABI. I assume you knew, but just in case you didn't, wanted to point it out.

mikeferguson:
Yeah, AFAIK, there is nothing downstream that loads move_base as a library (we are much more careful around anything library or plugin-related where ABI is super important). Do you have a use case where the ABI breakage causes an actual issue?

kyrofa:
I don't know of anything specific, but our tests flagged it. Have you considered using a more change-friendly idiom (e.g. d-pointer/pimpl)? Then you'd be able to make such changes without breaking ABI. IMHO trying to get your head around how folks use your library so you know what's safe to break and what isn't is fraught with peril :stuck_out_tongue: .

mikeferguson:
Honestly, this entire repo is basically EOL - ROS2 went and built an entirely different navigation stack, so this isn't getting ported over and there are no new ROS1 releases coming, so there really isn't a future release to introduce major API/ABI changes into

kyrofa:
That's fair. All the more reason to be careful with updates to stable releases, then!

PR #1011, opened 2020-07-13
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

Python 2 died on 1/1/2020.

### DIFF
--- a/amcl/test/basic_localization.py
+++ b/amcl/test/basic_localization.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+
 import sys
 import time
 from math import fmod, pi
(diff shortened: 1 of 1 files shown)


mikeferguson:
Note, this is actually fixed in the noetic-devel branch (which also has some other patches to be fully python3 compatible)

cclauss:
Cool.  Let’s move those changes to master because legacy Python died 195 days ago.

MatthijsBurgh:
Don't forget to include `from __future__ import print_function`. With this import you are sure that the correct arguments are provided.

cclauss:
The future import was only needed for legacy print statements that contained commas.  Given that none of these print statements contained commas, the future statement was not required.

Now that legacy Python is dead, the future is already here and this future statement is already deprecated.

MatthijsBurgh:
Most ROS users still use python2. So assuming python2 is dead is not valid in ROS country.

Plus adding the import prevents the addition of future changes which don't match the python3 api.

PR #1012, opened 2020-07-16
(AMCL) add resample limit cache [Melodic]

MatthijsBurgh:
This reduces the number of calculations significant as this function is called each time after adding a new sample during resampling.

### DIFF
--- a/amcl/src/amcl/pf/pf.c
+++ b/amcl/src/amcl/pf/pf.c
@@ -70,6 +70,9 @@ pf_t *pf_alloc(int min_samples, int max_samples,
   pf->pop_err = 0.01;
   pf->pop_z = 3;
   pf->dist_threshold = 0.5; 
+
+  // Number of leaf nodes is never higher than the max number of samples
+  pf->limit_cache = calloc(max_samples, sizeof(int));
   
   pf->current_set = 0;
   for (j = 0; j < 2; j++)
@@ -115,7 +118,9 @@ pf_t *pf_alloc(int min_samples, int max_samples,
 void pf_free(pf_t *pf)
 {
   int i;
-  
+
+  free(pf->limit_cache);
+
   for (i = 0; i < 2; i++)
   {
     free(pf->sets[i].clusters);
@@ -512,8 +517,16 @@ int pf_resample_limit(pf_t *pf, int k)
   double a, b, c, x;
   int n;
 
+  assert(k > 0);
+
+  if (pf->limit_cache[k-1] > 0)
+    return pf->limit_cache[k-1];
+
   if (k <= 1)
+  {
+    pf->limit_cache[k-1] = pf->max_samples;
     return pf->max_samples;
+  }
 
   a = 1;
   b = 2 / (9 * ((double) k - 1));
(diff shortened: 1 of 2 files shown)


MatthijsBurgh:
ping ;)

DLu (commented):
I'm not familiar with the internals of amcl as well as I should be. Can you give a bit of context? How often is `pf_resample_limit` called with the same values? Is there ever a scenario where the cache could be invalid?

MatthijsBurgh:
The cache should never be invalid. The size is depending on `max_particles`, the values inside the cache depend on `min_particles`, `max_particles`, `pop_z` and `pop_err`. These can be changed during a reconfigure callback, but in that case, the entire filter is deleted and created again, thus also the cache, so both size and values are "reset".

I can't give you a number for calling with the same value during one resampling, as that really depends on the cell size and how your particles are spread. But between multiple resampling steps there will be significant overlap of the input values.
By default, resampling happens every 2 sensor updates. So probably at 10 Hz, During each resampling, `pf_resample_limit` is called at least `min_particles` times (100 by default), so at least 1000 times per second.

MatthijsBurgh:
@DLu ping ;)

MatthijsBurgh:
@DLu ping ;)

MatthijsBurgh:
ping @aaronhoy @mikeferguson ;)

amcl/src/amcl/pf/pf.c
```diff
@@ -512,8 +517,16 @@ int pf_resample_limit(pf_t *pf, int k)
   double a, b, c, x;
   int n;
 
+  assert(k > 0);
```
mikeferguson: so, is this important, are there instances when someone would actually call with k <= 0? It will actually be removed from the Debian release (and, actually, for almost everyone since it is recommended to compile in RELEASE to avoid issues with runtime speed).

I'd prefer we avoid hard-to-debug memory issues here at runtime, and instead return pf->max_samples if k < 1, and then we know that all the lines below are OK. Although I would also suggest updating L525 below

amcl/src/amcl/pf/pf.c
```diff
@@ -512,8 +517,16 @@ int pf_resample_limit(pf_t *pf, int k)
...
+
+  if (pf->limit_cache[k-1] > 0)
+    return pf->limit_cache[k-1];
+
   if (k <= 1)
```
mikeferguson: if we return pf->max_samples for k < 1 above, make this (k == 1) for clarity

amcl/src/amcl/pf/pf.c
```diff
@@ -512,8 +517,16 @@ int pf_resample_limit(pf_t *pf, int k)
   double a, b, c, x;
   int n;
 
+  assert(k > 0);
+
+  if (pf->limit_cache[k-1] > 0)
```
mikeferguson, line 526: before we access limit_cache[k-1]  we really should validate that k <= max_samples to avoid memory corruption

mikeferguson (changes requested):
Ok, @MatthijsBurgh your persistence has paid off - I finally dragged out my copy of Probabilistic Robotics to review KLD resampling. This looks good - except for the three comments noted - rather than an assert (which will disappear when compiled in RELEASE), let's make sure that k < 1 returns reasonable values. Also, let's prevent memory corruption.

MatthijsBurgh:
@mikeferguson I have implemented your requests. I also added the changes to the kinetic PR, #1013, and the noetic PR, #1014.

PR #1014, opened 2020-07-16
(AMCL) add resample limit cache [Noetic]

MatthijsBurgh:
This reduces the number of calculations significant as this function is called each time after adding a new sample during resampling.

Noetic version of #1012


MatthijsBurgh:
ping ;)

PR #1017, opened 2020-07-21
Fix #796

### LINKED ISSUE #796: Error in fake_localization package: lookupTransform. Not enough time?
I have a problem with the package fake_localization.

When I set a new initial_pose in rviz, I get the error ""Failed to lookup transform!"":
imagen

This is my tf tree, I think that it's ok.
imagen

- I looking for a solution for the problem and I find this question with the same problem that me.
https://answers.ros.org/question/287587/lookuptransform-is-throwing-error/
The problem is that we need to wait for the transforms to become available.

- I need to wait for the tf at least 0.052 s

RESULTS: for map to base_link
Chain is: map -> odom -> base_link
Net delay     avg = 0.0262871: max = 0.0519673

- tf2 provides a tool that will wait until a transform becomes available.
imagen

Sorry for my English.
Regards and thank you.

rkoyama1623-jsk:
Use ros::Time(0) instead of timestamp in message so as not to fail to lookupTransform.

This change is suggested by @guohefu, @JavierLRH, and @mikeferguson in #796, and seems good to me too.
Thank you!

### DIFF
--- a/fake_localization/fake_localization.cpp
+++ b/fake_localization/fake_localization.cpp
@@ -247,7 +247,7 @@ class FakeOdomNode
       geometry_msgs::TransformStamped baseInMap;
       try{
 	// just get the latest
-        baseInMap = m_tfBuffer->lookupTransform(base_frame_id_, global_frame_id_, msg->header.stamp);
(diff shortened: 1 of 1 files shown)


DLu:
Thanks! Cherry picked to kinetic and noetic.

rkoyama1623-jsk:
Thanks for your response!

PR #1025, opened 2020-09-10
correctly delete previously allocated array

mechatheo:
found with clang leak sanitizer. Under heavy load calling the replanning service, led to strange freeing related errors and segfaults.


mikeferguson (approved):
Thanks!

PR #1027, opened 2020-09-11
Throttle spammy line

nickvaras:
This line really spams the logs, please consider throttling as suggested.


DLu:
Given that the information printed each time is theoretically different, it seems like it would make more sense to hide the printing behind a flag, rather than throttle it.

mikeferguson:
I'm gonna concur that throttling seems like the wrong approach - either putting it behind a flag or changing it to a DEBUG_NAMED (which can be turned back on easily) would make more sense.

PR #1028, opened 2020-09-21
Fixes #1026

### LINKED ISSUE #1026: [melodic][global_planner] Raw-pointers of class GlobalPlanner never initialized to null in default constructor
The raw pointers `p_calc_, planner_, path_maker_, orientation_filter_` of class GlobalPlanner are never initialized to nullptr during default construction, causing the check in the destructor to fail (and consequently causing a seg fault) if the planner was never initialized.

DLu:


global_planner/include/global_planner/planner_core.h
```diff
@@ -193,7 +193,7 @@ class GlobalPlanner : public nav_core::BaseGlobalPlanner {
         int publish_scale_;
 
         void outlineMap(unsigned char* costarr, int nx, int ny, unsigned char value);
-        unsigned char* cost_array_;
```
DLu, line 196: This is unused in the package.

mikeferguson (approved):

aaronhoy (approved):

global_planner/include/global_planner/planner_core.h
kyrofa: This seems like an unnecessary ABI break.

PR #1029, opened 2020-09-22
Change_map service to map_server [Rebase/Noetic]

DLu:
This is a redo of #461 for noetic, integrating changes that have been made in the last 4 years since the original PR. :roll_eyes: 

The CI will fail until the changes from https://github.com/ros/common_msgs/pull/152 are released, see https://github.com/ros/common_msgs/issues/167

The return codes from the service are not ideal (in that it only returns success or failure, not the more nuanced failure codes described in the service) but I was trying to keep close to @sloretz's original PR structure.


map_server/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ project(map_server)
 find_package(catkin REQUIRED
         COMPONENTS
             roscpp
+            roslib
```
mikeferguson: this seems to only be in the Cmake/package.xml - where is it actually used/required?

map_server/src/main.cpp
```diff
@@ -67,156 +68,218 @@ class MapServer
...
-            exit(-1);
-          }
+  private:
+    ros::NodeHandle n;
+    ros::Publisher map_pub;
+    ros::Publisher metadata_pub;
+    ros::ServiceServer service;
+    ros::ServiceServer change_map_srv_;
+    bool deprecated;
+    std::string frame_id_;
```
mikeferguson, line 107: should really be more consistent in naming here - member variable should all have the trailing underscore_

mikeferguson (changes requested):

EricWiener:
Any progress updates on this?

DLu:

EricWiener:
@DLu thanks for seeing if the tests pass. Does it look good to merge?

DLu:
I think so. Let's get the +1 from @mikeferguson first since he had formally reviewed it earlier.

EricWiener:
Any progress on this?

map_server/src/main.cpp
```diff
@@ -67,156 +68,218 @@ class MapServer
...
+        response.result = response.RESULT_SUCCESS;
       }
-      catch (std::runtime_error e)
+      else
       {
-          ROS_ERROR("%s", e.what());
-          exit(-1);
+        response.result = response.RESULT_UNDEFINED_FAILURE;
       }
+      ROS_INFO("Changed map to %s", request.map_url.c_str());
```
mikeferguson: Seems odd to tell the user we changed the map if the loadMapFromYaml failed? Should this message be moved into the if() statement only? (it looks like we already get error messages from the loadMapFromYaml function).

mikeferguson (approved):

engineeve:
Any chance this will be merged within the next month?

HappySamuel:
Hi @DLu 

I saw that there's an addition feature (change_map) to the map_server. Is this add-on feature available if install via apt? 
`$ sudo apt install ros-noetic-map-server`
Because i check wiki of [ros noetic map server](http://wiki.ros.org/map_server), it remains outdated.

Best,
Samuel

DLu:
No, and its been over a year since the last release, so its overdue.

PR #1036, opened 2020-10-11
move enabled-logic to the LayeredCostmap

dorezyuk:
Hey guys, 
I found it very easy to miss to check if a costmap is enabled in the Layer::updateCosts and Layer::updateBounds methods and think this should be better checked in the LayeredCostmap.


mikeferguson (approved):

azhural:
Any chance to add this to melodic as well?

PR #1037, opened 2020-10-11
add onOriginChanged logic

dorezyuk:
Hi guys,

I've added an additional function `onOriginChanged`.

The main idea behind this is to allow to implement layers/cosmaps which are not always rolling but also not entirely static (something in between). So the user can call LayeredCostmap::updateOrigin, if he wishes to move the costmap, but can treat it as static in the remaining time.

The concept fits also well in the current code base - allowing
1. unify the logic from ObstacleLayer and VoxelLayer
2. avoid redundant repainting in the StaticLayer (if the robot and thus the origin has not moved)

### DIFF
--- a/costmap_2d/plugins/static_layer.cpp
+++ b/costmap_2d/plugins/static_layer.cpp
@@ -267,6 +267,12 @@ void StaticLayer::reset()
   }
 }
 
+void StaticLayer::onOriginChanged()
+{
+  // we need to repaint everything
+  has_updated_data_ = true;
+}
+
 void StaticLayer::updateBounds(double robot_x, double robot_y, double robot_yaw, double* min_x, double* min_y,
                                double* max_x, double* max_y)
 {
@@ -275,6 +281,11 @@ void StaticLayer::updateBounds(double robot_x, double robot_y, double robot_yaw,
     if (!map_received_ || !(has_updated_data_ || has_extra_bounds_))
       return;
   }
+
+  // if the origin has not changed, skip the re-painting, since we cannot add
+  // any new information
+  if(!has_updated_data_)
+    return;
 
   useExtraBounds(min_x, min_y, max_x, max_y);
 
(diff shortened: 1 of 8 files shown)


costmap_2d/plugins/static_layer.cpp
```diff
@@ -276,6 +282,11 @@ void StaticLayer::updateBounds(double robot_x, double robot_y, double robot_yaw,
       return;
   }
 
+  // if the origin has not changed, skip the re-painting, since we cannot add
+  // any new information
+  if(!has_updated_data_)
+    return;
```
mikeferguson: I'm not convinced this is right:

- If map_received_ is true, then has_updated_data is set to true, so that case is OK
- but what about has_extra_bounds_ - that is handled in the base class - so it cannot be setting has_updated_data_. I think not doing the updates below when has_extra_bounds_ is true means that things like costmap clearing may not function properly?

I think this extra check and return should be removed.

mikeferguson (changes requested):

costmap_2d/plugins/static_layer.cpp
dorezyuk: good catch... writing 
```cpp
  if(!has_updated_data_ && !has_extra_bounds_)
    return;
```

would solve this. But I honestly don't understand the logic behind `useExtraBounds` for this layer: Its never set and never used... but thats maybe too much for this PR

mikeferguson: I believe "useExtraBounds" is only called in clearing the cost maps? It forces data to be copied over? I think removing this extra if/return statement will suffice.

dorezyuk: Ah yes, you are right, its triggered from clear-costmap-recovery. 

We could remove the if/return but the idea here is to stop re-rendering the entire local costmap if, we aren't actually moving (and thus if we don't have additional info from the static-layer). I've tested the `!has_updated_data_ && !has_extra_bounds_` case and it seems to work. But lastly its your call - shall I drop it?

dorezyuk: I take it as a yes

dorezyuk:
@mikeferguson friendly ping

mikeferguson:

PR #1039, opened 2020-10-14
[move_slow_and_clear] Add rosparam representing max vel rosparam name

708yamaguchi:
In order to use `move_slow_and_clear` with planners other than DWAPlannerROS (e.g. TrajectoryPlannnerROS), I added a rosparam named `max_trans_param_name` and `max_rot_param_name`.

When we use `move_slow_and_clear` in plugin, we cannot use remap and only the planners with `max_vel_trans`, `max_vel_theta` rosparam were available.

In this pull request, I added rosparams to change these max vel rosparam names.


mikeferguson (approved):

mikeferguson:
Also cherry-picked to noetic-devel

PR #1041, opened 2020-10-16
[global_planner] tolerance implementation

doisyg:
Implementation of the tolerance param which was never used.
See https://github.com/ros-planning/navigation/issues/262, https://github.com/ros-planning/navigation/pull/703 and https://github.com/ros-planning/navigation/pull/1040

First naive and not so optimized implementation which checks validity using `calculatePotentials` after filtering out lethal cells (so potentially CPU heavy for large tolerance).
Alternatively, a check for the cells value could be used without using `calculatePotentials`, or also filtering inscribed or even circumscribed cells. I am up for the debate.


derektan95:
Hi, could we please have these changes in ROS Melodic as well? It will be really helpful, thanks.

mikeferguson:

PR #1049, opened 2020-11-12
Fix Costmap2D future robot pose with static_transform_publisher

### LINKED ISSUE #1048: Costmap2D future robot pose with static_transform_publisher
When TF frames are published into the future, e.g. when using `static_transform_publisher`,
Costmap2D gets future stamped robot pose.
This is because it requests the latest transformation
https://github.com/ros-planning/navigation/blob/3c16b37504dc178c3075345ed88d681a6f83cc2e/costmap_2d/src/costmap_2d_ros.cpp#L579

This, in turn, can lead local planner to get `Extrapolation into the future` error when trying to convert the latest robot pose to the frame of the plan (`odom->map`).
https://github.com/ros-planning/navigation/blob/3c16b37504dc178c3075345ed88d681a6f83cc2e/base_local_planner/src/goal_functions.cpp#L118

This is because the latest robot pose has been got using the latest `base_link->odom` transform, while `odom->map` may be lagging behind (e.g. if it's also published by `static_transform_publisher`). And previous transformation of global plan from `map` to `odom` is also lagging behind
https://github.com/ros-planning/navigation/blob/3c16b37504dc178c3075345ed88d681a6f83cc2e/base_local_planner/src/goal_functions.cpp#L113

Proposed solution is to try to get robot pose for the current time by Costmap2D, and switch to the latest robot pose only if that is not possible.

pavloblindnology:
Fixes #1048

### DIFF
--- a/costmap_2d/src/costmap_2d_ros.cpp
+++ b/costmap_2d/src/costmap_2d_ros.cpp
@@ -583,12 +583,14 @@ bool Costmap2DROS::getRobotPose(geometry_msgs::PoseStamped& global_pose) const
   try
   {
     // use current time if possible (makes sure it's not in the future)
-    if (tf_.canTransform(global_frame_, robot_base_frame_, current_time)) {
+    if (tf_.canTransform(global_frame_, robot_base_frame_, current_time))
+    {
       geometry_msgs::TransformStamped transform = tf_.lookupTransform(global_frame_, robot_base_frame_, current_time);
       tf2::doTransform(robot_pose, global_pose, transform);
     }
     // use the latest otherwise
-    else {
+    else
+    {
       tf_.transform(robot_pose, global_pose, global_frame_);
     }
   }
--- a/costmap_2d/src/costmap_2d_ros.cpp
+++ b/costmap_2d/src/costmap_2d_ros.cpp
@@ -582,7 +582,15 @@ bool Costmap2DROS::getRobotPose(geometry_msgs::PoseStamped& global_pose) const
   // get the global pose of the robot
   try
   {
-    tf_.transform(robot_pose, global_pose, global_frame_);
+    // use current time if possible (makes sure it's not in the future)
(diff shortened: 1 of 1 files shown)


costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -582,7 +582,15 @@ bool Costmap2DROS::getRobotPose(geometry_msgs::PoseStamped& global_pose) const
   // get the global pose of the robot
   try
   {
-    tf_.transform(robot_pose, global_pose, global_frame_);
+    // use current time if possible (makes sure it's not in the future)
+    if (tf_.canTransform(global_frame_, robot_base_frame_, current_time)) {
```
mikeferguson: I know styling is inconsistent through navigation repo - but let's keep this file as consistent as possible, please move the opening brace to the it's own line.

costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -582,7 +582,15 @@ bool Costmap2DROS::getRobotPose(geometry_msgs::PoseStamped& global_pose) const
...
+      geometry_msgs::TransformStamped transform = tf_.lookupTransform(global_frame_, robot_base_frame_, current_time);
+      tf2::doTransform(robot_pose, global_pose, transform);
+    }
+    // use the latest otherwise
+    else {
```
mikeferguson: Same - new line for the brace

mikeferguson (changes requested):
This looks good - logic checks out. Can you update the style issues and retarget this to Noetic-devel (I will cherry-pick to back port to melodic)

costmap_2d/src/costmap_2d_ros.cpp
pavloblindnology: Done

pavloblindnology:
@mikeferguson Can you please cherry pick this to melodic?

PR #1052, opened 2020-11-22
Publish recovery behavior Noetic

DLu:
#1007 / #1051 for noetic. Will pass CI someday.


mikeferguson (approved):
Interesting - CI passes (it used 1.14.1 for move_base_msgs) - but as far as I can tell, public is still on 1.14.0 - we should probably wait to merge this until the noetic sync happens in a few days.

PR #1054, opened 2020-11-24
check if the layer is derived from costmap_2d::CostmapLayer

dorezyuk:
Hey guys,

in the current implementation, if the user defines layers not derived from costmap_2d::CostmapLayer to be cleared, we will just segfault. I've added a check if the cast can be performed so we get an error instead


mikeferguson (approved):

mikeferguson:
Looks good - can be merged once the CI completes successfully.

PR #1055, opened 2020-11-24
Fixed to use globally defined values

yukkysaito:
https://github.com/ros-planning/navigation/blob/noetic-devel/costmap_2d/include/costmap_2d/cost_values.h
Despite being defined in this file, used local values.


yukkysaito:
For reference, In this file, the global values are used.
https://github.com/ros-planning/navigation/blob/noetic-devel/costmap_2d/plugins/obstacle_layer.cpp#L64-L67

mikeferguson (approved):

PR #1056, opened 2020-11-27
Fix carrot planner

dorezyuk:
Hey guys, 

found out during testing, that the carrot-planner may crash the same way as global-planner did if not called initialize. This PR bundles some miner fixes

## Fixes
- memory leak of world_model
- unitilialized raw pointer
- add missing dependency of angles

### DIFF
--- a/carrot_planner/src/carrot_planner.cpp
+++ b/carrot_planner/src/carrot_planner.cpp
@@ -46,10 +46,10 @@ PLUGINLIB_EXPORT_CLASS(carrot_planner::CarrotPlanner, nav_core::BaseGlobalPlanne
 namespace carrot_planner {
 
   CarrotPlanner::CarrotPlanner()
-  : costmap_ros_(NULL), initialized_(false){}
+  : costmap_ros_(NULL), costmap_(NULL), world_model_(NULL), initialized_(false){}
 
   CarrotPlanner::CarrotPlanner(std::string name, costmap_2d::Costmap2DROS* costmap_ros)
-  : costmap_ros_(NULL), initialized_(false){
+  : costmap_ros_(NULL), costmap_(NULL), world_model_(NULL), initialized_(false){
     initialize(name, costmap_ros);
   }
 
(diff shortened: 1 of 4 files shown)


carrot_planner/src/carrot_planner.cpp
```diff
@@ -46,12 +47,17 @@ PLUGINLIB_EXPORT_CLASS(carrot_planner::CarrotPlanner, nav_core::BaseGlobalPlanne
...
 
   CarrotPlanner::CarrotPlanner(std::string name, costmap_2d::Costmap2DROS* costmap_ros)
-  : costmap_ros_(NULL), initialized_(false){
+  : costmap_ros_(NULL), costmap_(NULL), world_model_(NULL), initialized_(false){
     initialize(name, costmap_ros);
   }
+
+  CarrotPlanner::~CarrotPlanner() {
+    // deleting a nullptr is a noop
+    delete world_model_;
```
mikeferguson, line 59: I think we need "if world_model_" in front of the delete, otherwise we will crash if we destruct the object without calling initialize()

mikeferguson (changes requested):

carrot_planner/src/carrot_planner.cpp
dorezyuk: according to [this](https://stackoverflow.com/questions/4190703/is-it-safe-to-delete-a-null-pointer) its safe to delete a nullptr. here is an example from [online-gdb](https://onlinegdb.com/0qRjhBpzl). It's also done like this in this library in the destructor of [RotateRecovery](https://github.com/ros-planning/navigation/blob/da579109da77bda1729e21494c547f73cf57d53a/rotate_recovery/src/rotate_recovery.cpp#L90).

But if you had issues with this in the past (maybe on 'exotic' platforms) I can add the check. What do you think?

mikeferguson: Hmm, it does seem like this is OK - maybe it's just a very common anti pattern in this codebase? Seems like an unlikely to occur crash then.

mikeferguson (approved):

PR #1066, opened 2020-12-08
Fix Costmap2D future robot pose with static_transform_publisher

### LINKED ISSUE #1048: Costmap2D future robot pose with static_transform_publisher

pavloblindnology:
Fixes #1048
Supersedes #1049 for noetic

### DIFF
--- a/costmap_2d/src/costmap_2d_ros.cpp
+++ b/costmap_2d/src/costmap_2d_ros.cpp
@@ -582,7 +582,17 @@ bool Costmap2DROS::getRobotPose(geometry_msgs::PoseStamped& global_pose) const
   // get the global pose of the robot
   try
   {
-    tf_.transform(robot_pose, global_pose, global_frame_);
+    // use current time if possible (makes sure it's not in the future)
+    if (tf_.canTransform(global_frame_, robot_base_frame_, current_time))
+    {
+      geometry_msgs::TransformStamped transform = tf_.lookupTransform(global_frame_, robot_base_frame_, current_time);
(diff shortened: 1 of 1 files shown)


mikeferguson (approved):

mikeferguson:
Thanks for retargeting PR and updating style!

pavloblindnology:
@mikeferguson Can you please cherry pick this to melodic?

PR #1071, opened 2020-12-19
move_base crash when using default recovery behaviors

mattp256:
It looks like the recent update to publish recovery behavior leaves the recovery_behavior_names_ vector uninitialized if no recoveries are listed in configuration (i.e. using defaults). This caused move_base to crash during recovery on my setup, when trying to access the names vector to publish. Commit  b7379f7 is proposed to fix this.

Also, I thought the published info made more sense if it was 1-based instead of 0-based (e.g. "Executing behavior 1 of 2"). Commit d6ae66f is a proposal for that, but the crash fix can just be cherry-picked if this shouldn't change.

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -53,7 +53,7 @@ namespace move_base {
     as_(NULL),
     planner_costmap_ros_(NULL), controller_costmap_ros_(NULL),
     bgp_loader_("nav_core", "nav_core::BaseGlobalPlanner"),
-    blp_loader_("nav_core", "nav_core::BaseLocalPlanner"), 
+    blp_loader_("nav_core", "nav_core::BaseLocalPlanner"),
     recovery_loader_("nav_core", "nav_core::RecoveryBehavior"),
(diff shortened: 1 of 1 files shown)


mikeferguson (approved):

mikeferguson:
LGTM - @DLu you actually authored the changes that this references - can you double check it? I think we also need to back port to Melodic (Kinetic has not been merged yet)

DLu (changes requested):
Apologies for the bug. 

As for the indexing, the [message definition](https://github.com/ros-planning/navigation_msgs/blob/793d06ff51d20fcd94832161ccde268cb4ce5e38/move_base_msgs/msg/RecoveryStatus.msg#L2) denotes it as a 0-based index. I am ambivalent about keeping the +1 on the ROS_DEBUG but the current_recovery_number should stay as is.

mikeferguson:
CI was passing prior to build farm upgrades, the added commit should be fine (just removing a +1), going to merge so we can cherry pick and close out this bug

mikeferguson:
Cherry picked to melodic

PR #1072, opened 2020-12-20
Fix deadlock when starting map with map-update-frequency set to zero

dorezyuk:
Hi everyone,

I've noticed that if you set the update_frequency of the global costmap to zero, you run into a dead lock on planning, where 
- Costmap2DROS::mapUpdateLoop exits and 
- Costmap2DROS::start never exits its busy-wait while loop

The suggested fix resolves the issue

### DIFF
--- a/costmap_2d/src/costmap_2d_ros.cpp
+++ b/costmap_2d/src/costmap_2d_ros.cpp
@@ -359,7 +359,9 @@ void Costmap2DROS::reconfigureCB(costmap_2d::Costmap2DConfig &config, uint32_t l
 
   old_config_ = config;
 
-  map_update_thread_ = new boost::thread(boost::bind(&Costmap2DROS::mapUpdateLoop, this, map_update_frequency));
+  // only construct the thread if the frequency is positive
+  if(map_update_frequency > 0.0)
+    map_update_thread_ = new boost::thread(boost::bind(&Costmap2DROS::mapUpdateLoop, this, map_update_frequency));
 }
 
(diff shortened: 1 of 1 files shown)


costmap_2d/src/costmap_2d_ros.cpp
```diff
@@ -359,7 +359,9 @@ void Costmap2DROS::reconfigureCB(costmap_2d::Costmap2DConfig &config, uint32_t l
 
   old_config_ = config;
 
-  map_update_thread_ = new boost::thread(boost::bind(&Costmap2DROS::mapUpdateLoop, this, map_update_frequency));
+  // only construct the thread if the frequency is positive
+  if(map_update_frequency > 0.0)
+    map_update_thread_ = new boost::thread(boost::bind(&Costmap2DROS::mapUpdateLoop, this, map_update_frequency));
```
mikeferguson, line 365: what happens if we previously had frequency > 0 and reconfigure to 0?

dorezyuk: We destroy the thread when we enter the reconfigureCB in lines 326-328. But I've missed to set the map_update_thread_ to NULL - fixed that with the newest commit

mikeferguson (approved):

costmap_2d/src/costmap_2d_ros.cpp
mikeferguson: Thanks!

PR #1085, opened 2021-01-11
navfn: stop installing test headers

kyrofa:
The navtest executable is only built if FLTK is installed. However, the header it uses is installed regardless, and requires FLTK. Nothing else uses this header, and navfn doesn't depend upon FLTK, so stop installing the header. Also fix navtest to actually build when FLTK is installed.


mikeferguson (approved):

MatthijsBurgh:
Good idea to merge this into `melodic-devel` and `kinetic-devel`?

PR #1089, opened 2021-01-15
(AMCL)(Kinetic) use robot pose in tests

MatthijsBurgh:


mikeferguson (changes requested):
We haven't updated AMCL to use TF2 in kinetic, so the package.xml does not include the keys required for the python tf2 stuff you're using here. Please add the appropriate keys to the package.xml

(the reason this is still passing on the build farm is that all our dependencies get globbed together, and some packages like fake_localization have been ported to TF2 - so we're getting stuff pulled in. it's a low likelihood someone is going to run into this missing dependency, but it's better to be complete with our dependencies when possible).

mikeferguson (approved):

PR #1120, opened 2021-05-19
crop_map (melodic)

### LINKED ISSUE #1064: crop_map shifts origin by one extra pixel up
Every time the script is run it shifts origin by one pixel up.
E.g., if there is nothing to crop the origin is shifted up one pixel anyway.
The error is in this line,
https://github.com/ros-planning/navigation/blob/194d4958db5fa701c55ca0cd7b1934841f9ab1b2/map_server/scripts/crop_map#L67
which should look like
`dy = (map_image.size[1] - bounds[3] - 1) * resolution`

pavloblindnology:
Fixes #1064 for melodic


mikeferguson (approved):
Backport of #1067

PR #1122, opened 2021-05-21
map_server: Initialise NodeHandle outside the MapServer constructor

bergercookie:
Currently if an exception is raised inside the `MapServer` execution, it's caught by the `try-catch` statement in `main`. However at that point the `ros::NodeHandle` instance is already destroyed since it's constructed inside `MapServer::MapServer` and that's initialised inside the `try` scope.

Because of that the `ROS_ERROR` statement in the `catch` doesn't print anything and the user isn't informed of the exception.

In this PR I'm initialising ~the NodeHandle outside `MapServer` and I'm passing a reference to it during construction.~
a second NodeHandle outside `MapServer` in `main`

### DIFF
--- a/map_server/src/main.cpp
+++ b/map_server/src/main.cpp
@@ -238,6 +238,7 @@ class MapServer
 int main(int argc, char **argv)
 {
   ros::init(argc, argv, "map_server", ros::init_options::AnonymousName);
+  ros::NodeHandle nh("~");
   if(argc != 3 && argc != 2)
   {
     ROS_ERROR("%s", USAGE);


map_server/src/main.cpp
```diff
@@ -251,7 +252,7 @@ int main(int argc, char **argv)
 
   try
   {
-    MapServer ms(fname, res);
+    MapServer ms(private_nh, fname, res);
```
mikeferguson: Do we really need to pass in a special node handle? Isn't simply creating a node handle right after ros::init() above sufficient, such that we still have at least one NodeHandle in scope inside the catch()?

bergercookie: I thought it best of initializing and sharing a single NodeHandle instead, but sure, I could switch to that If you prefer.

bergercookie: Done

PR #1127, opened 2021-06-05
the pose of the robot requires a certain delay tolerance

DantesDawn:
When I was running this project,  to get the robot pose  often had a delay of 0.1 to 0.2 milliseconds, and an exception was thrown. I think the delay tolerance of 0.3ms can be increased here.

### DIFF
--- a/base_local_planner/src/goal_functions.cpp
+++ b/base_local_planner/src/goal_functions.cpp
@@ -115,7 +115,8 @@ namespace base_local_planner {
 
       //let's get the pose of the robot in the frame of the plan
       geometry_msgs::PoseStamped robot_pose;
-      tf.transform(global_pose, robot_pose, plan_pose.header.frame_id);
+      //I think that obtaining the pose of the robot requires a certain delay tolerance
+      tf.transform(global_pose, robot_pose, plan_pose.header.frame_id, ros::Duration(0.3));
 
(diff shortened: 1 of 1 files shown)


skpro19:
@DantesDawn could [this](https://github.com/ros-planning/navigation/issues/1131#issue-923022431) issue be related to your pull request?

DantesDawn:
@skpro19  I think my pull request should not have much to do with your isssue.  If your error or exception is due to the delay of tf2 coordinate conversion, it may be related to my pull request.

base_local_planner/src/goal_functions.cpp
```diff
@@ -115,7 +115,8 @@ namespace base_local_planner {
 
       //let's get the pose of the robot in the frame of the plan
       geometry_msgs::PoseStamped robot_pose;
-      tf.transform(global_pose, robot_pose, plan_pose.header.frame_id);
+      //I think that obtaining the pose of the robot requires a certain delay tolerance
+      tf.transform(global_pose, robot_pose, plan_pose.header.frame_id, ros::Duration(0.3));
```
dorezyuk, line 119: Would it make sense to make the tolerances used for transforming the global path (4 lines above set to 0.5s) and for transforming the global_pose consistent? Maybe you can just set both to 0.5s

PR #1128, opened 2021-06-07
Added navfn path length threshold

DatSpace:
This PR simply adds a new parameter for Navfn ROS, that allows you to limit the max allowed distance of the generated path.

### DIFF
--- a/navfn/include/navfn/navfn_ros.h
+++ b/navfn/include/navfn/navfn_ros.h
@@ -178,7 +178,7 @@ namespace navfn {
 
       void mapToWorld(double mx, double my, double& wx, double& wy);
       void clearRobotCell(const geometry_msgs::PoseStamped& global_pose, unsigned int mx, unsigned int my);
-      double planner_window_x_, planner_window_y_, default_tolerance_;
+      double planner_window_x_, planner_window_y_, default_tolerance_, plan_lenght_threshold_;
       boost::mutex mutex_;
       ros::ServiceServer make_plan_srv_;
       std::string global_frame_;
--- a/navfn/include/navfn/navfn_ros.h
+++ b/navfn/include/navfn/navfn_ros.h
@@ -178,7 +178,7 @@ namespace navfn {
 
       void mapToWorld(double mx, double my, double& wx, double& wy);
       void clearRobotCell(const geometry_msgs::PoseStamped& global_pose, unsigned int mx, unsigned int my);
-      double planner_window_x_, planner_window_y_, default_tolerance_, plan_lenght_threshold_;
+      double planner_window_x_, planner_window_y_, default_tolerance_, plan_length_threshold_;
       boost::mutex mutex_;
       ros::ServiceServer make_plan_srv_;
       std::string global_frame_;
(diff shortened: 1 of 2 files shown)


navfn/include/navfn/navfn_ros.h
```diff
@@ -178,7 +178,7 @@ namespace navfn {
 
       void mapToWorld(double mx, double my, double& wx, double& wy);
       void clearRobotCell(const geometry_msgs::PoseStamped& global_pose, unsigned int mx, unsigned int my);
-      double planner_window_x_, planner_window_y_, default_tolerance_;
+      double planner_window_x_, planner_window_y_, default_tolerance_, plan_lenght_threshold_;
```
DLu:
```suggestion
      double planner_window_x_, planner_window_y_, default_tolerance_, plan_length_threshold_;
```

navfn/src/navfn_ros.cpp
```diff
@@ -80,6 +80,7 @@ namespace navfn {
       private_nh.param("planner_window_x", planner_window_x_, 0.0);
       private_nh.param("planner_window_y", planner_window_y_, 0.0);
       private_nh.param("default_tolerance", default_tolerance_, 0.0);
+      private_nh.param("plan_lenght_threshold", plan_lenght_threshold_, 50.0);
```
DLu: The default behavior should be that there is no threshold, which is the current behavior.

navfn/src/navfn_ros.cpp
```diff
@@ -407,11 +408,21 @@ namespace navfn {
...
+    double prev_wx, prev_wy;
+    double path_len = 0;
+
     for(int i = len - 1; i >= 0; --i){
       //convert the plan to world coordinates
       double world_x, world_y;
       mapToWorld(x[i], y[i], world_x, world_y);
 
+      if (i < len - 1){
+        path_len += sqrt(pow(world_x - prev_wx, 2.0) + pow(world_y - prev_wy, 2.0));
```
DLu: Preferrable to use [`hypot`](http://www.cplusplus.com/reference/cmath/hypot/)

DLu (changes requested):
What is the overall goal here? I.e. What do you want to happen when the planned path is too long?

navfn/src/navfn_ros.cpp
DatSpace: Agreed. Done.

navfn/src/navfn_ros.cpp
DatSpace: Changed the default value to 0 and added a check to not do anything in that case.

navfn/include/navfn/navfn_ros.h
DatSpace: Renamed typos.

DatSpace:
The goal is to be able to avoid going all around in a path. For example if you have a room with a doughnut like shape and the path is blocked somewhere, I dont want the robot to pathfind all the way around to the other side of the blocked path.

By setting a limit to the path (optionally) we can avoid that very simply.

With the way it is implemented, navfn will throw a warning and then it will try to re plan until it times out. After that it will go into recovery as expected, which will clear any possible obstacles in the path etc, making it a very good way to do this, I believe.

mikeferguson:

PR #1137, opened 2021-07-07
Dima/update origin

dorezyuk:
Hey, 
I've noticed that the updateOrigin function would always create a local copy of the window... I've rewrote the function such that it can do the update inplace without the extra buffer. This gives some nice performance boost (more prominent for larger costmaps, in the supplied perf-flle ~30%).

Bonus: fixup the assignment operator. 

Best, 
Dima

### DIFF
--- a/costmap_2d/CMakeLists.txt
+++ b/costmap_2d/CMakeLists.txt
@@ -218,3 +218,21 @@ install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
   PATTERN ".svn" EXCLUDE
 )
+
+###############
+## Benchmark ##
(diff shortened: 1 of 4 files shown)


costmap_2d/CMakeLists.txt
```diff
@@ -215,3 +218,21 @@ install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
   PATTERN ".svn" EXCLUDE
 )
+
+###############
+## Benchmark ##
+###############
+
+set(${PROJECT_NAME}_BENCHMARK ON)
```
dorezyuk: maybe we should set this to off by default...

mikeferguson:

PR #1138, opened 2021-07-09
avoid copying of data when dealing with observations

dorezyuk:
Hey,

I've noticed that we create a deep copy of the point clouds every time we are querying to get the observations. I've exchanged the raw pointer to a const shared pointer (we do not modify the observations).

Additionally I've changed the box-filter algorithm to work without an additional buffer.

The changes reduced the CPU load of a mbf-node (just running the voxel layer) on my system by ~7% (its not much, but...)

Best,
Dima


mikeferguson:

Issue #1114, opened 2021-05-07
Why is observation_persistence is compared to the last update time instead of the current time?


nightingale0131:
This is more of a question than an issue, but I noticed that the `observation_persistence` parameter in the obstacle layer is used to compare observations with the last known update rather than the current time.

This can be seen in the `ObservationBuffer::purgeStaleObservations()` function (`observation_persistence` is passed into `observation_keep_time_`). This is an issue for me because I have an observation source that does not populate the buffer at a given rate, in fact, it very rarely populates the buffer. Instead of the observation being cleared after `observation_persistence` (say 1s) has passed, it hangs around indefinitely in the buffer until the next observation is received.

Does anyone know why this is the case? Is this by design or a bug?
I ask because it's not how I would have expected `observation_persistence` to be used.

PR #1140, opened 2021-07-13
clear already reported observations

dorezyuk:
Hey,

The current getObsevation function calls purgeStaleObservation - this function is in the current state a noop, since we don't update the last_updated_ time (as addressed in #1114).

The other change is clearing already reported observations. Currently, for observation_persistence larger than the update_frequency, same points would be painted onto the observation layer. I think it makes sense to drop the observations already incorporated into the costmap.

Best,
Dima


mikeferguson:

PR #1143, opened 2021-07-15
Make costmap size use double instead of int

lukschwalb:
Not sure if this is a bug but I don't see any reason not to use a double for the costmap size.
The getSizeInMeters calls also use double as a return type.
I'm working on a small vehicle where a size of less than a meter would come in handy.


mikeferguson:
Retriggering CI

mikeferguson:
It seems that the integer size has remained in ROS2 Navigation2 - and so it seems pretty late in the lifecycle of ROS1 nav to make this change (especially when the successor to this is keeping the integer size). I'm going to close this, since I think the tradeoff between "making a map less than a meter" isn't worth the possible downsides of making this a double

Issue #1160, opened 2021-08-13
AMCL miscalculates orientation covariance for clusters


christofschroeter:
pf_cluster_stats() determines clusters in the particle set, and calculates mean and covariance over the entire set as well as for each individual cluster.

It uses the formula for circular standard deviation to calculate the covariance matrix entry for the orientation: https://en.wikipedia.org/wiki/Directional_statistics#Measures_of_location_and_spread

`set->cov.m[2][2] = -2 * log(sqrt(m[2] * m[2] + m[3] * m[3]));`

However, this formula only yields the correct result if the sum of sample weights is 1.0. That does not seem to be a problem for the particle set covariance, as pf_cluster_stats() is generally called after normalizing particle weights. 

Unfortunately, when there are multiple clusters in the particle set, with each cluster containing only a fraction of the particles, the sum of weights in each cluster is not 1.0, and the formula used is not valid for calculation of the cluster's covariance matrix (orientation deviation). The covariance is always overestimated in this case. The smaller the cluster in relation to the full particle set, the stronger the deviation.
(Of course, it is possible to ignore cluster statistics and just use the set statistics to describe the estimated uncertainty. But somehow that defeats the purpose of using a particle filter at all, i.e. possibility to estimate multi-modal distributions?)

Normalization of the weights would result in scaling cluster->m[2]/m[3] by 1.0 / cluster->weights, or just offsetting the result by -2*log(1.0 / cluster->weights).

christofschroeter:
By the way the correctness of the calculation can be verified by checking against the simple non-circular calculation for a small-uncertainty case (where circular calculation should not really be necessary, but must still give the correct result).

PR #1161, opened 2021-08-13
Fix estimation of orientation covariance for clusters

christofschroeter:
#1160 AMCL miscalculates orientation covariance for clusters

### DIFF
--- a/amcl/src/amcl/pf/pf.c
+++ b/amcl/src/amcl/pf/pf.c
@@ -662,7 +662,7 @@ void pf_cluster_stats(pf_t *pf, pf_sample_set_t *set)
     // Covariance in angular components; I think this is the correct
     // formula for circular statistics.
     cluster->cov.m[2][2] = -2 * log(sqrt(cluster->m[2] * cluster->m[2] +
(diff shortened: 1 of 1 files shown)


mikeferguson:
Do you have any sense of how this might affect a particular tuning of AMCL? I know there are some other spots in the codebase where math was wrong, but when we fixed it the behavior for a particular tuning would become quite bad.

christofschroeter:
It does not affect the behaviour of the particle set nor the calculated mean of the distribution. Only the calculated variance of the orientation is changed (which is purely informational, not used in the AMCL loop). More specifically, only in a case where multiple particle clusters form and the covariance in a specific cluster may become relevant. This seems to be a rather rare case in our applications, but magnitude of uncertainty may depend on the used robot and sensors.

The one case where this has a visible effect on the particle distribution (and how I found it) is if you read out the estimated variances and then use them to re-initialize the particle distribution, e.g. when implementing persistent localization on top of AMCL (and only if there actually are multiple particle clusters and you select the best one or any specific cluster - which you should do to get a reasonable variance in this situation of course).

mikeferguson (approved):

PR #1180, opened 2021-11-20
Added publishZeroVelocity() when state_ is changed to PLANNING.

ryan-sandzimier:
If a robot is on it's way to the goal and a new MoveBaseActionGoal is published, the robot continues moving at the last commanded velocity for the duration of planning a path to the new goal. When the global planner is slow (such as for sbpl_lattice_planner), this is problematic. #750 and #751 fix this for the initial goal. However, if a new goal is published before the robot reaches the preceding goal, the same problem exists. 

Solved this by adding publishZeroVelocity() when state_ is changed to PLANNING so that a zero velocity is commanded prior to making the plan.

### DIFF
--- a/move_base/cfg/MoveBase.cfg
+++ b/move_base/cfg/MoveBase.cfg
@@ -29,5 +29,7 @@ gen.add("oscillation_distance", double_t, 0, "How far in meters the robot must m
 gen.add("make_plan_clear_costmap", bool_t, 0, "Whether or not to clear the global costmap on make_plan service call.", True)
 gen.add("make_plan_add_unreachable_goal", bool_t, 0, "Whether or not to add the original goal to the path if it is unreachable in the make_plan service call.", True)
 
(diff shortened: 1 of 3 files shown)


mikeferguson (changes requested):
So, I know of at least a few people who leverage this "feature" - in that the robot won't stop real hard if they can actually get a new plan in time (basically, it's a lot smoother to let the robot coast for a second). I don't think we can merge this as-is into an LTS release that is 4 years into the cycle - but if you put it behind a parameter, that defaults to the old behavior, we could merge it.

ryan-sandzimier:
> So, I know of at least a few people who leverage this "feature" - in that the robot won't stop real hard if they can actually get a new plan in time (basically, it's a lot smoother to let the robot coast for a second). I don't think we can merge this as-is into an LTS release that is 4 years into the cycle - but if you put it behind a parameter, that defaults to the old behavior, we could merge it.

@mikeferguson Added a parameter that defaults to old behavior.

mikeferguson:

PR #1197, opened 2022-05-10
Changed move_base action name

yuckey0524:
Changed move_base action name.
Because original move_base action is located under global name space.
When we want to use multiple move_base under one ros_core, move_base action name will conflict.
So, I located move_base action under private node handle.


mikeferguson:
Yeah, there's just no way we are going to change the public API in a released version of navigation. You should be able to accomplish this same thing using a namespace and remapping in your launch file.

PR #1198, opened 2022-05-16
use boost::placeholders::_N instead of deprecated _N

lucasw:
This eliminates a warning message about _1/_2 etc. and boost/bind.hpp being deprecated encountered building this on Ubuntu 22.04 (and from similar changes in other repos probably Arch and Gentoo see them also).

I also had to change to C++17 to avoid std::share_mutex log4cxx build errors- but could separate that out if necessary

    /usr/include/log4cxx/boost-std-configuration.h:10:18: error: ‘shared_mutex’ in namespace ‘std’ does not name a type

### DIFF
--- a/amcl/CMakeLists.txt
+++ b/amcl/CMakeLists.txt
@@ -4,10 +4,6 @@ project(amcl)
 include(CheckIncludeFile)
 include(CheckSymbolExists)
 
-if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 11)
-endif()
-
 find_package(catkin REQUIRED
   COMPONENTS
     diagnostic_updater
--- a/amcl/src/amcl_node.cpp
+++ b/amcl/src/amcl_node.cpp
@@ -26,7 +26,7 @@
 #include <cmath>
 #include <memory>
 
-#include <boost/bind.hpp>
+#include <boost/bind/bind.hpp>
 #include <boost/thread/mutex.hpp>
 
 // Signal handling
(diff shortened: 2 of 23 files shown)


amcl/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ include(CheckIncludeFile)
 include(CheckSymbolExists)
 
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD 17)
```
mikeferguson: the recommended C++ for Noetic is 14 - I don't think we want to change this to be 17 - can the rest of these changes be done without changing the version?

mikeferguson (changes requested):

amcl/CMakeLists.txt
lucasw: If I delete that line entirely it will default to 14 on Ubuntu 20.04 and 17 on Ubuntu 22.04- would that be preferable?

mikeferguson: Yeah, I think that is probably the correct thing to do here

lucasw:
The boost bind change was later done differently with lambdas in https://github.com/ros-planning/navigation/pull/1225/commits/80bf439cf98b0da5deb7c6d38075f8d112fc03ca, hpp includes in https://github.com/ros-planning/navigation/pull/1225/commits/eb26f83a20feba47c2b14b9d601320fb8915925f,  c++ version in https://github.com/ros-planning/navigation/pull/1225/commits/977c955393f6ba151362eb67e841c14e91f72da9 and merged in https://github.com/ros-planning/navigation/pull/1225

PR #1200, opened 2022-06-02
Check if stamp of transforomed pose is non-zero

furushchev:
If the chain of transformations between `global_frame_` and `robot_base_frame_` are all broadcast-ed as static TF frames (using TF2), the stamp of transformed frame is always zero, which is trapped as timeout error.
With this PR if the stamp is zero, no timeout is checked.


mikeferguson (approved):
This is a great catch

PR #1201, opened 2022-06-09
`SimpleScoredSamplingPlanner`- try fallback generators even if valid trajectories were found (option)

rayvburn:
### Why this MR was created?

Currently It is impossible to use multiple trajectory generators that are treated equally (when second and next ones are not fallback generators). It is a clear limitation to local planner design.

### What was done?

I've added extra parameter to the `SimpleScoredSamplingPlanner` constructor so user can actually select if, e.g., the second generator is the fallback one or should always be checked (just like the primary generator).

### Does it affect API of the previous applications?

No, because:

- default value is set to false which does not change behavior compared to the current version
- extra constructor argument is equipped with default value so API of the present applications does not change


mikeferguson:

PR #1207, opened 2022-07-06
gradient_path: the gradx_ and grady_ can be negative, so we check the…

tyuownu:
We can see that `gradx_` and `grady_` value can be negative, so we can just compare the sum to 0.0, to check the cell have been calculated.


rsmyczyn:
good

mikeferguson:

PR #1226, opened 2022-11-28
[AMCL] Add option to force nomotion update after initialpose

stwirth:
* Adds a new boolean parameter `force_update_after_initialpose`. When set to `true`, an update is forced on the next laser scan callback, such as when the `/request_nomotion_update` service is called. This often results in an improved robot pose after a manual (not very precise) re-localization - without a need for the robot to move.
* Fixes a bunch of compiler warnings (unused variable `now`, catching exceptions by value), normalizes how tf exceptions are caught.

### DIFF
--- a/amcl/cfg/AMCL.cfg
+++ b/amcl/cfg/AMCL.cfg
@@ -28,6 +28,7 @@ gen.add("beam_skip_distance", double_t, 0, "Distance from a valid map point befo
 gen.add("beam_skip_threshold", double_t, 0, "Ratio of samples for which the scans are valid to consider as valid scan", 0.3, 0, 1)
 
 gen.add("tf_broadcast", bool_t, 0, "When true (the default), publish results via TF.  When false, do not.", True)
+gen.add("force_nomotion_update_after_initialpose", bool_t, 0, "When true, force a pose update after a new /initialpose is received.", False)
 gen.add("gui_publish_rate", double_t, 0, "Maximum rate (Hz) at which scans and paths are published for visualization, -1.0 to disable.", -1, -1, 100)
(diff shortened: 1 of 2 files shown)


stwirth:
I added another commit to add a second parameter `force_update_after_set_map` which - when set to `true` - forces an update after the `/set_map` service has been used.

amcl/cfg/AMCL.cfg
```diff
@@ -28,6 +28,8 @@ gen.add("beam_skip_distance", double_t, 0, "Distance from a valid map point befo
 gen.add("beam_skip_threshold", double_t, 0, "Ratio of samples for which the scans are valid to consider as valid scan", 0.3, 0, 1)
 
 gen.add("tf_broadcast", bool_t, 0, "When true (the default), publish results via TF.  When false, do not.", True)
+gen.add("force_nomotion_update_after_initialpose", bool_t, 0, "When true, force a pose update after a new /initialpose is received.", False)
+gen.add("force_nomotion_update_after_set_map", bool_t, 0, "When true, force a pose update after a new map (and pose) has been set via the /set_map service.", False)
```
mikeferguson: is there a reason to include "nomotion" in the names? These are pretty long variable names, and "force_update_after_X" seems to convey exactly what is going on.

stwirth: Sounds good, I'll rename.

stwirth: @mikeferguson done

mikeferguson (approved):

stwirth:
Thanks @mikeferguson!

stwirth:
@mikeferguson would you be so kind and bloom a noetic release to get this in?

mikeferguson:
https://github.com/ros/rosdistro/pull/35816

stwirth:
Awesome, thank you @mikeferguson!

PR #1229, opened 2022-12-26
fix global planner default tolerance

ChristofDubs:


ChristofDubs:
wrong target branch, sorry

PR #1236, opened 2023-02-14
Fix static layer not transforming to global frame

siferati:
## Description

Static layer wasn't transforming to `global_frame_` when computing the update bounds.
This resulted in the static obstacles not being drawn correctly in the costmap when `map_frame_` and `global_frame_` are different.

Easy to reproduce if you set `map_frame_` and `global_frame_` to different values and disable every layer except the static one.


mikeferguson:

PR #1234, opened 2023-02-06
1802 show jig nav

yuanfanwang:
navigation がactuator の状況を見て、footprint を決定する処理を[#1802](https://github.com/LexxPluss/LexxAuto/issues/1802) で行うようにしたい。

PR #1237, opened 2023-02-15
1802 show jig nav3

yuanfanwang:
#1234 
added extended_footprint for passing the test.


mikeferguson:
I'm thinking you meant to open this against a different repo - but there is no way this is able to be reviewed and merged here (300+ commits touching 120 files)

PR #1246, opened 2023-08-08
fix global planner path end outside tolerance: simplified

ChristofDubs:
simpler alternative of https://github.com/rapyuta-robotics/ros-navigation/pull/84


ChristofDubs:
wrong target branch

Issue #1010, opened 2020-07-10
YAML errors in map_server not caught


celynw:
When running `rosrun map_server map_server map.yaml` or equivalent, any errors cause it to close without producing any output.
For instance, setting the resolution as a string.
I see in the code (map_server/src/main.cpp) it is supposed to produce a `ROS_ERROR` if the resolution could not be loaded, but this is never printed.
```c++
        try {
          doc["resolution"] >> res;
        } catch (YAML::InvalidScalar &) {
          ROS_ERROR("The map does not contain a resolution tag or it is invalid.");
          exit(-1);
        }
```
In my usage I never get any of these error messages.

I can't find and specifically catch what _is_ being thrown, but catching `std::exception::what()` does work:
```
yaml-cpp: error at line 0, column 0: bad conversion
```

### Environment:
- OS: Ubuntu 18.04
- Branch: `melodic-devel`
- Commit: `b496f689`

acaviedes15:
any update?

maker-ATOM:
Setting,
```
resolution: "0.050000"
```
within the `.yaml`

Allows the current method to parse the resolution, is this acceptable?

PR #1251, opened 2023-09-17
Use YAML::BadConversion exception to catch undefined and invalid YAML parameters

AlaHammouda:
This pull request addresses issue https://github.com/ros-planning/navigation/issues/1010

* Use YAML::BadConversion exception to catch undefined and invalid YAML parameters.

### DIFF
--- a/map_server/src/main.cpp
+++ b/map_server/src/main.cpp
@@ -210,25 +210,25 @@ class MapServer
 #endif
       try {
         doc["resolution"] >> res;
-      } catch (YAML::InvalidScalar &) {
+      } catch (YAML::BadConversion &) {
         ROS_ERROR("The map does not contain a resolution tag or it is invalid.");
         return false;
       }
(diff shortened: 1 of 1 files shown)


map_server/src/main.cpp
```diff
@@ -210,25 +210,25 @@ class MapServer
 #endif
       try {
         doc["resolution"] >> res;
-      } catch (YAML::InvalidScalar &) {
+      } catch (YAML::BadConversion &) {
```
mikeferguson, line 213: I don't know YAML exceptions well - but should we instead be using the parent class of both InvalidScalar and BadConversion - YAML::RepresentationException?

maker-ATOM:
Setting,
```
resolution: "0.050000"
```
within the `.yaml`

Allows the current method to parse the resolution, is this acceptable?

mikeferguson:

Issue #1125, opened 2021-05-29
Warning: TF_REPEATED_DATA ignoring data with redundant timestamp for frame odom_comb


fjp:
Hi, when working with `amcl` for example the mir-robot [Gazebo demo (existing map)](https://github.com/dfki-ric/mir_robot#gazebo-demo-existing-map) a lot of the following warnings are logged:

```
Warning: TF_REPEATED_DATA ignoring data with redundant timestamp for frame odom_comb at time 23.567000 according to authority unknown_publisher
         at line 278 in /tmp/binarydeb/ros-noetic-tf2-0.7.5/src/buffer_core.cpp
```

I guess this is related to https://github.com/ros/geometry2/issues/467. To reduce the number of `TF_REPEATED_DATA` warnings the laser scan publish rate needs to be reduced, because `amcl` depends on rate of incoming laser messages.

According to the discussion [here](https://github.com/dfki-ric/mir_robot/issues/88) this seems to be a bug in `amcl`:

> Basically, once `amcl` has published a TF for a given time stamp `t`, it must never publish it again. So there are two options: Either publish the first message and then drop all subsequent ones, or delay publishing the message until amcl is certain that this is the definitive one. Both have their advantages and disadvantages.

Can you provide some more hints on how to fix this warning? I assume the problem is in [`laserReceived()`](https://github.com/ros-planning/navigation/blob/2b807bd312fac1b476851800c84cb962559cbc53/amcl/src/amcl_node.cpp#L1119)?

mikeferguson:
I don't actually see how it would be possible for AMCL to publish two transforms with the same timestamp unless it receives two laser scans with the same timestamp - as far as I can see, the only place we publish transforms is inside laserReceived() - and we use the timestamp of the laser message received. To me that indicates either an issue with the Gazebo laser plugin, or some issue in ROS that is causing the same message to be delivered to AMCL twice.

fjp:
I can confirm that the warning log messages can be reduced when the `update rate` of the laser is reduced. Probably there is an issue in the  [`gazebo_ros_gpu_laser`](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_gpu_laser.cpp) plugin that I am using or maybe there is an issue in gazebo and its [`sensor`](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/sensors/Sensor.hh) update behaviour.

[...]

PR #1260, opened 2024-02-26
Warning: TF_REPEATED_DATA ignoring data with redundant timestamp for frame odom_comb at time 23.567000 according to authority unknown_publisher

Daviesss:
This pull request addresses issue https://github.com/ros-planning/navigation/issues/1125

   I solved this problem by making a small tweak to the amcl_node.cpp file. In order to decrease the frequency of TF_REPEATED_DATA warnings, I lowered the publishing rate of the laser scan because amcl relies on the rate of incoming laser messages. I made this adjustment directly in the .cpp node, and it worked perfectly for me. As a result, the redundant messages no longer appear, which is fantastic.

### DIFF
--- a/amcl/src/amcl_node.cpp
+++ b/amcl/src/amcl_node.cpp
@@ -836,7 +836,7 @@ void
 AmclNode::checkLaserReceived(const ros::TimerEvent& event)
 {
   ros::Duration d = ros::Time::now() - last_laser_received_ts_;
-  if(d > laser_check_interval_)
+  if(d < laser_check_interval_)
   {
     ROS_WARN("No laser scan received (and thus no pose updates have been published) for %f seconds.  Verify that data is being published on the %s topic.",
              d.toSec(),


mikeferguson (changes requested):
This seems wrong - laser_check_interval_ is set to 15 seconds - so with your change, if the last laser scan has arrived within 15 seconds this message prints (every 15 seconds, when the timer event hits).

If your laser scans are really showing up only once every 15 seconds (or more than 15 seconds in the past) - you've got bigger problems a few ROS_WARN messages.

Daviesss:
I get you . But in my case ,the trigger  warning message when the laser scan has been received within the specified interval, then the opposite condition would be when the time duration d since the last laser scan received is less than the laser_check_interval_. I think the condition should be :

 if (d < laser_check_interval_) {
    ROS_WARN("Laser scan received within the specified interval of %f seconds.",
             d.toSec());
}

else {
   ROS_WARN("No Laser scan received within the specified interval of %f seconds.",
             d.toSec());
}

With this condition, the warning message will be triggered when the time duration since the last laser scan received is less than the laser_check_interval_, indicating that a laser scan has been received within the expected interval.

mikeferguson:

PR #1261, opened 2024-02-29
Transforms to global frame before checking quaternion

robbiefish:
This PR allows goals to be accepted if the quaternion is valid once the goal has been transformed into the `global_frame` instead of requiring the quaternion to be valid in the original frame it was sent in.

For my application, I have defined the transform from `earth -> map`. My map frame is in the ENU local tangent plane. Since there is a rotation between ECEF and ENU, that rotation is expressed in my transform tree.

The `global_frame` that I configure for the planner is still the `map` frame, and I want to be able to send waypoints in both the `earth`, and `map` frame.

When I attempt to send a goal point in the `earth` frame, the quaternion is not "mostly vertical" in the `earth` frame, but it is once it has been transformed to the `global_frame`.

### DIFF
--- a/move_base/src/move_base.cpp
+++ b/move_base/src/move_base.cpp
@@ -650,12 +650,12 @@ namespace move_base {
 
   void MoveBase::executeCb(const move_base_msgs::MoveBaseGoalConstPtr& move_base_goal)
   {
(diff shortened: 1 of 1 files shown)


mikeferguson:

PR #1264, opened 2024-05-15
The GridPath:: getPath function stopped prematurely while backtracking the path, resulting in the generation of incorrect global paths.

JackJu-HIT:
Through analysis of the algorithm and theoretical thinking, we have found that there is an issue with the code for the backtracking path of GridPath:: getPath. When searching for the backtracking path, it did not restrict the range of the inflated map. As the index mapping to Cartesian coordinates does not have uniqueness, if the search map range is not restricted, there is a good chance of finding an index value outside the map range that is the same as the starting point, and the backtracking function will terminate prematurely, resulting in the generation of an incorrect path that does not include the starting point.


mikeferguson:

